Issue #7, opened 2026-09-23
Bug: an out-of-range Ctab src_type can panic on the RT thread


ryangavin:
- `importance` and `mask_of` index `TONES[ty]` without a bounds check. Clamp or validate at parse time, and add a fuzz/corpus test.

---
Milestone: **M1 · Chord recognition 1:1** · Spec: [docs/genos-features.md](../blob/main/docs/genos-features.md) · Backlog: [docs/backlog.md](../blob/main/docs/backlog.md)

ryangavin:
Decision (PR #39): an out-of-range Ctab/Ctb2 source chord type (≥ 34, including 0x22 Cancel) is replaced with Maj7 at parse time instead of rejecting the style. Maj7 is the documented Style Creator default source chord (RM p.28), it matches our default_for rules, and one bad channel record shouldn't make a whole style unloadable.

ryangavin:
Decision (PR #39): out-of-range chord ids inside theory.rs (chord_tones/mask_of/importance) read as a major triad, not sus2 as the old clamp did, because a triad is the most neutral fallback. It is only a safety net: the parser clamp and the Cancel gating in plays() mean no valid input reaches it.

ryangavin:
Decision (PR #39): invalid tempo (0 µs) and time signature (numerator 0, denominator exponent ≥ 8) meta events are ignored and the style keeps the defaults of 120 BPM and 4/4. These values are impossible on the hardware and would otherwise produce a zero-length bar.

Issue #6, opened 2026-09-23
Keyboard transpose feeds the chord root


ryangavin:
- Transpose shifts both the played notes and the chord root sent to the style. Cover Keyboard, Song and Master transpose (song transpose only as far as it affects playing).

---
Milestone: **M1 · Chord recognition 1:1** · Spec: [docs/genos-features.md](../blob/main/docs/genos-features.md) · Backlog: [docs/backlog.md](../blob/main/docs/backlog.md)

ryangavin:
Decisions made in PR #40 (the owner may overrule after playtesting):

- Decision: Master transpose shifts the style's output notes after chord transposition and does not move the chord root, because the manual says Master "transposes the pitch of the entire sound" while only Keyboard is described as including "the chord root for triggering Style playback". Shifting the output keeps High Key and note-limit octave placement and Bypass channels the same as the Genos would.
- Decision: the chord display shows the chord the style follows (fingered chord + Keyboard transpose), with the fingered chord shown next to it when they differ, because Keyboard transpose is defined as changing the chord that triggers the style. Master transpose does not change the displayed chord.
- Decision: a transpose change applies to notes started afterwards. Notes already sounding (yours and the style's) keep their pitch and are released at it, because Yamaha tone generators apply transpose at note-on, and doing it this way avoids re-attacks and stuck notes.
- Decision: a Keyboard transpose change while a chord is held (or remembered) moves the style's chord right away, and parts follow their RTR rules, because Keyboard transpose is defined in terms of the keys and the held keys now stand for the new chord. Otherwise the band would stay in the old key until the next chord.
- Decision: chord recognition runs on the fingered keys and the engine shifts the result. This is equivalent to recognising the transposed keys, since recognition doesn't depend on key, and it is what makes the previous decision possible.
- Decision: drum/SFX kit exclusion for the style is decided per part at load time from the part's initial voice (drum parts ch 9/10, or bank MSB 126/127), because the manual says transpose never applies to kits and the engine must not look anything up at play time. A kit selected mid-pattern by a later program change is not tracked.
- Decision: keyboard (RH/LH) notes are always transposed, because the built-in synth's keyboard slots only hold melodic GM programs today. If kit voices are added to the keyboard parts, they should be excluded the same way.
- Decision: Song transpose is not implemented, because yahaha has no MIDI song player, so it has nothing to act on in live play. "Transpose MIDI Input" behaves as On, because our keyboard reaches us over MIDI and transposing it is the whole point.
[...]

Issue #4, opened 2026-09-23
Chord Cancel and the 1+5 / 1+8 behaviour


ryangavin:
- Cancel (root + b2 + 2) only exists in Fingered, On Bass and AI Fingered.
- Decide and document what each part plays under Cancel, 1+5 and 1+8, using CASM chord mute plus the corpus-consistency check (M2).

---
Milestone: **M1 · Chord recognition 1:1** · Spec: [docs/genos-features.md](../blob/main/docs/genos-features.md) · Backlog: [docs/backlog.md](../blob/main/docs/backlog.md)

ryangavin:
These are the decisions made in #43 (Chord Cancel, 1+8 and 1+5 playback rules). All of them are open to playtesting and can be overruled:

- Decision: Chord Cancel = the no-chord state. Rhythm channels and CASM-autostart channels play as recorded; all other parts are released immediately and rest until the next chord. Because: the OM (p.46) defines Cancel as "a state in which no chord is input", and the engine already defines that state (before the first chord) this way. In the corpus, autostart appears only on rhythm channels, so in practice this means "rhythm only".
- Decision: Chord Cancel does not trigger Sync Start (unchanged). Because: the OM (p.66) says the Style starts when "you play a chord", and Cancel is the absence of a chord.
- Decision: 1+8 and 1+5 use the CASM chord-mute bits 30/31 as they are, with no fallback routing. Because: every corpus style that uses chord-type routing (the NightCruiser and SmoothItOver bass channels) partitions all 34 types across its source channels. 30/31 always land on the major or "everything else" source. The bits are the author's intended routing.
- Decision: under 1+8, chord-following parts (every NTT except Bypass) play only the root, in octaves. Chromatic notes and the Guitar "fifth" string collapse to the root. Because: Yamaha describes 1+8 as accompaniment based only on the root.
- Decision: under 1+5, Melody-type tables keep root, 5th and the 2nd/4th. The 3rd and 6th go to the 5th, the 7th goes to the octave, and chromatic notes snap to the scale degree below. Chord-table parts keep only root and 5th. Because: 1+5 must fit both major and minor, and the 2nd and 4th are the only other degrees the two share. Collapsing everything to root/5th would flatten walking bass lines more than needed.
- Decision: NTT Bypass channels still play as written under 1+8/1+5, 3rds included. Because: Bypass means the author asked for no chord conversion.

Issue #2, opened 2026-09-23
Recognizer: match the Data List chord table exactly


ryangavin:
- Cover all 37 entries, including the ones that have no MIDI code: M7b5, (b5) and mM7b5. Decide which CASM type each of those maps to.
- Handle the voicings where notes may be left out, e.g. `1+3+(5)+7`.
- Produce 1+8. Right now an octave collapses to major.
- Resolve ambiguous inversions (C6 vs Am7, dim7, aug, m6 vs m7b5) and record each rule in the spec.
- Write one test per row.

---
Milestone: **M1 · Chord recognition 1:1** · Spec: [docs/genos-features.md](../blob/main/docs/genos-features.md) · Backlog: [docs/backlog.md](../blob/main/docs/backlog.md)

ryangavin:
Decisions made while implementing this ticket in #44. Each one is also written up in the spec, docs/genos-features.md §C.4a. The owner may overrule any of them after playtesting.

- Decision: a single key is not a Fingered chord, and no two-note set is either except 1+5 (fifth either way up) and 1+8. The previous chord stays. Because RM p.9 says the thing that sets AI Fingered apart is that "less than three notes" can indicate a chord, and the Data List's only two-note rows are 1+5 and 1+8. **Behaviour change:** before, C alone gave C, and C+E gave C.
- Decision: 1+8 means two or more held keys that all share one pitch class. Because the name "1+8" is root plus octave, and a lone key is covered by the decision above.
- Decision: ambiguous pitch-class sets are resolved in this order: (1) the root is the lowest note; (2) fewest omitted notes; (3) the lowest note's role, ranked root > 5th > 3rd > b5/#5/4th > 6th/7th > tension; (4) Data List table order. Because rule 1 is the only inversion rule the manuals imply (Fingered On Bass takes the lowest note as bass), and it makes dim7 and aug take the lowest note as root. Rules 2 and 3 pick the reading a player most likely means when the root is not in the bass. Examples: E G A C = Am7/E, G A C E = C6/G, Eb G A C = Cm6/Eb, G C F = Csus4/G.
- Decision: in Fingered On Bass, a set that is not a table chord but whose notes above the bass form a table chord of three or more notes is read as that chord over the bass (F# C E G = C/F#). A full-table reading always wins, so D C E G = Cadd9/D. Plain Fingered does not recognise the slash-only form. Because On Bass is defined as "the lowest note becomes the bass", and plain Fingered only knows the table shapes.
[...]

PR #39, opened 2026-09-23
Clamp malformed style data so the RT transposer can't panic (#7)

ryangavin:
## What / why
`importance()` and `mask_of()` in `theory.rs` indexed `TONES[ty]` without a bounds check. A Ctab/Ctb2 record with a source chord type of 34 or more could therefore panic the engine thread. So could a Cancel chord reaching the transposer. The engine also plays autostart and drum parts with `Chord::new(src_root, src_type)` before any chord is held, which puts a raw `src_type` directly on the RT path.

**At parse time (`sff.rs`)**
- The source chord type is validated. Values 0..=33 pass through. Anything else, including 0x22 Cancel, becomes Maj7 (see Decisions).
- MIDI data bytes are masked to 7 bits, so a malformed track can't send a key or program of 128 or more into the engine or synth.
- The parser used to panic on some malformed headers and meta events. Now:
  - An MThd length under 6, or one that overruns the file, returns an error.
  - A PPQ of 0 returns an error.
  - Tempo meta events of 0 are ignored.
  - Time signatures with numerator 0 or denominator 2^8 or more are ignored.
- Tick accumulation saturates instead of overflowing. Bar length is never 0, which previously caused a divide-by-zero.
- The OTS octave decode no longer overflows an `i8`.
- Section-length bar rounding saturates, so a saturated end tick can't overflow it (debug panic, or a silent wrap to one bar in release).
- A Ctab/Ctb2 destination channel outside 9-16 falls back to the source channel's accompaniment part (see Decisions), so style notes never land on the keyboard or pad channels 1-8.

**Defence in depth (`theory.rs`)**
- Every chord-tone lookup now goes through a bounds-checked `chord_tones()`. An out-of-range id reads as a major triad; the old fallback was sus2. `mask_of`, `importance`, the recognizer builder and `guitar_voicing` all use it.
- `plays()` uses checked shifts, so a type or root outside the mute masks doesn't play instead of overflowing.
- `Chord::name` and `guitar_voicing` accept any type, root or bass value.

The rest of the `theory.rs` audit found no other unchecked indexing that style data can reach:
- NTR, NTT and RTR decode through enums with fallbacks.
- High Key is taken `% 12`, and source root `% 12`.
- Note limits are masked to 7 bits. Inverted limits skip the fold and only clamp.
- `mid_lo` and `mid_hi` are only compared.
- Source channels are masked to `& 0x0F`. Destination channels are forced into 8..=15.

## How tested
- `cargo build --release`, `cargo test --release` (26 passed, 1 ignored that was already ignored) and plain `cargo test` all pass. The debug build has overflow checks on, so it catches shift and arithmetic overflow too.
- Corpus tests ran for real: the new corpus test reports "35 styles, 469 channel rules".
- `cargo clippy --release`: still 32 warnings, the same as before, and none are in new code.
- New tests:
  - `sff::source_chord_type_is_clamped`: Ctab and Ctb2 records with src_type 34, 63, 0x7F and 0xFF.
  - `sff::malformed_ctab_transposes_without_panicking`: synthetic Ctb2 and Ctab records with out-of-range source types and garbage zone bytes (0x00, 0x07, 0x7F, 0x80, 0xFF) for NTR, NTT, RTR, High Key and inverted note limits. Each record goes through the full `parse()` of a minimal style file, then every key is transposed under every chord.
  - `sff::truncated_and_corrupted_files_do_not_panic`: parses every prefix of that file, plus 4 deterministic byte flips at every offset. Any parse that succeeds is then run through the transposer.
  - `theory::out_of_range_types_do_not_panic`: hand-built rules that skip the parser clamp, across all NTR and several NTT types. Also exercises `plays`, `transpose`, `guitar_voicing` and `name` with a type or root out of range.
  - `sff::huge_deltas_do_not_overflow_section_length`: a Main A marker, then 20 events with delta `FF FF FF 7F`. It panics in a debug build without the saturating rounding.
  - `sff::dest_channel_stays_on_accompaniment_parts`: destination bytes 0, 7, 0x1F and 0xF3.
  - Both `exercise()` helpers now also assert that `transpose()` returns `None` or a note ≤ 127, that `plays(rule, Cancel)` is true only for drum parts, and (in `sff`) that every parsed `dest_ch` is in 8..=15.
  - `theory::corpus_every_rule_every_chord`: every channel rule of every corpus style, plus the default rules. Covers all 128 keys with `transpose`, and grouped triads with `transpose_group`, under every chord type including Cancel, all 12 roots, slash chords and the rule's own source chord.
- On the old `theory.rs`, both theory tests fail with an index-out-of-bounds panic, which confirms they reproduce the bug.

## Decisions
- Decision: an out-of-range source chord type (≥ 34, including 0x22 Cancel) is replaced with Maj7 (2) at parse time rather than rejecting the style, because Maj7 is the documented Style Creator default source chord (RM p.28, spec C.5). It is also what our `default_for` uses, and a single bad channel record shouldn't make a whole style unloadable.
- Decision: out-of-range chord ids inside `theory.rs` read as a major triad, not sus2 as the old `chord_tones` clamp did, because a triad is the most neutral fallback. This is only a safety net: the parser clamp and the Cancel gating in `plays()` mean no valid input reaches it.
- Decision: invalid tempo (0 µs) and time signature (numerator 0, denominator exponent ≥ 8) meta events are ignored. The style keeps the defaults of 120 BPM and 4/4, because these values are impossible on the hardware and would otherwise produce a zero-length bar.
- Decision: a Ctab/Ctb2 destination channel below 9 is replaced with the source channel's accompaniment part (`src_ch | 0x08`), because Style Creator only offers destinations 9-16, source and destination normally match, and channels 1-8 carry the player's Right/Left voices and pads. No corpus style has a destination below 9.

Not in scope, noted for the owner: `live.rs` indexes `held[k]` with a raw CoreMIDI data byte. That comes from live input, not style data, and sibling tickets are editing that file. Its note-off path needs `& 0x7F` in a ticket of its own.

Closes #7


ryangavin:

**Verdict: no blocking findings.** Approve from my side, with the nonblocking items below.

### What I verified
- Checked out `origin/m1/7-rt-panic` (ce2803c) with the corpus symlinked. `cargo test --release`: 24 passed, 1 ignored. `corpus_every_rule_every_chord` really ran: "35 styles, 469 channel rules", which matches the 22 .STY + 13 .sty files in the corpus.
- `cargo clippy --release --all-targets`: the only warnings in `sff.rs`/`theory.rs` are on lines this PR doesn't touch (sff.rs:140, 334; theory.rs:196, 209, 313, 353, 354, 455). No new warnings.
- RT path audit. Everything below is bounded, and none of it allocates or locks:
  - `engine.rs`: `chord_for` (autostart/drum `Chord::new(src_root, src_type)`), `emit_at_index`, `revoice`, `sound_stop_acmp` and `volume`.
  - `theory.rs`: `transpose`, `transpose_group` and `guitar`.
  - How each is bounded: `src_type` is clamped at parse. `chord_tones` is checked. `plays` uses `checked_shr`. `fold_into` clamps its output to 0..=127. `dest_ch` and `src_ch` are masked with `& 0x0F`. `volume()` guards `ch - 8`. `sound_stop_acmp` returns early on Cancel before calling `chord_tones`.
- Decision check: the Maj7 fallback is well sourced. The Genos 2 Reference Manual says the default Source Root/Chord is CM7 and that Initialize Style restores it (Style Creator > SFF Edit section, around the Source Root/Chord parameter table). It also matches `ChannelRule::default_for`. The source-type list in spec C.5 matches `TONES` order 0..=33.

### Blocking
None.

### Nonblocking
1. **The debug build still overflows on huge tick totals. `saturating_add` moves the overflow to the next line instead of fixing it.** At `src/sff.rs:779`, `let len = (((end - start) + tpb / 2) / tpb).max(1) * tpb;` overflows when `end_tick` is near `u32::MAX`, and saturation now makes that value easy to reach.
   - Repro (throwaway test, not pushed): an SFF track with a `Main A` marker, then 20 events of delta `FF FF FF 7F` + `9B 3C 64`.
   - In a debug build, `parse()` panics with `attempt to add with overflow` at sff.rs:779. In release it wraps to a 1-bar section (384 ticks), and the events never play.
   - This runs on the load path, not the RT thread, and nothing panics in release. Still, the PR body says debug builds catch arithmetic overflow and that plain `cargo test` passes, and the fuzz test only runs this path under `--release`, so it can't see this.
   - Suggested fix: `(end - start).saturating_add(tpb / 2)` plus `saturating_mul`. Better still, reject tracks whose end tick exceeds some sane bound. Add the repro as a test.
2. **Out-of-range `dest_ch` is not clamped to the accompaniment channels.** `parse_ctab` sets `dest_ch: d[9] & 0x0F`, so a malformed Ctab with dest 0..7 sends style notes to the keyboard/pad channels (R1, Left, …).
   - It doesn't panic: `volume()` guards it, and `parts` uses `saturating_sub(8) & 7`.
   - It does collide with the player's own voices. Also, `all_off` only re-centres bends and clears CC1/CC64 on channels 8..16.
   - Suggestion: `d[9] & 0x0F | 0x08`, or fall back to `src_ch`. Style destination channels are 9–16 by definition.
3. **Merge-conflict risk with the recognizer tickets.** The two `TONES[ty]` → `chord_tones(ty)` edits in `Recognizer::new` (theory.rs:183, 187) touch the code #1–#6 are rewriting. The edits are trivial, but whoever merges second should re-apply them rather than accept either side blindly. Post-merge, `grep -n 'TONES\[' src/theory.rs` should show only the `chord_tones` accessor.
4. **Test strength.** `exercise()` (both copies) discards every result. Cheap assertions would turn the smoke test into a contract:
   - `transpose(..)` returns `None` or `<= 127`.
   - `plays(r, Chord::new(_, CANCEL))` is `is_drum_part(r.dest_ch)`.
5. **The corpus test passes silently when `corpus/` is missing** (`eprintln!("no corpus; skipping")`). This follows the existing convention, so it's fine for this PR. A follow-up could make it fail when an env var like `YAHAHA_REQUIRE_CORPUS` is set, so CI and the workflow gates can't pass vacuously.
6. **`live.rs` `held[k]` with a raw CoreMIDI byte.** You flagged this as out of scope. Agreed, but please make sure it gets a ticket rather than getting lost. A one-line `& 0x7F` fixes it, and it's the only remaining unchecked index I found that external input can reach on the MIDI thread.

The core fix is correct and well tested.

ryangavin:
Review round 1 is addressed in 646f798.

**Addressed**
- **Section-length overflow (sff.rs):** bar rounding now uses `saturating_add`, and the bar count is clamped to `u32::MAX / tpb` so the length stays a whole number of bars. Your repro is now `huge_deltas_do_not_overflow_section_length`: a Main A marker, then 20 × `FF FF FF 7F`. It panics on the old code in a debug build and passes now. I ran plain `cargo test` (debug) as well as `--release`.
- **Destination channel below 9 (parse_ctab):** such a destination now falls back to `src_ch | 0x08`, the source channel's accompaniment part. This is covered by `dest_channel_stays_on_accompaniment_parts`, and the `sff` exercise helper asserts that every parsed `dest_ch` is in 8..=15. A raw scan of all 461 Ctab/Ctb2 records in the corpus found none below 9, so real styles don't change.
- **exercise() asserts:** both helpers now assert that `transpose()` returns `None` or a note ≤ 127, and that `plays(r, Cancel) == is_drum_part(r.dest_ch)`.

**Declined / deferred**
- **TONES → chord_tones conflicts with #1-#6:** there's nothing to change on this branch. The only `TONES[` left is the fallback inside `chord_tones()` itself. Whoever merges the recognizer tickets should grep for `TONES[` and route new lookups through `chord_tones()`.
- **Env var to make a missing corpus a hard failure:** declined here. It changes a test convention shared by every corpus test in several files that other tickets are editing, so it fits better as its own small ticket. For this PR I checked the corpus tests ran: 35 styles, 469 channel rules.
- **live.rs:191 `held[k]` on the MIDI thread:** agreed that it's a real RT panic. It's out of scope for #7 (live input, and the file is being edited by sibling tickets), so it needs its own ticket; the fix is `& 0x7F`. I've noted it in the PR body.

The PR body's Decisions section now includes the destination-channel fallback.

Gates: `cargo build --release` passes. `cargo test --release` passes 26 with 1 ignored that was already ignored, and the corpus tests executed. `cargo clippy --release` shows no warnings in lines this PR touches.

ryangavin:
Rebased onto `integration/m1-chord-recognition` (now carrying #1–#6) and squashed to one commit (f30e66a).

**Conflicts (all in `src/theory.rs`; `sff.rs` merged cleanly):**
- `chord_tones` / `mask_of`: kept #2's `mask_of_set` helper and #7's `TONES.get(..).unwrap_or(TONES[0])`. TONES now has 38 entries, so the display-only ids 35–37 (M7b5 / (b5) / mM7b5) read their own tones instead of being treated as out of range. Only ids ≥ 38 fall back to a major triad.
- Recognizer scoring loop: took #2's rewrite. #7's edits there were to the old loop that #2 removed, and the new loop reads `FINGERED` through `mask_of_set`, so it never indexes TONES.
- `guitar_voicing`: kept #2's `casm()` + clamp and added #7's `bass % 12`.
- Tests: kept #4's 1+8/1+5/cancel tests alongside #7's.

**Coverage for the new ids:**
- No unchecked `TONES[` indexing is left. `mask_of` and `importance` go through `chord_tones`, and `plays()` still applies `casm()` before the Cancel check and the checked shifts, so Cancel plays only on drum parts, as #4 specified.
- The parse-time clamp is unchanged: any src_type ≥ 34 becomes Maj7. `source_chord_type_is_clamped` now also covers 35–38.
- `corpus_every_rule_every_chord`, `malformed_ctab_transposes_without_panicking` and `truncated_and_corrupted_files_do_not_panic` now iterate every chord type from 0 to 37. They also assert that 35–37 give the same `plays` and `transpose` results as their CASM type.

**Gates:** release build is OK. `cargo test --release` and debug `cargo test` both pass 161 with 1 ignored (`synth::loudness_probe`). The corpus tests ran on 35 styles and 469 channel rules. Golden and recognizer snapshots are unchanged. Clippy shows no new warnings in the touched code.

Issue #5, opened 2026-09-23
Chord detection area: Upper, and Manual Bass


ryangavin:
- Upper mode: the chord comes from the right hand, fingering becomes Fingered* (no 1+5, 1+8 or Cancel), Manual Bass is on, the style's Bass channel is muted, and the Left part plays bass.

---
Milestone: **M1 · Chord recognition 1:1** · Spec: [docs/genos-features.md](../blob/main/docs/genos-features.md) · Backlog: [docs/backlog.md](../blob/main/docs/backlog.md)

ryangavin:
Decisions made in #42 (implementation of Upper detection + Manual Bass). Playtest and overrule as needed:

- Decision: **Fingered\* needs at least 3 distinct pitch classes, and a Cancel/1+8/1+5 result is ignored so the previous chord holds.** Because: the Data List p.44 table shows 1+8 and 1+5 are the only Fingered shapes under 3 notes, and Fingered\* removes both. The effect is that right-hand melody notes and dyads never re-chord the Style.
- Decision: **Under Fingered\*, the bass is always the chord root (no slash chords from the right hand).** Because: the manuals say Fingered\* is "basically the same as Fingered", and Fingered uses the root as bass.
- Decision: **Upper mode uses the single existing split as Split Point (Left). The chord section is the keys above it.** Because: OM p.51 puts the chord section to the right of Split Point (Left). Separate split points come with #31.
- Decision: **Manual Bass makes the Left part sound, with the Style's Bass voice, even if the LEFT voice is off. Turning it off restores the user's Left voice and on/off state.** Because: OM p.51 says the Bass part's voice "is assigned to the left part", and the point is that the left hand plays the bass.
- Decision: **The Bass voice is played at the keys you play (only the Left part's own octave setting applies).** Because: the manuals specify no transposition.
- Decision: **Manual Bass also mutes Stop Accompaniment's bass note.** Because: it is the Style Bass part's voice.
- Decision: **With MIDI out only, Manual Bass mutes ch 11 and sends the left hand on ch 2, but sends no bass-voice program change.** Because: we can't restore an external device's Left voice afterwards.
- Decision: **Selecting Upper turns Manual Bass on each time (its documented default). In Lower, the setting is kept but has no effect.** Because: OM p.51 says Manual Bass defaults to On when Upper is selected and is only available in Upper mode.
- Decision: **The area is not stored in OTS (OTS stores neither fingering nor split).** Registration storage is left to #36.

Issue #3, opened 2026-09-23
Fingering modes: Single, Multi, Fingered, On Bass, AI Fingered, Full, AI Full


ryangavin:
- Make the fingering type a setting (§ Chord following spec). Implement:
  - Single Finger: white/black key below the root.
  - Multi Finger auto-detection.
  - Fingered: bass is the root.
  - Fingered On Bass: this is today's behaviour.
  - AI Fingered: infer from under 3 notes using the previous chord.
  - Full Keyboard and AI Full Keyboard: detect across the whole keyboard, with the documented limits (no Sync Stop, no 9th/11th/13th under AI Full).

---
Milestone: **M1 · Chord recognition 1:1** · Spec: [docs/genos-features.md](../blob/main/docs/genos-features.md) · Backlog: [docs/backlog.md](../blob/main/docs/backlog.md)

ryangavin:
## Decisions

The manuals leave these open (spec § C.2 marks the Multi Finger rule and the AI inference rule as unspecified). The owner may overrule any of them after playtesting.

- Decision: the default stays **Fingered On Bass**, because it is today's behaviour and changing it would be a regression outside this ticket.
- Decision: **Single Finger** takes the highest held key as the root. Any other black key below it gives m, any white key gives 7, and both give m7. There is no distance limit, and octave doublings of the root are ignored. This is because the manual says only "a black/white key to its left".
- Decision: **Multi Finger** reads Fingered first when 3 or more pitch classes form a recognized chord (not Cancel). Otherwise it uses the Single Finger reading when the keys are a Single Finger shape: one or two other keys, within an octave below the root, of different colours. There is one exception: a perfect fifth above the lower key stays Fingered 1+5. This is because C-Eb-G must stay Cm (not Gm7), while a root plus an adjacent key is unambiguous Single Finger.
- Decision: **Multi Finger has no Chord Cancel**, so C-Db-D reads as the Single Finger shape Dm7. This is because OM p.46 lists Cancel only for Fingered, On Bass and AI Fingered.
- Decision: **AI Fingered** reads 3 or more pitch classes as Fingered, with Cancel. With fewer than 3, it keeps the previous chord if every held note is one of its tones (or its bass). Otherwise:
[...]

Issue #1, opened 2026-09-23
Test harness: golden snapshots for recognizer and transposer


ryangavin:
- Build a table-driven test that covers every chord type in the Data List, in all 12 roots, in every inversion, and with the optional notes omitted.
- Add a snapshot mode to `yahaha sim`: a corpus style plus a chord script produces a per-part note list, saved under `tests/golden/`. Record the output only for our own scripts, and never commit style data.
- Show a diff on every change, so any change to chord behaviour is a deliberate, reviewed one.

---
Milestone: **M1 · Chord recognition 1:1** · Spec: [docs/genos-features.md](../blob/main/docs/genos-features.md) · Backlog: [docs/backlog.md](../blob/main/docs/backlog.md)

ryangavin:
Decisions made while building the golden harness (PR #45). The owner may overrule any of them after playtesting.

- Decision: recognizer expectations follow Fingered On Bass, where the lowest note becomes the on-bass note unless it is the root, because that is what the recognizer implements today. #3 adds plain Fingered, where the bass is always the root.
- Decision: when a pitch-class set spells more than one chord, the reading whose root is the lowest note wins. So C E G A = C6, A C E G = Am7, C G A = C6 (not Am7/C), and C Eb Gb A = Cdim7. I chose this because the Data List gives every shape in root position and says nothing about inversions, and Fingered On Bass treats the lowest note as the bass. Sets with no reading rooted on the lowest note, or with several, are covered by `ambiguous_voicings`, which accepts any valid reading until #2 sets the priority rules.
- Decision: the 3 uncoded rows (M7b5, (b5), mM7b5) only check root and bass for now, because #2 decides which CASM type they map to.
- Decision: in snapshot scripts, buttons press one tick before their slot, because a press exactly on a beat or bar line would otherwise depend on float rounding in the engine, which made fills start a beat later on some styles.
- Decision: snapshots list note starts and lengths but not velocities or pitch bends, because those would add noise until RTR Pitch Shift (#10) is real.
- Decision: `#` starts a script comment only at the start of a token, because `F#m7b5` must stay a chord.

PR #40, opened 2026-09-23
Keyboard and Master transpose feed the chord root and output (#6)

ryangavin:
## What

This adds Genos-style TRANSPOSE for live play (OM p.61, RM p.42).

- **Keyboard transpose (−12…+12):** the notes you play sound shifted, and so does the chord root and on-bass note sent to the style. The engine gets the chord as fingered and moves it by the Keyboard value (`shift_chord`). Because of this, changing Keyboard transpose while a chord is held moves the band straight away, and each sounding part follows its RTR rule, exactly as if you had played the new chord.
- **Master transpose (−12…+12):** everything pitched is shifted. That covers your notes and the style's output, which is shifted when each note is sent (after NTT/NTR, High Key and note limits), as the Genos does. Drum parts (ch 9/10) and any part whose voice is a drum or SFX kit (bank MSB 126/127) are left alone (`Prepared::kit`, precomputed at load). The chord the style follows is not changed.
- **Played notes:** the input thread sends each note at key + Keyboard + Master, to both the `yahaha` port and the built-in synth, since both are fed through `Out`. A new `Keys` table remembers the channel and note each held key sounded as. The note-off goes to that same place even if transpose or the split changed while the key was down, so notes can't get stuck. This also replaces the old "release on both channels in case the split moved" workaround.
- **CLI:** `--transpose N` (Keyboard) and `--master-transpose N`, each checked to be within −12…12.
- **TUI:** `;` / `'` set Keyboard −/+, `:` / `"` set Master −/+, and `/` resets both (the Genos resets when you press − and + together). The keys are listed on the help line. The chord line shows both values in a short `transpose kbd +2 master +0` readout. When Keyboard transpose is non-zero it also shows the fingered chord next to the chord the style follows, e.g. `Am7/G (fingered Gm7/F)`.
- **Real-time safety:** nothing new allocates, locks or can panic. The Keyboard + Master shift reaches the input thread through a single `AtomicI8`, and the engine gets `Cmd::Transpose` on the existing UI ring. The UI only changes the atomic after the command is on the ring, so if the ring is ever full nothing changes and played notes, the band and the display stay in step.

## How tested

- `sim::transpose::keyboard_transpose_moves_chord_root`: C7/E then Am with Keyboard +2 gives byte-for-byte the same output as fingering D7/F# then Bm. The snapshot shows `chord` = Bm and `played` = Am.
- `sim::transpose::keyboard_transpose_change_moves_held_chord`: holding C and moving Keyboard to +5 mid-bar gives exactly the same output as playing F at that moment.
- `sim::transpose::master_transpose_shifts_output_not_drums`: with Master −3, every pitched note-on/off matches the untransposed run shifted by −3, and drum-part events are unchanged. Timing and event count are identical.
- `sim::transpose::master_change_mid_note_releases_old_pitch`: Master and Keyboard changes while notes sound leave no note-on without its matching note-off.
- `sim::transpose::stop_accompaniment_follows_keyboard_transpose`: with the band stopped and Stop Accompaniment on, a transpose change re-sounds the bass and pad in the new key.
- `sim::transpose::stop_accompaniment_silent_stays_silent`: once the Stop Accompaniment notes have been silenced (start then stop), a Keyboard change moves the remembered chord but sounds nothing.
- `sim::transpose::master_transpose_skips_kit_voice_on_any_part`: a synthetic style with a bank MSB 126 voice on ch 14 marks ch 9/10/14 as kits; Master +5 shifts ch 13 and leaves ch 14 alone.
- `live::tests::input_transposes_notes_not_recognition`: drives `Input::key_msg` end to end. Notes sound transposed on their split channel, C-E-G is recognised as C (not D) with a +2 shift, releases go to the original channel and pitch after the split and shift change, `ChordReleased` fires only when the last chord key lifts, and a retrigger stops the old note first.
- `live::tests::keys_release_where_they_sounded`: notes are released at the pitch they started at, retriggers are handled, and notes beyond the MIDI range fold back by an octave.
- `corpus_full_performance_no_stuck_notes` now changes Keyboard and Master transpose three times mid-performance across all 35 corpus styles. It still requires zero unbalanced notes and no bent parts at the end.
- Gates: `cargo build --release` and `cargo test --release` pass (28 passed, 1 ignored as before; the corpus and soundfont tests ran against the symlinked corpus). `cargo clippy --release` shows no new warnings: its warning set is identical to the base branch. I also checked the UI with `yahaha screen`.

## Decisions

- Decision: Master transpose shifts the style's output notes after chord transposition and does not move the chord root, because the manual says Master "transposes the pitch of the entire sound" while only Keyboard is described as including "the chord root for triggering Style playback". Shifting the output keeps High Key and note-limit octave placement and Bypass channels the same as the Genos would.
- Decision: the chord display shows the chord the style follows (fingered chord + Keyboard transpose), with the fingered chord shown next to it when they differ, because Keyboard transpose is defined as changing the chord that triggers the style. Master transpose does not change the displayed chord.
- Decision: a transpose change applies to notes started afterwards. Notes already sounding (yours and the style's) keep their pitch and are released at it, until the next chord change revoices the style's notes at the new setting, because Yamaha tone generators apply transpose at note-on, and doing it this way avoids re-attacks and stuck notes.
- Decision: a Keyboard transpose change while a chord is held (or remembered) moves the style's chord right away, and parts follow their RTR rules, because Keyboard transpose is defined in terms of the keys and the held keys now stand for the new chord. Otherwise the band would stay in the old key until the next chord.
- Decision: with the band stopped and Stop Accompaniment on, a Keyboard change re-sounds the bass and pad in the new key only if they are still sounding, even with the keys released. Once they have been silenced (e.g. by Start/Stop), the remembered chord moves but nothing sounds until the next chord, because a transpose button press should not start sound that was not already playing.
- Decision: chord recognition runs on the fingered keys and the engine shifts the result. This is equivalent to recognising the transposed keys, since recognition doesn't depend on key, and it is what makes the previous decision possible.
- Decision: drum/SFX kit exclusion for the style is decided per part at load time from the part's initial voice (drum parts ch 9/10, or bank MSB 126/127), because the manual says transpose never applies to kits and the engine must not look anything up at play time. A kit selected mid-pattern by a later program change is not tracked.
- Decision: keyboard (RH/LH) notes are always transposed, because the built-in synth's keyboard slots only hold melodic GM programs today. If kit voices are added to the keyboard parts, they should be excluded the same way.
- Decision: Song transpose is not implemented, because yahaha has no MIDI song player, so it has nothing to act on in live play. "Transpose MIDI Input" behaves as On, because our keyboard reaches us over MIDI and transposing it is the whole point.
- Decision: each target is limited to −12…+12, independently, as on the Genos. Keyboard + Master can add up to ±24 on played notes, and any note pushed past 0…127 folds back by an octave.
- Decision: the TUI keys are `; ' : " /`, because they were free and sit together on the keyboard. The owner may prefer other keys, or mapping them to the Launchkey later.

Closes #6
Part of #6


ryangavin:

Checked out `m1/6-transpose` at c20f749, symlinked the corpus and soundfonts, and ran the gates myself:
- `cargo build --release`: OK.
- `cargo test --release`: 25 passed, 1 ignored. The corpus tests really ran: no "no corpus; skipping" line with `--nocapture`, and 35 styles were loaded.
- `cargo clippy --release`: exactly the same 31 warnings as `integration/m1-chord-recognition`, so nothing new.

### Verified against the manuals
- RM p.42 / OM p.61 say: Master "transposes the pitch of the entire sound"; Keyboard transposes "including the chord root for triggering Style playback"; range −12…+12; "Transpose is not applied to the Drum Kits and SFX Kits Voices". The implementation follows all of this. Kit detection by bank MSB 126/127 is right: the Data List puts Drum kits (incl. Revo!Drums) at MSB 127 and SFX kits (incl. Revo!SFX) at MSB 126.
- The Master-after-NTT/limits decision is sound. It matches tone-generator-level transpose, and `master_transpose_shifts_output_not_drums` checks it byte for byte.

### Throwaway probes I ran (not committed)
- **All 35 corpus styles, 23 steps each.** Chord changes were interleaved with Keyboard/Master changes: Keyboard −3…+3 and Master +12 down to −11, applied mid-note and through RTR revoice. I then called `stop()`. Result: note-on/off stayed balanced per (channel, note) for every style, so no stuck notes.
- **Keyboard 0 → +12 with a chord held while running.** Output is byte-identical to no change. That is correct, because the chord root is a pitch class, so the band doesn't jump an octave.
- **Sync Start state with a remembered chord, then Keyboard change.** Nothing sounds and the band doesn't start. Correct.
- **RT paths:** `Keys` is a fixed 128-entry array on the input thread, and the shift is read through `AtomicI8`. `shift_key`'s loops are bounded because |d| ≤ 24. `master()` is a table lookup. No allocations, locks or panics were added on the engine or input threads.

### Blocking
None found.

### Nonblocking
1. **`src/ui.rs` `set_transpose` closure:** `key_shift` is stored before `tx.push(Cmd::Transpose)`, and the push result is thrown away (`let _ =`). If the UI ring is full, played notes use the new shift but the engine (style chord and Master) keeps the old one. The screen shows the engine's snapshot, so it would not show the mismatch. Suggestion: store `key_shift` only when the push succeeds, or retry the push.
2. **The kit exclusion for non-drum channels (`Prepared::kit`, MSB ≥ 126) is never tested.** No corpus style has a kit outside dest 8/9 (I checked all 35), so that branch has no coverage. A small unit test building `Prepared` from a synthetic init with a Bank MSB 127 and Program Change on a phrase channel would cover it.
3. **Keys don't go through `Input::key_msg` in any test.** `Keys` is tested on its own, but nothing covers the split + `key_shift` + retrigger path end to end through `Out`. Consider a test that feeds raw note messages to `Input` and checks what comes out.
4. **Stop Accompaniment re-attacks on a transpose change after the keys are released.** `set_transpose` re-sounds the remembered chord's bass and pad when `stop_acmp` is on. That fits the existing "stop-acmp notes persist" behaviour, but it gives a fresh attack without a key press. It may be worth a line under Decisions so the owner can playtest it.
5. **A Master change followed by a chord change moves notes that were already sounding.** Notes sustained across the change get re-pitched to the *new* Master on the next revoice (`revoice` → `self.master(...)`). That is consistent and has no stuck notes. The Decisions only say sounding notes "keep their pitch", so it's worth adding "until the next chord change re-voices them".
6. **`/` resets both targets.** On the Genos, −/+ together resets the target the buttons are currently assigned to. This is already under Decisions and is fine for now.
7. **Chord-line width.** The transpose text is appended to the chord line (about 80 extra columns) and will be cut off on narrow terminals. It's cosmetic.
8. **Merge-conflict surface.** The `Snapshot` literals in `ui.rs` (`default_snap`, `screen_html`) and the `key_msg` rewrite in `live.rs` are likely to conflict with sibling tickets that touch the split, the input path or the snapshot. Whoever merges second should expect to rebase.

Verdict: no blocking issues. The behaviour matches the manual, the tests check behaviour (equivalence against fingering the transposed chord, and an exact shift with drums untouched), and the corpus stuck-note test now exercises transpose.

ryangavin:
Round 1 review addressed in a8ca185.

**Addressed**
- `set_transpose` ordering: the UI now pushes `Cmd::Transpose` first and only stores `key_shift` (and updates its own value) if the push succeeded. With a full ring nothing changes, so played notes, the band and the display can't disagree.
- `Prepared::kit` MSB ≥ 126 branch: new `sim::transpose::master_transpose_skips_kit_voice_on_any_part` builds a synthetic style with a bank MSB 126 voice on ch 14. It checks the kit table is ch 9/10/14, and that Master +5 shifts ch 13 and leaves ch 14 alone.
- End-to-end input: new `live::tests::input_transposes_notes_not_recognition` drives `Input::key_msg`. It captures output through the synth ring and covers split channels, the shifted notes, recognition on the fingered keys, releases after the split and shift change, `ChordReleased` timing and retrigger.
- Stop Accompaniment re-attack: a Keyboard change now re-sounds bass and pad only if the Stop Accompaniment notes are still sounding. Once they have been silenced, the chord moves but stays quiet (`stop_accompaniment_silent_stays_silent`). This is recorded under Decisions.
- Decisions wording: sounding notes keep their pitch "until the next chord change revoices the style's notes at the new setting". The `set_transpose` doc comment is updated to match.
- Narrow terminals: the chord-line readout is now `transpose kbd +2 master +0` (about 30 columns), and the key hints moved to the help line.

**Declined**
- `/` resets both targets: kept as already recorded under Decisions. There is no per-target "assigned" selection in the TUI yet, so reset-both is the closest single-key equivalent.
- Merge conflicts with sibling tickets (`Snapshot` literals, the `key_msg` rewrite): these come with the feature itself. I'll resolve them when the integration branch moves.

Gates: `cargo build --release` passes. `cargo test --release` gives 28 passed, 1 ignored, and the corpus tests ran on all styles. `cargo clippy --release` shows no new warnings in touched code.

ryangavin:
Integrator: I rebased this onto `integration/m1-chord-recognition` (after #1–#5) and squashed it to one commit.

**How I resolved the conflicts**
- `engine.rs`: `set_chord` now takes the fingered chord and shifts it by Keyboard transpose (from this PR). It keeps `prev` so #4's `catch_up` still brings in parts that the previous chord kept silent. `set_transpose` calls `catch_up` after `revoice` too, so a Keyboard change while running behaves exactly like playing the new chord, which is what this PR says it should do. `note_on` keeps #5's Manual Bass mute of the Bass part first, then applies the Master shift.
- `live.rs`: #5's `route[128]` (which side each held key is on, used for chord-section membership, Full Keyboard, Upper/Fingered* and Sync Stop) now sits next to this PR's `Keys` table (the channel and transposed note each key sounded as). Note-offs go through `Keys`, so the old route-based releases on both channels and the "release on both channels in case the split moved" fallback are gone. A retrigger now releases the old note first even on the same side, as this PR intended. All of the fingering, Upper and Full Keyboard logic from #3 and #5 is unchanged.
- `main.rs`, `ui.rs`, `sim.rs`: both sides kept. That covers the CLI flags `--fingering`/`--upper`/`--no-manual-bass` and `--transpose`/`--master-transpose`, the `Options` fields, the TUI keys `f d D` and `; ' : " /`, and the `Step::ManualBass`/`Step::Transpose` variants (both marked as test-only in the script grammar).

**Gates**
- `cargo test --release`: 154 passed, 1 ignored (the existing synth loudness probe). The corpus and golden tests ran with no skips.
- Golden snapshots are unchanged, so nothing needed regenerating, and ticket #1 left no `#[ignore = "M1 #6"]` tests.
- `cargo clippy --release` gives the same 29 warnings as the integration base, with no new ones.

PR #41, opened 2026-09-23
Fingering types: Single, Multi, Fingered, On Bass, AI Fingered, Full, AI Full (#3)

ryangavin:
## What

Makes the chord fingering type a setting (spec § Chord following, RM p.9, OM p.46), with all seven Genos types:
Single Finger, Fingered, Fingered On Bass (the default, today's behaviour), Multi Finger, AI Fingered, Full Keyboard, AI Full Keyboard.

- **`src/fingering.rs`** (new): the `Fingering` enum and `detect()`, a dispatch layer around the existing recognizer. All recognition of 3+ pitch classes goes through one function (`fingered()`), which calls `Recognizer::recognize`. Everything below three pitch classes (Single Finger, Multi Finger dyads, AI inference, 1+5/1+8 in AI) is decided in `fingering.rs`, so it doesn't depend on the table. The recognizer's internals are unchanged, and the tests pass both on this branch and on a local merge with #44. **Merge note for #44:** `live.rs::recompute` conflicts; keep this branch's version, but whichever of #41/#44 lands second **must** also have `fingered()` call `rec.recognize_keys(s.mask, s.low % 12, s.keys, true)` and `plain()` call it with `on_bass = false` (`Scan.keys` is already there), and add tests that On Bass and Fingered `[36, 48]` give C1+8. Taking this branch's `recompute` without that switch silently loses #44's 1+8 in the default mode.
- **Merge note for #42 (Upper detection):** #42 also rewrites `live.rs::recompute`. RM p.9: only Fingered* is available when the detection area is Upper, so in the merged `recompute`, when `shared.upper` is set, bypass `fingering::detect` and use #42's path (On Bass recognition over the right hand + `fingered_star`); `detect()` runs only in Lower.
- **Setting**: `--fingering single|multi|fingered|on-bass|ai|full|ai-full` on `yahaha play`. In the TUI, `[f]` cycles through the types in Genos screen order. The status line shows `fingering <name> [f]`.
- **Input thread** (`live.rs`): the type lives in `Shared.fingering` (an `AtomicU8`). Input now also tracks `held_all`, the keys held anywhere on the keyboard. In the Full types, right-hand notes also trigger recognition. Detection works on the held-key array in place: no allocation, no locks.
- **Documented limits**:
  - Full and AI Full Keyboard: Sync Stop is turned off and cannot be turned back on. The engine thread reads `Shared.fingering` on every wake and calls `Engine::allow_sync_stop`, so the TUI key and the Launchkey pad are both refused (no command is involved, so a full ring can't leave Sync Stop on). The TUI shows `SYNC STOP n/a`.
  - Chord Cancel is only recognized in Fingered, On Bass and AI Fingered.
  - AI Full Keyboard cannot play 9th, 11th or 13th chords.

## How tested

- New unit tests in `fingering.rs`:
  - Single Finger shapes (M / m / 7 / m7, root doubling, Bb-C-C').
  - Fingered vs On Bass (C/E vs C, C/F# vs no chord) and Cancel.
  - Multi Finger: Fingered wins for 3+ notes (Cm, G7, the Cm7 shell), lone root and doubled root → major, 1+5, dyads, no Cancel (C-Db-D → Dm7), and the canonical Single Finger M / m / 7 / m7 shapes on all 12 roots.
  - AI Fingered: keeps the previous chord, lone note → major / 1+8, dyad inference, 1+5 both ways, tritone resolved by the previous chord (F-B over C → G7, Bb-E over F → C7), foreign bass → no chord.
  - Full Keyboard: left-hand bass + right-hand chord, foreign bass (C/F#), left-hand chord + a right-hand melody note that is a chord tone or not (A, Bb, B, C#, D, F, F#, G, C all stay C), right-hand-only chords with a root-moving top note, melody alone, no Cancel.
  - AI Full: C9 → C7, Cadd9 → C, left-hand chord + melody, left-hand dyad + melody inferred, a lone melody note makes no change.
  - Engine test (uses the corpus): Full types turn Sync Stop off and block it.
  - Live path (`live.rs`): `Input::key_msg` ignores right-hand notes outside the Full types, reads both hands in Full, and clears `held_all` on note-off, including 0x90 with velocity 0.
- Round 2 follow-ups: D-F-C (AI Full → Dm7) vs D-F-E (→ Dm), E-G-B-C → Em and A-C-E-F → Am in Full, C-G | E-Bb → C7 in Full, Multi Finger F-G / D-G / E-G → G7.
- `cargo build --release`, `cargo test --release`: 30 passed, 1 ignored (77 passed on a local merge with #44). The corpus was symlinked and `corpus_full_performance_no_stuck_notes` ran on all 36 MOX_v2 styles.
- `cargo clippy --release`: no new warnings in touched code (the `live.rs` `recompute` collapsible-if predates this PR).
- `yahaha screen` render checked: the status line fits at 150 columns.

## Decisions

The manuals leave these open (spec § C.2 marks the Multi Finger rule and the AI inference rule as unspecified). The owner may overrule any of them after playtesting.

- Decision: the default stays **Fingered On Bass**, because it is today's behaviour and changing it would be a regression outside this ticket.
- Decision: **Single Finger** takes the highest held key as the root. Any other black key below it gives m, any white key gives 7, and both give m7. There is no distance limit, and octave doublings of the root are ignored. This is because the manual says only "a black/white key to its left".
- Decision: **Multi Finger** reads 3 or more pitch classes as Fingered (not Cancel, bass on the root), so C-Eb-G stays Cm (not Gm7) and the C-Eb-Bb shell stays Cm7. The exception: when the Fingered reading leaves out a chord note and the keys are a close Single Finger shape (a black and/or white key within a minor third below the root), it is Single Finger, so Bb-C-C# is C#m7, not Bbm(add9) without its fifth. Otherwise a Single Finger shape (at most one black and one white pitch class, each within an octave below a held root key) is Single Finger. With 2 pitch classes, a perfect fifth up from the lowest key is 1+5 (C-G); any other dyad is Single Finger (G-C → C7, Bb-C-C' → Cm). One pitch class, doubled or not, is the major chord, not 1+8. The cost: the Single Finger G7 shape G + C (the white key a fifth below) reads C1+5; G7 still works as F-G, E-G or D-G. This is because RM p.9 gives "press the root key only" for Single Finger major, and Multi exists to accept Single Finger fingering.
- Decision: **Multi Finger has no Chord Cancel**, so C-Db-D reads as the Single Finger shape Dm7. This is because OM p.46 lists Cancel only for Fingered, On Bass and AI Fingered.
- Decision: **Fingered** (and AI Fingered's 3+ notes) is the On Bass reading with the bass on the root, and a bass note that is not a chord tone makes no chord (F#-C-E-G → no chord; On Bass gives C/F#). This is because Fingered reads the root as the bass (RM p.9). Once #44 lands this becomes `recognize_keys(on_bass = false)`.
- Decision: **AI Fingered** reads 3 or more pitch classes as Fingered, with Cancel. With fewer than 3, it keeps the previous chord if every held note is one of its tones (or its bass). Otherwise:
  - Two notes are inferred from the interval: m3 → m, M3 → major, m6 → major 1st inversion, M6 → minor 1st inversion, m7 → 7, M7 → maj7, 5th → 1+5, 4th → 1+5 on the upper note. A 2nd is a passing note and keeps the chord.
  - A tritone is the 3rd and 7th of a dominant 7th, which fits two roots a tritone apart. It takes the root nearer the previous chord on the circle of fifths (F-B over C → G7, Bb-E over F → C7); with no previous chord or a tie, the lower note is the 3rd (B-F → G7).
  - One pitch class is the major chord on it, or 1+8 when held in more than one octave. This is because a single key can only name a root, and AI Fingered is "basically the same as Fingered", which has 1+8.

  This is because the RM says "based on the previously played chord", and keeping the chord when the notes fit it is the most conservative reading.
- Decision: **Fingered, Multi Finger and AI Fingered drop the bass**. The root is always the bass, as RM p.9 says for Fingered. Only On Bass and the Full types produce slash chords.
- Decision: **Full Keyboard** reads with On Bass rules (the lowest key is the bass) and needs 3 or more pitch classes, so a melody alone never changes the chord. Melody is set aside by rules that don't depend on the chord table:
  1. A chord at or below the split (3+ pitch classes, not over a foreign bass) is the chord, and every right-hand note is melody. So a left-hand C triad stays C under any right-hand melody note, chord tone or not (A does not make C6, Bb does not make C7). This is the RM's own example: "a chord with your left hand and a melody note with your right".
  2. Otherwise the whole keyboard is read, dropping the highest key while the rest is no chord, or while the rest is a chord on a different root. So right-hand C-E-G with C# on top stays C, and C-E-G-B stays Cmaj7.
  3. A chord over a bass that is not one of its tones (C/F#) is used only if no reading is a plain chord or inversion, unless the top key is needed for it (the reading without it is a chord on the same root). So left-hand F# with right-hand C-E-G is C/F#, not C(b5)/F#.

  The cost of rule 2: a right-hand-only first-inversion maj7 with the root on top (E-G-B-C, A-C-E-F) reads as the minor triad below it (Em, Am), because the top key moves the root and is taken as melody. And a top key that keeps the root stays in the chord: left-hand C-G with right-hand E-Bb is C7, which is why left-hand C-G + right-hand E + a melody note on top changes the type with the melody (A → C6, B → Cmaj7, D → Cadd9); the root never moves.

  Full Keyboard needs 3 pitch classes, so it never produces 1+5 or 1+8. This is a deliberate departure from "similar to Fingered", so a two-note melody interval never re-chords the band.

  The cost of rule 1: a left-hand triad plus right-hand tensions reads as the triad. A pianist voicing the whole chord in the right hand, or putting only the bass in the left, gets the full chord.
- Decision: **AI Full Keyboard** is Full Keyboard, else AI inference on the lowest two pitch classes after the melody is dropped from the top (left-hand C-E + right-hand D → C). A single note is treated as melody and never changes the chord, because otherwise every right-hand melody note would re-chord the band.
- Decision: under **AI Full Keyboard**, 9th, 11th and 13th chords are reduced to the chord without the tension rather than rejected: 9/13/7b9/7#9/7b13/7#11 → 7, maj9/maj7#11 → maj7, m9/m11 → m7, add9 → major, m(add9) → m, 6/9 → 6, mMaj9 → mMaj7. This is because "cannot be played" should still follow the harmony the player means.
- Decision: in the Full types, **the split still decides which voice the notes sound on** (left-hand or right-hand channel). Only chord detection spans the keyboard.
- Decision: **changing the fingering type while keys are held** takes effect on the next key event, not immediately. This is because the input thread only runs on MIDI input, and re-reading the chord on a settings change would re-chord the band mid-bar without a new key. The player just re-strikes.

Closes #3


ryangavin:

I checked out `2cc36e0` with the corpus symlinked and ran the gates. `cargo build --release` passes. `cargo test --release` gives 27 passed and 1 ignored, and the corpus tests actually ran (MOX_v2 is present, and `full_keyboard_disables_sync_stop` did not skip). `cargo clippy --release` shows no new warnings in the touched code; the `map_or(false, …)` hits in `ui.rs` follow the existing pattern.

I checked the manual claims against the text: RM p.9 fingering table, OM p.46 (Cancel is only for Fingered, On Bass and AI Fingered), OM Sync Stop note (not with Full or AI Full), and RM "9th, 11th and 13th chords cannot be played" for AI Full. All four are accurate. The input thread is RT-safe: `detect`/`scan` are stack-only, `from_u8` can't go out of range, and `to_u8`'s `unwrap` is infallible. The dispatch-layer shape is right.

To check how this branch integrates with sibling #44 (the Data List table rewrite, which is what "call into the existing recognizer" is meant to be robust against), I made a throwaway local merge of `origin/m1/2-chord-table` into this branch. I resolved the `live.rs` conflict with this PR's side and ran `cargo test --release fingering`. **3 of this PR's 8 tests fail**. That merge is where these two tickets meet, so I'm treating the root causes as blocking here.

### Blocking

1. **Multi Finger's "root only = major" depends on a recognizer quirk that #44 removes.** `src/fingering.rs:221` sends a lone key (n == 1) to `fingered()`. It returns major only because the current `theory::compute` maps one pitch class to major, and that is not a Data List Fingered chord. #44 makes a single key "no chord, keep the previous chord". After the merge, `multi_finger` fails at `fingering.rs:358` with `name(m, [48]) = "-"`, but it should be `"C"`. Single Finger's root-only rule is spec'd (RM p.9: "Press the root key only"), and Multi explicitly accepts Single Finger fingerings. So Multi must call `single()` for n == 1 itself and not borrow Fingered's behaviour.
   - The same applies to root plus its octave (`[36,48]`): Single Finger says C, while #44 makes Fingered say C1+8. Pick one and write it down under Decisions.
   - The `s.n <= 2` fallback at line 221 also relies on today's dyad inference (C+E=C, C+Eb=Cm, …), and #44 deletes that too. Decide what Multi does with non-shape dyads without assuming it.

2. **AI Fingered's lone-note rule depends on the same quirk.** At `fingering.rs:237`, `1 if single_note => fingered(rec, s)` is documented in the PR as "One note falls back to Fingered" and tested as `[50]` over Am7 giving `"D"`. After the merge it gives `"-"` (the test fails at `fingering.rs:370`). AI Fingered is the mode whose whole point is <3-note inference (RM p.9: "less than three notes can be played to indicate the chords (based on the previously played chord, etc.)"). So the lone-note rule has to be written in `ai()` explicitly, for example "a lone note not in the previous chord → major on that note". It should not depend on whatever Fingered happens to do today. Update the Decision accordingly.

3. **Full Keyboard's melody handling is unstable. Melody notes that are tones of some chord re-chord the band, and the only test is cherry-picked.** `full()` (`fingering.rs:265`) returns the first reading over *all* held keys before it drops any melody notes. The test only uses melody notes that make no chord (Db, F#). With a left-hand C triad `48,52,55` and one right-hand melody note, on this branch:
   - melody A (69) gives **C6**
   - B (71) gives **Cmaj7**
   - Bb (70) gives **C7**
   - D (74) gives **Cadd9**

   After the #44 merge, the PR's own test case (melody C#, 73) gives **C#mMaj7b5/C**, and `full_keyboard` fails at `fingering.rs:387`. So every melody step re-chords the style, and the result depends on the table. The RM example is "a chord with your left hand and a melody note with your right". It is at least as reasonable to read that as saying the melody should not disturb the chord. Needed:
   - A rule that doesn't depend on table accidents. For example, prefer a reading with no melody notes when the rest is already a ≥3-note chord. Or at least don't let a single note above an already-recognized chord turn it into a chord with a different root.
   - Tests that cover chord-tone melody notes and non-chord-tone melody notes, not just non-chord ones.

   If you keep "melody tones extend the chord" (C+Bb gives C7), keep it on purpose, record it as a Decision, and test it. It must still not produce the root change above.

### Nonblocking

4. **Plain Fingered is On Bass with the bass removed** (`fingering.rs:161`). With F# C E G it gives C, because today's recognizer has an on-bass fallback. A Genos in Fingered (not On Bass) has no slash reading for a foreign bass. #44 adds `recognize_keys(mask, low, keys, on_bass)` for exactly this, and it also handles 1+8, which needs a key count that `Scan` doesn't track. Plan to switch the `fingered()` wrapper to it once #44 lands. Adding a key count to `Scan` now would make that a one-line change.

5. **The AI dyad inference barely uses the previous chord.** Tritone at `fingering.rs:255`:
   - B–F over C gives G7, which is good.
   - F–B over C gives **C#7**, but the same G7 shell voiced F-under-B is just as common, and in C it is G7.
   - E–Bb over F gives C7 (correct), but Bb–E gives F#7.

   The RM says "based on the previously played chord". Use `prev` to choose between the two tritone-related dominants, for example the one whose root is closer to prev on the circle of fifths. Likewise, E–G with no previous chord gives Em, and with prev = C it keeps C, which is fine.

6. **AI Full drops the AI inference after melody drops.** With left-hand C–E plus a right-hand melody F#, `full()` drops F# and is left with n == 2, so the loop exits. `ai(s, …)` is then called on the *original* 3-note scan, gets `fingered` → None, and the chord doesn't change. The 2-note rule never sees the left-hand dyad. Consider running `infer_dyad` on the reduced scan.

7. **Multi Finger canonical m7 shapes on C# and F# misread** (current tables):
   - Bb–C–C# (`46,48,49`) gives Bbm(add9), where Single Finger gives C#m7.
   - Eb–F–F# (`51,53,54`) gives Ebm(add9), where Single Finger gives F#m7.

   These are table issues, since m(add9) is 1+2+b3+5 with no omittable 5th (DL p.45), and they should go away once #44 lands. Please add a test that loops all 12 roots through the Single Finger M / m / 7 / m7 shapes (nearest keys to the left) in Multi, so that stays locked in.

8. **Edge case:** Multi with `46,48,60` (Bb, C, and C an octave up) gives `-`, but Single Finger gives Cm. The octave-doubled root is taken as `high`, which puts the black key more than 12 semitones below it. Measuring from the lowest root key would fix it.

9. **`let _ = ch.ui_tx.push(Cmd::SyncStopAllowed(..))`** (`ui.rs:129`, `ui.rs:375`). If the ring is full, the command is dropped. The fingering switches to Full, but the engine keeps Sync Stop on and the UI shows "n/a". Retry the push, or have the engine read `shared.fingering` directly.

10. **No live-path test.** Nothing drives `Input::key_msg` to show that right-hand notes are ignored in non-Full modes and included in Full modes, or that `held_all` is cleared on note-off (including running-status `0x90 vel 0`). `sim.rs` also can't select a fingering type, so there is no offline check of the modes against styles.

11. **Switching fingering type while keys are held doesn't recompute.** The new mode only takes effect on the next key. That's probably fine, but it's worth a line in Decisions.

**Verdict: changes requested (blocking 1–3).** 1 and 2 are small, local fixes: make Single Finger and AI own their <3-note rules. 3 needs a deliberate melody rule and honest tests.

ryangavin:
## Round 1 review: changes

Two commits are pushed: f3f7877 and a87b1fd. Every test passes on this branch (30 passed, 1 ignored, with the corpus tests running). On a local merge with #44 (`m1/2-chord-table`) 77 pass. That merge has one conflict, in `live.rs::recompute`; keep this branch's side of it.

### Blocking: all fixed
- **Multi Finger, lone key:** a single pitch class now calls `single()` directly, so `[48]` gives C. Root plus octave (`[36,48]`) also gives C (Single Finger major), not 1+8; see Decisions. Dyads no longer go through the recognizer: a perfect fifth up from the lowest key is 1+5 (built explicitly), and every other dyad is Single Finger.
- **AI Fingered, fewer than 3 notes:** the rules are now written out in `ai()` and `infer_dyad()`, and the recognizer is not called for fewer than three pitch classes. A lone note gives the major chord on it, or 1+8 when it is doubled. 1+5 is built explicitly in both directions. `[50]` over Am7 gives D on both branches. The Decision is updated.
- **Full Keyboard melody:** a new rule that does not depend on the chord table:
  1. A left-hand chord (3+ pitch classes, at or below the split) wins, and the right hand is melody.
  2. Otherwise, the top key is dropped while the rest is no chord, or while the rest is a chord on a different root.

  A test covers a left-hand C triad under the melody notes A, Bb, B, C#, D, F, F#, G and C, and every case stays C on both branches. Further tests cover a right-hand-only C-E-G with C# or F# on top (gives C) and C-E-G-B (gives Cmaj7).

### Nonblocking: addressed
- **Plain Fingered:** the On Bass reading is now rejected when the bass is foreign to the chord. F#-C-E-G gives no chord in Fingered and C/F# in On Bass. `Scan` now also carries a key count.
- **AI tritone:** the rule now picks the dominant nearer the previous chord on the circle of fifths. F-B over C gives G7, and Bb-E over F gives C7.
- **AI Full with a dyad left over:** when Full finds no chord, the lowest two pitch classes, with the melody dropped, go to keep or infer. Left-hand C-E with a right-hand melody note gives C, or keeps the previous chord when that chord contains both notes.
- **Multi, Single Finger m7 shapes in all 12 roots:** a new test, `multi_finger_single_shapes_all_roots`, checks M, m, 7 and m7 for all 12 roots. On this branch, C# and F# read as m(add9) without its fifth. To fix that without waiting for #44, a Fingered reading that leaves out a chord note now loses to a close Single Finger shape (black and white keys within a minor third below the root). The C-Eb-Bb shell still reads as Cm7.
- **Multi, distance from a doubled root:** a key's distance is now measured from the nearest held root key above it. `[46,48,60]` gives Cm.
- **Sync Stop dropped by a full ring:** `Cmd::SyncStopAllowed` is removed. The engine thread now reads `Shared.fingering` on every wake and calls `allow_sync_stop`, so there is no message left to drop.
- **Live-path tests:** new `live::tests` cover right-hand notes being ignored in On Bass, Fingered, AI and Multi, and read in Full. They also check that `held_all` and `held` are cleared on 0x80 and on 0x90 with velocity 0.
- **Switching the fingering type while keys are held:** recorded as a Decision. The new type takes effect on the next key event.

### Declined
- **Switching to `recognize_keys(on_bass=false)`:** declined for now because that function does not exist until #44 lands. The foreign-bass rejection above gives the same behaviour in the meantime. The PR body has a merge note: `fingered()` and `plain()` should call `recognize_keys(s.mask, s.low % 12, s.keys, on_bass)` so that Fingered and On Bass get 1+8.
- **Selecting a fingering type in `sim.rs`:** declined. `sim` takes chords directly and never runs recognition, so a fingering option would have nothing to act on. The live-path tests above cover the input thread instead.

### Also changed
- **Full Keyboard with a foreign bass:** when the chord over a foreign bass needs the top key, that key is kept rather than dropped as melody. #44's table reads F#-C-E as C(b5), so without this, left-hand F# plus right-hand C-E-G gave C(b5)/F#. It now gives C/F#.

The PR body's Decisions section is updated for Multi, Fingered, AI Fingered, Full, AI Full and switching the type while keys are held.

ryangavin:

I checked out `a87b1fd` (detached, because the branch is held by another worktree), symlinked the corpus and ran the gates.
- `cargo test --release`: 30 passed, 1 ignored. The corpus was present: `corpus/MOX_v2/FunkyFinger.S930.STY` exists, so `full_keyboard_disables_sync_stop` really ran.
- `cargo clippy --release` shows no new warnings in touched code. `live.rs:164` (collapsible if) predates this PR, and `ui.rs:587` is an existing `map_or` line that was only moved.
- I made a throwaway merge with `origin/m1/2-chord-table` (#44), taking this branch's side of the `live.rs::recompute` conflict: 78 passed, 0 failed.

I also wrote a throwaway probe test with about 55 inputs across all seven types and ran it on both this branch and the merge. The outputs are quoted below. I re-checked the manual claims against RM p.9 and the OM Fingering/Sync Stop notes, and they are accurate.

### Round-1 blocking items: verified fixed
1. **Multi Finger, lone key and doubled root.** `[48]` gives C and `[36,48]` gives C on both branches. The dyad path no longer calls the recognizer.
2. **AI Fingered with fewer than 3 notes.** `[50]` over Am7 gives D and `[38,50]` gives D1+8 on both branches. `[36,48]` gives C1+8 after the merge.
3. **Full Keyboard melody.** A left-hand C triad under right-hand A, Bb, B, C#, D, F, F#, G or C stays C on both branches. Am7 in the left hand with a right-hand G stays Am, and G7 with a C melody stays G7. The root-changing top-note rule also works: right-hand C-E-G-C# gives C, and C-E-G-F# gives C. The same-root extension (C+Bb gives C7 when it's all in the right hand) is now a deliberate, recorded and tested Decision, which is what round 1 asked for.

The round-1 nonblocking fixes also check out:
- Tritone rule: F-B over C gives G7, and Bb-E over F gives C7.
- Multi Finger on all 12 roots.
- `[46,48,60]` gives Cm.
- Sync Stop is now read from `Shared.fingering` on every engine wake, so no ring command can be lost.
- Live-path tests cover right-hand notes being ignored or read, and `held_all` being cleared on 0x80 and on 0x90 with velocity 0.
- The declined items (`sim.rs` fingering and `recognize_keys` before #44) are declined for sound reasons.

### Real-time safety
The `detect` path is stack-only. The only `unwrap` is `to_u8`, which can't fail. `from_u8` wraps with `%`. Indexing in `single_shape`, `held[(k + up)]`, stays at or below `s.high`, because `k + up` is the next key above `k` with the root's pitch class and `high` is one. `full()` does at most (held keys × 2) scans of 128 entries per event, which is fine. `Fingering::parse` allocates, but only on the CLI path.

### Blocking
None.

### Nonblocking
1. **Merge-order hazard: the default mode loses #44's 1+8.** The PR tells whoever merges to keep this branch's `recompute`. On the throwaway merge, that makes the default Fingered On Bass `[36,48]` give **`-`**, where #44's live path gives **C1+8**. Fingered `[36,48]` also gives `-`, because `fingered()` still calls `recognize()`, which assumes one key per pitch class. The merge note in the body says what to do, but no test would catch it if someone forgets. Whichever of #41/#44 lands second should switch `fingered()` and `plain()` to `recognize_keys(s.mask, s.low % 12, s.keys, on_bass)` in that same PR, and add `FingeredOnBass [36,48] => C1+8` plus `Fingered [36,48] => C1+8` to `fingered_vs_on_bass`.
2. **Conflict with #42 (Upper detection).** #42 rewrites `live.rs::recompute` around `self.rec.recognize` and adds `shared.upper`. RM p.9 says: "When the chord detection area is set to 'Upper,' only 'Fingered*' is available". So after both land, `detect()` has to be bypassed or forced to Fingered* when `upper` is set, and `[f]`/the status line should show that. Please coordinate with #42 rather than leaving it to a merge-time guess.
3. **A left-hand open fifth plus right-hand melody changes the chord type on every note.** Left-hand C-G (36,43) with right-hand E (64) plus one melody note on top gives:

   | melody | chord |
   |---|---|
   | A | C6 |
   | B | Cmaj7 |
   | Bb | C7 |
   | D | Cadd9 |
   | F, Ab, F# | C |

   Rule 1 only covers a left-hand chord of 3+ pitch classes, and rule 2 keeps a same-root extension. So a scale run over this very common piano voicing changes the chord type on most notes, and each change bumps `generation` and re-transposes the style. The root never moves, so this matches the recorded Decision and I'm not blocking on it. A cheap improvement: when the top key is above the split and the reading without it is already a complete 3+ pitch-class chord on the same root, prefer that reading. That keeps C for the run but still lets a right-hand-only C-E-G-Bb block chord read as C7. Worth a playtest note either way.
4. **A right-hand-only first-inversion maj7 with the root on top reads as the iii minor.** E-G-B-C (64,67,71,72) gives **Em**, and A-C-E-F gives Am, because the top root "moves the root" and gets dropped as melody. With a left-hand bass these read correctly (F + A-C-E-F gives Fmaj7). This is the inherent ambiguity of rule 2, but it deserves one line in the Full Keyboard Decision.
5. **Multi Finger C-G is 1+5, which hides the Single Finger G7 shape "G + the white key C to its left".** It's documented, and F-G still gives G7. Consider noting in the Decision that the fourth below (D-G) and the second below (F-G) still give G7, so the owner knows which G7 shapes survive.
6. **Full / AI Full can never produce 1+5 or 1+8.** For example, `[36,48,60]` gives `-` in Full, and `[36,48]` in AI Full with prev C gives `-`. The RM says Full is "similar to Fingered", and Fingered has both. Needing 3 pitch classes is a defensible protection against melody, and it's recorded. Just flagging that it's a deliberate departure.
7. **Wrong test comment.** At `fingering.rs:~560`, `assert_eq!(name(&r, f, &[50, 53, 76], c), "Dm"); // (D-F-C would be Dm7, a chord)`: key 76 is **E**, not C. The test passes because D-E-F isn't a chord, not for the reason the comment gives. Either use 72 (C) and assert what actually happens, or fix the comment.

**Verdict: approve from the adversarial side (no blocking findings).** Items 1 and 2 are integration hazards that the integration-branch owner needs to handle when #44 and #42 land.

ryangavin:
Round 2 review follow-ups, pushed in 6c1efda. Gates: `cargo build --release` passes, and `cargo test --release` passes 30 with 1 ignored. The corpus was symlinked, and `corpus_full_performance_no_stuck_notes` ran on all 36 MOX_v2 styles. `cargo clippy --release` shows no warnings in `fingering.rs`.

**Addressed**
- **Test comment for AI Full `[50,53,76]`:** fixed. That chord is D-F-E, which is no chord, so E is treated as melody and D-F is inferred as Dm. I added the case the old comment meant: `[50,53,72]` (D-F-C) → Dm7.
- **Full Keyboard, right hand only, first-inversion maj7 with the root on top:** added tests (E-G-B-C → Em, A-C-E-F → Am). The case is now in the `full()` doc comment and in the Full Keyboard Decision as the cost of rule 2.
- **Multi Finger C-G = 1+5 hides the Single Finger G+C shape:** added a line to the Decision, with tests showing that F-G, D-G and E-G still read as G7.
- **Full / AI Full never produce 1+5 or 1+8:** this departure from "similar to Fingered" is now stated in the Decision and the module doc. (AI Full can still give 1+5 through its dyad inference.)
- **Merge hazard with #44:** strengthened the merge note in the PR body. Whichever of #41/#44 lands second **must** switch `fingered()`/`plain()` to `recognize_keys(s.mask, s.low % 12, s.keys, on_bass)` and add tests that On Bass and Fingered `[36,48]` give C1+8. This can't be done on this branch yet, because `recognize_keys` only exists on #44.
- **#42 (Upper detection):** added a merge note to the PR body. In the merged `recompute`, when `shared.upper` is set, bypass `fingering::detect` and use #42's On Bass + `fingered_star` path (RM p.9: only Fingered* in Upper). This can't be done on this branch yet, because `shared.upper` only exists on #42.

**Declined**
- **Full Keyboard: prefer the reading without a right-hand top key when that reading is a complete chord on the same root.** It breaks the standard shell voicing. Left-hand C-G with right-hand E-Bb has to be C7, but under the rule the right-hand top key Bb would be dropped, because C-G-E is a complete C triad. That would leave C. A test now pins C-G | E-Bb → C7. The drift the reviewer describes (C-G | E + A → C6, and so on) never moves the root, and it is recorded as the cost of rule 2. With the left hand we have no way to tell a melody note from a chord tone the player wants.

ryangavin:
Integration rebase onto integration/m1-chord-recognition (after #45 / #1 and #44 / #2):

- Squashed the four branch commits and rebased; conflicts were in `src/main.rs` (module list: kept both `mod fingering` and `#[cfg(test)] mod golden`; usage line: kept both `--fingering` and #1's `sim <style> <"C Am F G7" | script file>`) and `src/live.rs::recompute` (kept this branch's `fingering::detect` dispatch, as the merge note asks).
- Applied the merge note for #44: `fingered()` now calls `rec.recognize_keys(s.mask, s.low % 12, s.keys, true)` and `plain()` calls it with `on_bass = false` (dropping the local foreign-bass filter, which #44's plain Fingered path already does). So 1+8 is not lost in the default mode.
- New test `fingering::fingered_one_plus_eight_and_short_inputs`: On Bass and Fingered give C1+8 for [36, 48] and [36, 48, 60], C1+5 for C-G, and no chord for a lone key or a C-E dyad.
- Golden harness (#1): no `#[ignore = "M1 #3"]` tests existed; `golden_snapshots` passes unchanged (the sim path does not go through fingering), so no snapshot regeneration was needed.
- Gates: build ok; `cargo test --release` 127 passed, 1 ignored (pre-existing synth test), corpus symlinked and the corpus/golden tests ran; clippy shows no new warnings in touched code (the `live.rs` recompute collapsible-if and the `ui.rs` `map_or` in the moved SYNC STOP flag are pre-existing patterns).

Issue #36, opened 2026-09-23
Registration Memory 1–10 + banks, Freeze, Registration Sequence


ryangavin:
_Details in the spec._

---
Milestone: **M7 · Registration, Multi Pads, Playlist** · Spec: [docs/genos-features.md](../blob/main/docs/genos-features.md) · Backlog: [docs/backlog.md](../blob/main/docs/backlog.md)

ryangavin:
Implemented in PR 99. Decisions where the manuals are silent: see the PR body and docs/registration.md (Decisions section).

ryangavin:
Done in #99, merged to main via #113.

Issue #31, opened 2026-09-23
Parts model: Right 1–3 + Left, three split points, Left Hold, per-part octave


ryangavin:
- Replaces the current "voice slots" with the Genos part model. OTS and Registration target these parts.

---
Milestone: **M6 · Parts, voices & live play** · Spec: [docs/genos-features.md](../blob/main/docs/genos-features.md) · Backlog: [docs/backlog.md](../blob/main/docs/backlog.md)

ryangavin:
Decisions made in PR #61 (keyboard parts model). The owner playtests and may overrule any of them.


- **Decision: no "whole Style" volume fader on the Panel page, because** the Genos Panel tab has one but it scales all 8 style channels. Under the CC7 principle, that would need a hidden per-part multiplier or would overwrite the eight parts' CC7s. The Style page covers the style parts one by one, and master stays the only non-CC gain. Panel faders 5–8 are left dark for now.
- **Decision: the port carries exactly what sounds, so a part that is off sends nothing, Left included, because** that is the Genos model (a part that is off is silent) and it keeps port and synth identical.
  - Consequence: the left hand no longer goes out on ch 2 while Left is off (the default, as on the Genos).
  - To hear the left hand in Ableton, turn Left on (`l`, `8`, the page-3 pad, Panel fader button 4, or an OTS). Under Manual Bass it always sounds.
  - Owner: overrule this if the Ableton set needs ch 2 unconditionally.
- **Decision: each part's octave shift is applied to the notes it sends on the port, because** what the port carries should be what the part plays. With layered parts in different octaves, the Ableton tracks then match the built-in synth.
- **Decision: voices stay synth-only (no Program Change on the port), because** before this PR the port never carried the right/left voice either. An Ableton set picks its own instruments, and a Program Change could hit whatever is mapped on those tracks.
- **Decision: pedals, wheels, pressure and program changes from the keyboard go to all three Right parts, on or off, and not to Left, because** a sustain release must always reach a layered note that is still ringing. Left never received them before, and Left Hold (not in this PR) is the Genos's own sustain for Left.
- **Decision: the fader page starts on Panel, because** the complaint is that your own parts weren't on the faders, and Panel is the Genos Mixer's first tab. One press of the master-fader button gives the old Style layout.
[...]

PR #42, opened 2026-09-23
Chord Detection Area: Upper, with Fingered* and Manual Bass (#5)

ryangavin:
## What

Chord Detection Area **Upper** and **Manual Bass**, following OM p.51 and RM p.9.

- **Upper mode:** the chord comes from the keys above the split, which here is Split Point (Left). Those keys still sound on the Right voices (ch 1). The left hand plays the Left part (ch 2) and never changes the chord. Sync Stop follows the right hand: releasing every right-hand key counts as releasing the chord.
- **Fingered\*** (`live::fingered_star`): Fingered without 1+5, 1+8 or Chord Cancel, and the bass is always the root. Right-hand melody fragments (single notes, octaves, fifths, thirds, chromatic runs) leave the current chord alone. This is a gate at the input layer that runs after the recognizer; the recognizer internals are not touched (#2 and #3 own those).
- **Manual Bass** (defaults to On when you select Upper; only takes effect in Upper mode):
  - Engine: the Style's Bass part (ch 11) is muted, the note sounding at the switch is released, and Stop Accompaniment's bass note is muted too. Implemented as a gate in `Engine::note_on`, so no code path can get around it.
  - Built-in synth: the Left part plays the current Style's Bass voice (GM fallback, Finger Bass when the Style has none). It sounds even when the LEFT voice is off. Turning Manual Bass off restores the user's Left voice and its on/off state.
- **Settings:**
  - CLI: `--upper` and `--no-manual-bass`.
  - TUI: `d` toggles the area and `D` toggles Manual Bass (Upper only).
  - Display: a new status line shows the area, Fingered\*, Manual Bass, which keys detect the chord, and what the left hand does. The LEFT label reads `bass: <voice>` while Manual Bass is on.
- **Input routing:** the handler now records where each held key went (LH, RH, chord section). Note-offs and Sync Stop follow that record, which fixes note-offs when the split or the area changes while a key is held.

All real-time paths stay allocation- and lock-free. The route table is a `[u8; 128]` inside `Input`, and the state lives in atomics.

## How tested

- `live::detection_area` tests drive the real `Input` handler, capturing its output through the synth ring:
  - Lower mode detects the chord from the left hand.
  - Upper mode detects the chord from the right hand. A left-hand triad does not detect. Melody notes, fifths and octaves are ignored. An inversion is reported in root position with no slash. Releasing the left hand does not send ChordReleased.
  - Note-offs follow the note-ons across a mode switch.
  - Fingered\* rejects 1+8, 1+5 (both inversions), two-note shapes and Cancel, and drops the bass.
- `sim::manual_bass` (corpus: SlowWalker):
  - The Bass part plays before the switch, stays silent while Manual Bass is on, and returns after.
  - Every bass note that was sounding at the switch gets released.
  - The other parts are unchanged.
  - Stop Accompaniment's bass is muted while its pad chord still sounds.
- `synth::layer_tests::manual_bass_sounds_the_left_hand` (SoundFont):
  - The left hand sounds with LEFT off.
  - Note-offs still land when Manual Bass is toggled off mid-note.
  - Checks the bass program mapping.
- Gates:
  - `cargo build --release` is clean.
  - `cargo test --release`: 26 passed, 1 ignored (pre-existing probe). The corpus and SoundFont were symlinked, and no test reported skipping.
  - `cargo clippy --release`: 31 warnings before and after, with the same set per file. No new warnings.

## Decisions

- Decision: **Fingered\* needs at least 3 distinct pitch classes, and a Cancel/1+8/1+5 result is ignored so the previous chord holds.** Because: the Data List p.44 table shows 1+8 and 1+5 are the only Fingered shapes under 3 notes, and Fingered\* removes both. The effect is that right-hand melody notes and dyads never re-chord the Style, which is what makes Upper mode playable.
- Decision: **Under Fingered\*, the bass is always the chord root (no slash chords from the right hand).** Because: the manuals say Fingered\* is "basically the same as Fingered", and Fingered uses the root as bass. The left hand carries the bass line anyway.
- Decision: **Upper mode uses the single existing split as Split Point (Left). Keys at or below it are the left hand; keys above it are the chord section.** Because: OM p.51 puts the chord section to the right of Split Point (Left). The app has one shared Style+Left split for now; #31 adds the three separate split points.
- Decision: **Manual Bass makes the Left part sound, with the Style's Bass voice, even if the LEFT voice is off. Turning it off restores the user's Left voice and on/off state.** Because: OM p.51 says the Bass part's voice "is assigned to the left part", and the point of the mode is that the left hand plays the bass.
- Decision: **The Bass voice is played at the keys you play (no octave shift beyond the Left part's own octave setting).** Because: the manuals specify no transposition. Style bass parts already sit in the range of the left-hand keys.
- Decision: **Manual Bass also mutes Stop Accompaniment's bass note.** Because: it is the Style Bass part's voice, and Manual Bass mutes that part.
- Decision: **With MIDI out only (no built-in synth), Manual Bass mutes ch 11 and sends the left hand on ch 2, but sends no program change for the bass voice.** Because: we can't restore an external device's Left voice afterwards, so the voice assignment stays in the built-in synth.
- Decision: **Selecting Upper turns Manual Bass on each time (its documented default). Switching back to Lower keeps the setting but it has no effect.** Because: OM p.51 says Manual Bass is set to On by default when Upper is selected and is only available in Upper mode.
- Decision: **The area is not stored in OTS, since OTS does not store fingering type or split.** Registration storage (DL: Style group, Freeze "Fingering Type") is left to #36.
- Decision: **Keys held while the area is switched count toward the chord of the new area: a held key belongs to the chord section whenever it is on the chord side of the split at the time.** Because: Genos detects the chord from the keys currently held in the chord detection area, not from the keys pressed while the old area was active. The new chord takes effect at the next chord-section key press.

Not in this PR: Ensemble Voice forcing the area back to Lower (no Ensemble Voices yet), Full Keyboard/AI modes under Upper (#3 owns fingering types), and a Launchkey control for the area.

Closes #5


ryangavin:

**Verdict: no blocking findings.** Nonblocking suggestions are below.

### What I checked
- Checked out `b08bcfa` detached, with the corpus and SoundFont symlinked. `cargo build --release` is clean. `cargo test --release`: 26 passed, 1 ignored. `sim::manual_bass::*` and `synth::layer_tests::manual_bass_sounds_the_left_hand` really ran: `SlowWalker.T552.sty` and `GeneralUser-GS.sf2` are both present, and neither test printed a skip.
- `cargo clippy --release`: I diffed the warnings in live/engine/synth/ui/main against `integration/m1-chord-recognition`. They are the same set, with only line numbers shifted. Nothing new.
- **Fidelity, checked against the manual text:**
  - OM p.51: the chord section is to the right of Split Point (Left), Fingered* is set automatically, and Manual Bass mutes the Style's bass part and assigns its voice to the Left part. It defaults to On when Upper is selected and is only available in Upper.
  - RM p.9: Fingered* is Fingered without 1+5, 1+8 and Cancel, and "In the Fingered type, the root of the chord is always used as the bass note".
  - Data List p.44: every Fingered shape apart from 1+8/1+5 has three or more tones, even with the optional (parenthesised) tones left out. Fewer than three notes is only allowed in *AI* Fingered (RM p.9).

  So these decisions match the documentation: at least 3 pitch classes, bass = root (`fingered_star` sets `bass: None`), the split key goes to the left hand, and the chord keys still sound the Right voices.
- **Real-time safety:** `Input` routing uses a fixed `[u8; 128]` and atomics. `Engine::note_on` adds a single branch. The synth reads atomics. There is no allocation, lock or panic path on the MIDI or engine threads.
- **Stuck notes:**
  - `set_manual_bass(true)` releases the sounding notes on ch 11 through `off_where`.
  - Muted notes never go into `sounding`, so no stray note-offs and no re-pitch on a chord change.
  - The synth Left note-off follows `pl.lh`, so it still lands after Manual Bass is toggled mid-note.
  - `Engine::load` keeps the `manual_bass` flag across a style swap, and the UI refreshes `bass_program` on the only style-swap path (ui.rs:419).

### Nonblocking
1. **The chord mask gets polluted across an area switch.** src/live.rs, `recompute` (~l.188) and the note-on routing (~l.215).
   - A key keeps the `R_CHORD` bit it got at note-on, so after `d` it keeps feeding the new mode's chord mask.
   - Repro (throwaway test): in Upper, hold RH F#4 (78) as a melody note. Press `d` to go to Lower. Play LH A-C-E (45/48/52).
   - Expected: **Am**. Actual: **Am6** (`Chord { root: 9, ty: 9 }`). It stays wrong until the F# is released.
   - The reverse happens too. LH keys held while switching Upper→Lower are not chord keys, so the chord doesn't follow them until they are pressed again.
   - The PR's own test works around this ("C + C = C").
   - Suggested fix: keep `R_LH`/`R_RH` for the note-off routing (that part is correct and worth keeping), but build the mask in `recompute` only from keys whose side matches the *current* area. Or re-classify the held keys when the area changes. Neither needs an allocation.
2. **No test for a split change while a key is held.** The PR says routing now fixes note-offs when the *split* moves under a held key, including the repeated note-on path that sends an LH off and then an RH on. Only the area switch is tested. A 10-line test on `Rig` that moves `shared.split` between the on and the off would pin it.
3. **The Bass voice comes only from the style's initial setup.** `info.voices[10]` is built from the init/SInt messages. If a style changes the Bass program per section, Manual Bass keeps playing the first voice. That's fine for M1, but worth a `// TODO` or a backlog note.
4. **`--no-manual-bass` is lost after pressing `d` twice.** Because of the "Upper turns Manual Bass on each time" decision, `--upper --no-manual-bass` followed by `d`,`d` turns it back on. OM says it goes On "automatically by the default settings", which reads as a default rather than a forced reset. That supports the decision, so I'm only flagging it for the owner's playtest.
5. **The mixer doesn't show that Bass is muted.** Manual Bass mutes ch 11 through a separate flag, not the `parts` mask, so the part row still looks active. Consider dimming the Bass row, or adding "(manual bass)" to its label, while it's in effect.
6. **Magic numbers in `fingered_star`.** It uses `30 | 31` for 1+8/1+5. Tickets #2/#3 are changing the recognizer, so named constants next to `CANCEL` in `theory.rs` would be safer. I understand this PR avoids touching theory.rs to reduce conflicts. Today the recognizer never returns 30, and returns 31 only for 2-note masks, so the pitch-class count check already covers both.

### Scope
The change stays within input routing, the engine gate, the synth Left voice and the UI. It does not touch the recognizer internals (#2/#3) or the split points (#31). Conflict risk is low. The main overlap is the `Input::key_msg` routing, which any other ticket that touches input will have to rebase over.

ryangavin:
Round 1 review addressed in bb43ef4.

**Addressed**
- Stale chord membership across an area switch: the per-key route table now stores only the side of the split (LH/RH). Whether a key counts as a chord key depends on its side and the *current* area, so after Upper→Lower a held RH F# no longer turns Am into Am6. LH keys held through Upper→Lower count as chord keys as soon as the next chord key is pressed. Sync Stop now fires when the last held key on the current chord side is released (a scan of the 128-entry table, no allocation). Test: `held_keys_join_the_chord_of_the_current_area`, which covers both directions and includes your repro.
- Split change while a key is held: new test `repeated_note_on_after_a_split_change_moves_the_note`. It checks the LH off followed by the RH on, that the note-off follows the note, and that Sync Stop only fires once the remaining chord keys are up.
- Per-section Bass voice: added a TODO on `style_bass_program` and a follow-up line under #5 in docs/backlog.md.
- The parts row now dims the Bass part and shows "(muted: Manual Bass)" while Manual Bass is on.
- Replaced the magic numbers 30/31 with `theory::ONE_PLUS_EIGHT` / `ONE_PLUS_FIVE`.

**Declined**
- Manual Bass coming back on after `--upper --no-manual-bass` then d,d: kept as documented (OM p.51: selecting Upper turns Manual Bass on by default). It's already in the Decisions section for the owner to playtest.

Gates: cargo build --release, cargo test --release (28 passed, 1 pre-existing ignore, corpus tests ran), clippy with no new warnings in touched code.

ryangavin:
## Integration: rebased onto `integration/m1-chord-recognition` (after #1–#4)

Squashed the two ticket commits into one and rebased. Conflicts were in `live.rs`, `ui.rs`, `main.rs` and `theory.rs`, all against #3 (fingering types). #3 and #5 both rewrote how `Input` tracks held keys. The resolution keeps both behaviours:

- **Held-key tracking:** #5's `route[u8; 128]` (which side each key went to) replaces #3's `held` / `held_all` / `zone_held`. `recompute` builds the `[bool; 128]` that `fingering::detect` expects on the stack from `route`: the chord side in the current area, or both sides for the Full Keyboard types in Lower. No allocation. Note-offs and Sync Stop keep #5's route-based logic.
- **Upper vs. fingering type:** In Upper, the selected fingering type is overridden by Fingered\* (OM p.51: "only Fingered\* is available" / "Fingering type is set to Fingered\* automatically"). The selected type is kept and applies again when you switch back to Lower. Full Keyboard in Lower still reads both hands (#3's behaviour).
- **Sync Stop:** New `Shared::sync_stop_allowed()` returns true in Upper, and otherwise follows `Fingering::allows_sync_stop()`. The engine loop and the status line both use it, so a Full Keyboard selection does not disable Sync Stop while Upper/Fingered\* is in effect.
- **`theory.rs`:** kept #5's `ONE_PLUS_EIGHT` / `ONE_PLUS_FIVE` constants next to #2's new ones.
- **UI and CLI:** both `f` (fingering) and `d`/`D` (area / Manual Bass) are kept. The fingering readout adds "(Upper: Fingered\*)" while Upper is on. The usage line lists `--fingering` and `--upper [--no-manual-bass]`, and `ui::Options` carries both.
- **`sim::Step::ManualBass`:** #1's script grammar now builds `Step` outside tests, so this variant showed up as dead code (a new build warning). I gave it `cfg_attr(not(test), allow(dead_code))`, since it isn't in the script grammar yet.
- **#3's tests:** `full_keyboard_reads_both_hands_and_tracks_releases` now asserts on `route` instead of the removed arrays. The behaviour it checks is unchanged.
- **New test:** `live::detection_area::upper_overrides_the_fingering_type`. With Full Keyboard or Single Finger selected, Upper still reads only right-hand Fingered\* chords and allows Sync Stop. Back in Lower, Full Keyboard reads C/E again.

### Decisions
- Decision: **In Upper, Fingered\* is #3's plain `Fingered` (`fingering::detect(Fingered)`) followed by `fingered_star` (at least 3 pitch classes, no 1+5, 1+8 or Cancel, bass = root).** The PR originally used `Recognizer::recognize` (On Bass reading) and then dropped the bass. Because: the manual defines Fingered\* as "basically same as Fingered" with those three removals, and #3 now owns the definition of Fingered. One consequence: a right-hand shape whose lowest key is not a chord tone makes no chord, so the chord holds, as in Fingered, instead of reading as that chord with the bass dropped.
- Decision: **Switching to Upper doesn't overwrite the stored fingering type.** Because: on Genos, returning to Lower gives you back your fingering choice, and it keeps `f` meaningful.

### Golden / ignored tests
- #1's golden snapshots and recognizer table pass unchanged. No `UPDATE_GOLDEN` was needed.
- There are no `#[ignore = "M1 #5"]` tests.

### Gates
- `cargo build --release`: clean, 0 warnings.
- `cargo test --release`: 145 passed, 1 ignored (the existing loudness probe). The corpus and SoundFont were symlinked, and the golden and corpus tests ran.
- `cargo clippy --release`: no new warnings in the code touched. The one new warning this merge introduced (a range loop in `recompute`) was fixed.

Issue #14, opened 2026-09-23
SFF1 NTT codes above 5, and Cntt precedence


ryangavin:
- Decode every SFF1 NTT value instead of falling back to Melody.
- Confirm Cntt vs Ctb2 precedence using the oracle.

---
Milestone: **M2 · Chord-to-note conversion 1:1** · Spec: [docs/genos-features.md](../blob/main/docs/genos-features.md) · Backlog: [docs/backlog.md](../blob/main/docs/backlog.md)

ryangavin:
Decisions from #56. The corpus evidence is in the PR body.

- Decision: when a Ctab has a Cntt, the Cntt overrides the NTT table, but Bass On is the Ctab "Bass" code OR the Cntt bit 7, because every corpus Cntt for a Ctab Bass channel is `01H` (Melody) with bit 7 clear (7/7 Cntt styles). Replacing Bass On would make those Bass parts ignore slash chords, while 193/194 SFF2 styles give the Bass part Bass On.
- Decision: `Cntt` never overrides a `Ctb2`, because Ctb2 already stores NTT and Bass On per zone. Wierzba/Bedesem say Cntt is not used with Ctb2, and no corpus file mixes them (194 Ctb2 files, 7 Ctab+Cntt, 5 Ctab only).
- Decision: SFF1 Ctab NTT codes `06H`–`0AH` decode as the Cntt/Ctb2 tables with the same numbers (Harmonic Minor 5th … Dorian 5th), because Ctab documents only `00H`–`05H` and those numbers mean nothing else anywhere. Bit 7 means Bass On, and codes above `0AH` play as Melody. No corpus file uses any of these (all 78 Ctab records are 0/1/2/3/5).

ryangavin:
Decision update (PR #56, review round 1): an SFF1 Ctab NTT byte has no Bass On bit. Decision: `80H`–`FFH` (and anything else above `0AH`) play as Melody without Bass On, and Ctab Bass On comes only from the `03H` Bass code, because nothing documents a Bass On bit in a Ctab byte and treating `80H` as Bypass + Bass On would shift a Bypass channel's roots on slash chords. No corpus file uses such a byte.

On the Cntt Bass On decision (unchanged): this departs from the literal Wierzba/Bedesem reading (Cntt bit 7 = Bass on/off). The corpus evidence is one library (7 MOX_v2 .T552 files), so the oracle may overrule it.

Issue #11, opened 2026-09-23
NTT 5th-variation tables (aug/dim handling)


ryangavin:
- Melodic minor 5th, Harmonic minor 5th, Natural minor 5th and Dorian 5th currently behave like their base tables. Implement the flattened/sharpened-5th handling.

---
Milestone: **M2 · Chord-to-note conversion 1:1** · Spec: [docs/genos-features.md](../blob/main/docs/genos-features.md) · Backlog: [docs/backlog.md](../blob/main/docs/backlog.md)

ryangavin:
Decision (PR #55): the aug and dim chord types are the CASM types whose 5th is altered and that have no perfect 5th alongside: aug, M7aug, 7aug, dim, dim7, m7b5 and 7b5. (b5) plays as 7b5, so it counts. M7b5 plays as M7(#11), which has a perfect 5th, so it does not. Neither do 7(#11) and 7(b13). Why: the Reference Manual only says "augmented and diminished chord types". b5 chords are the diminished family, and keeping a perfect 5th in a chord that already contains one would be pointless.

ryangavin:
Decision (PR #55): only the 5th changes between a base minor table and its "5th" variant. The base tables now keep the perfect 5th over aug and dim chords. Over dominant, dim7 and similar chords they still adjust the other degrees the way they did before (for example the b7 over C7). Why: this ticket is about the aug/dim 5th. The manual's "Other notes are not changed" also describes major/minor-only behaviour for the 6th and 7th, and that is a larger change with golden impact. It should be a separate ticket, checked against the corpus oracle (#8). Corpus: the base tables are used mostly in Intros and Endings, and the "5th" tables mostly in Mains and Fills.

ryangavin:
Revised decisions (PR #55, after review):

- Decision (replaces the earlier deferral): the minor tables are a major/minor switch that moves only the degrees each table names:
  - Melodic Minor: the 3rd.
  - Harmonic Minor: the 3rd and 6th.
  - Natural Minor: the 3rd, 6th and 7th.
  - Dorian: the 3rd and 7th.

[...]

PR #43, opened 2026-09-23
Chord Cancel, 1+8 and 1+5 playback rules (#4)

ryangavin:
## What

This PR sets what the band plays under Chord Cancel, 1+8 and 1+5.

- **Chord Cancel (type 34).** The manual calls Cancel "a state in which no chord is input" (OM p.46). So the engine now resolves Cancel to the same state as before the first chord, through `engine::effective_chord`:
  - Rhythm channels and channels with the CASM autostart bit keep playing, as recorded at their source chord.
  - Every other part is released at once and rests until the next chord arrives.
  - Cancel still does not trigger Sync Start.
  - A chord that ends the no-chord state up to 40 ms after the beat (`LATE_CHORD_NS`) still gets the downbeat: `Engine::catch_up` starts the notes the resting parts skipped in that window and the pattern still holds. It also covers Start with no chord and parts un-muted by CASM chord-mute routing. It uses a fixed-size buffer, so nothing is allocated on the RT thread.
  - `revoice` uses the same resolution.
  - `transpose()` now returns `None` for Cancel. Before, it could index past the 34-entry tables and panic on the RT thread.
- **1+8 (type 30).** The CASM chord-mute bit 30 still decides which channels play. Parts that follow the chord play only the root, in octaves. This covers Melody, Chord, Bass, the minor tables and Guitar. The Guitar table's "fifth" string (source C#) plays the octave.
- **1+5 (type 31).** The CASM chord-mute bit 31 decides which channels play. Parts that follow the chord play only notes that fit both major and minor:
  - Root, 5th, 2nd and 4th are kept.
  - Melody moves 3rd -> 5th, 6th -> 5th and 7th -> octave.
  - Chord-table parts keep root and 5th only: after the Chord mapping, a 2nd falls to the root and a 4th rises to the 5th.
- **Chromatic passing notes** used to keep their offset from the scale degree, so Eb over C1+8 sounded as C#. Over 1+8 and 1+5 they now snap to the scale degree below.
- **Patterns recorded over 1+5 / 1+8** (`src_type` 30/31) are read with a monotonic major scale. The old scale for 31 was `[0,2,7,5,...]`, which sent a source G to A when playing a major chord. The Chord table keeps the root and moves the 5th to the target's 5th.
- The rules are also recorded in `docs/genos-features.md` §C.3.

## Corpus evidence

I dumped all 461 CASM channel rules in `corpus/MOX_v2` (36 styles):
- 426 rules allow every chord type.
- The rest are all Bass channels (NightCruiser, SmoothItOver). They split the bass across 2–3 source channels whose chord-mute masks **partition** the 34 types:
  - Maj only / aug+m+m7 / everything else.
  - Minor family (bits 8–18) / major+dominant family (bits 0–7, 19–33).
- In every split, 1+8 and 1+5 go to exactly one source, and that source is always in the major or "everything else" group, never the minor-only one. Authors use the mute bits to *route* these chords, not to silence parts. So applying the chord-mute bits for types 30/31 as they are is correct, and the bass is never left silent.
- The CASM autostart bit appears only on rhythm channels (dest 9/10) in the corpus. "No-chord state" and "rhythm only" therefore give the same result there.

## How tested

- Unit tests in `theory.rs`:
  - `one_plus_eight_plays_only_the_root`: Melody, Chord, Bass, HarmonicMinor, Dorian5 and Guitar, from CM7 and Cm7 sources, all 24 chromatic keys.
  - `one_plus_five_is_quality_neutral`: also checks Chord NTT (Root Fixed and Root Trans) from every source type 0–29 and every key, and asserts that only root and 5th come out.
  - `source_recorded_over_one_plus_five`.
  - `cancel_has_no_pitches`.
- Sim tests in `sim.rs`:
  - `chord_cancel_leaves_only_rhythm` runs over **all corpus styles** (C -> Cancel -> F). At the Cancel instant every resting part is balanced off. Only rhythm and autostart channels play while cancelled, rhythm keeps going, and the band comes back on F.
  - `chord_cancel_does_not_sync_start`.
  - `late_chord_after_no_chord_keeps_the_downbeat` runs over **all corpus styles**. After Cancel, and after Start with no chord, an F that arrives 20 ms late leaves exactly the notes sounding that an on-time F would. It fails without `catch_up`.
  - `corpus_one_plus_eight_and_one_plus_five` runs over **all corpus styles** (G1+8 -> D1+5 -> C). Channels with no Bypass zone play only G under G1+8, and only D/E/G/A under D1+5. The bass plays under both chords whenever it plays under C. Any part whose Main A rules allow the chord type and root, and that plays under G / D major over the same bars, must also play under G1+8 / D1+5. This stops the pitch checks from passing on a silent channel.
- Mutation check: with the old `theory.rs` restored, the 1+5 corpus test fails. For example, BluesOrganTrio bass plays B under D1+5.
- Gates:
  - `cargo test --release`: 27 passed, 1 ignored. The corpus is symlinked and the tests ran; no "skipping" lines.
  - `cargo clippy --release` shows no new warnings in touched code. The remaining warnings in these files were there before this PR.

## Decisions

- Decision: Chord Cancel = the no-chord state. Rhythm channels and CASM-autostart channels play as recorded; all other parts are released immediately and rest until the next chord. Because: the OM (p.46) defines Cancel as "a state in which no chord is input", and our engine already defines that state (before the first chord) this way. In the corpus, autostart appears only on rhythm, so in practice this matches PLAN.md's "mutes everything except rhythm".
- Decision: Chord Cancel does not trigger Sync Start (unchanged). Because: the OM (p.66) says the Style starts when "you play a chord", and Cancel is the absence of a chord.
- Decision: 1+8 and 1+5 use the CASM chord-mute bits 30/31 as they are, with no fallback routing. Because: every corpus style that uses chord-type routing partitions all 34 types across its source channels, including 30/31. The bits are the author's intended routing.
- Decision: under 1+8, chord-following parts (every NTT except Bypass) play only the root, in octaves. Chromatic notes and the Guitar fifth string collapse to the root. Because: Yamaha describes 1+8 as accompaniment based only on the root, and this matches how the existing 1+8 target scale was written.
- Decision: under 1+5, Melody-type tables keep root, 5th and the 2nd/4th. The 3rd and 6th go to the 5th, the 7th goes to the octave, and chromatic notes snap to the scale degree below. Chord-table parts keep only root and 5th. Because: 1+5 is meant to work under both major and minor, and the 2nd and 4th are the only other scale degrees that are the same in both. Collapsing everything to root/5th would flatten walking bass lines more than needed.
- Decision: NTT Bypass channels still play as written under 1+8/1+5, 3rds included. Because: Bypass means the author asked for no chord conversion; only the chord-mute bits can silence them.
- Decision: an Ending pressed under Chord Cancel plays rhythm only. Because: it follows directly from Cancel being the no-chord state, and the manuals give Endings no exception. This may surprise in playtest.
- Decision: a chord that ends the no-chord state within the late-chord allowance (40 ms) catches up the downbeat notes of the parts that were resting. Because: this is the same allowance every other chord change already gets, and without it a slightly late chord after Cancel loses the whole band's downbeat.
- Out of scope, filed as #46: a chord that collapses voices onto one pitch (common under 1+8) can make `revoice` send two note-ons for the same key. On/off counts stay balanced, but the first note-off releases both.

Closes #4


ryangavin:

I checked out `m1/4-cancel-1-5-1-8` (b276225) with the corpus symlinked and confirmed that all 36 MOX_v2 styles load and run. `cargo test --release` gives 26 passed, 1 ignored, and prints no "skipping" lines. `cargo clippy --release` shows no new warnings: the ones in touched files (theory.rs:330, 372, 373 and 488, engine.rs:913) were already on the base.

I also wrote throwaway probes, which I did not commit:
- a pitch-class sweep over every NTT (Melody, Chord, Bass, all minor/Dorian variants and Guitar), with RootTrans and RootFixed, every source type 0–33 and keys 48–71, played over G1+8 and G1+5;
- two corpus sim scripts, described under B1 and N2 below.

### What holds up

- **Cancel = no-chord state.** The manuals only define it as "a state in which no chord is input" (OM p.46; DL p.45). Reusing the engine's existing no-chord semantics is the most defensible reading.
- **No Sync Start on Cancel.** Reasoned correctly.
- **1+8.** Every NTT except Chord, over every source type, gives only the root. Guitar over 1+8 and 1+5 also gives only root, or root and 5th.
- **1+5 Melody-family tables.** Output is only root, 2nd, 4th and 5th for all 34 source types.
- **Balance.** A full performance script mixing Cancel, 1+8 and 1+5 with Main/Fill/Break/Intro/Ending leaves on/off counts balanced on every corpus style.
- **RT safety.** `effective_chord` is branch-only, with no allocation, locks or panics. The new `transpose` guard removes a real out-of-bounds panic path for Cancel.
- **Corpus analysis of chord-mute routing.** Convincing. Using bits 30/31 as they are is right.

### Blocking

**B1. Re-entering after Cancel drops the downbeat when the chord arrives slightly late** (engine.rs `set_chord` / `revoice`, together with the new `effective_chord`).

The engine has a deliberate late-chord allowance, `LATE_CHORD_NS` = 40 ms: notes that started just before a chord change are re-pitched instead of lost. Under Cancel, this PR silences every non-rhythm part and does not emit their notes at all. So when the player comes back with a chord a few ms after the barline, those downbeat notes never existed and there is nothing to re-pitch. They are gone until each part's next note-on. Coming back in after a Cancel break is exactly when players land a hair late.

Repro, per corpus style: script C@0 → Cancel@2⅓ bars → F at bar 4 + Δ. I counted non-drum note-ons in the first ⅛ bar of bar 4:

| style | F on time (Δ=0) | Δ=20 ms after Cancel | Δ=20 ms, Am instead of Cancel |
|---|---|---|---|
| NightCruiser | 1 | **0** | 2 |
| OrganCruise | 1 | **0** | 2 |
| CoolRevibed | 12 | **1** | 22 |
| DownTime | 12 | 6 | 18 |
| JumpinJiveOrgan | 12 | 6 | 16 |

Across the corpus, 7 of 35 styles lose every non-drum downbeat note at Δ=20 ms after Cancel. Without Cancel, the same timing loses none on any style. (The same gap exists for a first chord arriving while the band is already running, but this PR is what defines Cancel as "release and rest", so the way back out of that state belongs in scope.)

Expected fix: when a chord ends the no-chord state, emit the chord-following note-ons that were skipped within `LATE_CHORD_NS` before `now`. Or, equivalently, remember skipped note-ons in a fixed-size ring so nothing is allocated on the RT thread. Add a sim test with the chord arriving about 20 ms late after Cancel.

**B2. The stated 1+5 rule for Chord-table parts doesn't match the code, and the unit test hides it** (theory.rs `chord_map` → `scale_map` fallback, and `one_plus_five_is_quality_neutral`).

The PR body and docs/genos-features.md §C.3 both say "Chord-table parts keep only root and 5th". But any Chord-NTT note that is not a chord tone falls through to `scale_map` and comes out as the 2nd or 4th. Concrete cases:
- `Ntt::Chord`, `Ntr::RootFixed`, src_type 2 (CM7), key 50 (D) over G1+5 → pitch class 9 (A).
- Key 53 (F) → pitch class 0 (C).
- The same thing happens for src types 0, 1, 10–20 and more: 436 hits in the sweep.

`no_third_outputs` checks every NTT against `{7, 9, 0, 2}`, so it cannot catch this. The only Chord-NTT assertion uses the chord tones E, G and B.

Either choice is defensible (a sus2/sus4 colour over a power chord is quality-neutral), but code, doc and test must say the same thing. Pick one:
- **Code change:** snap Chord-NTT non-chord tones to root/5th under 30/31, and assert `{root, 5th}` for Chord NTT across all 12 source pitch classes.
- **Doc change:** reword the Decision and §C.3 to "Chord-table parts: chord tones go to root/5th; tensions follow the 1+5 scale (2nd/4th)", and add a test that pins it.

### Non-blocking

- **N1. The duplicate same-key note-on issue gets worse.** The PR lists it as out of scope. With the probe script, CoolRevibed goes from 4 duplicates on a plain-chord baseline to 10 under 1+8/1+5, and SlowWalker from 2 to 7. Counts stay balanced, but on external MIDI gear the first note-off cuts both voices. Please file it as a follow-up ticket now, since 1+8 makes it common rather than rare.
- **N2. Ending while in Cancel.** Pressing Ending after Cancel now plays a drums-only ending, because every non-rhythm part rests. This is consistent with the no-chord definition, but it may surprise in playtest (Yamaha arrangers usually keep the last chord for the ending). Worth a line under "Decisions".
- **N3. Sim test gaps.**
  - `corpus_one_plus_eight_and_one_plus_five` only proves the bass is not silenced. It would still pass if every other chord-following channel went silent under 1+8/1+5, because the pitch checks are vacuous on an empty set. Consider asserting that each channel which plays under C and whose chord-mute bit 30/31 is set also plays under G1+8/D1+5.
  - The Bypass exemption only looks at Main A's CASM. That is correct for the script as written, but it will break silently if someone adds a section change.
- **N4. Chord NTT with RootTrans can move a voice more than an octave.** Under 1+8/1+5 there are jumps of up to 13 semitones: src_type 3 or 23, key 54 → 67, and key 66 → 79. This probably comes from the existing RootTrans shift plus the "keep movement small" delta and is likely already present on the base. It is not introduced here; flagging it for the NTT-fidelity ticket.
- **N5. Merge-conflict risk.** `scale()`, `scale_map()` (the signature changed) and `chord_map()` are core transposer code that sibling NTT tickets are likely to touch. A heads-up on the integration branch would help whoever merges second.

**Verdict:** changes requested (B1, B2).

ryangavin:
Round 1 review is addressed in 284caae.

**Blocking**
- **B1 (late chord after Cancel loses the downbeat): fixed.** The new `Engine::catch_up` runs after `revoice` on every chord change while the band is running. It scans the current section's events from `now - LATE_CHORD_NS` up to the play cursor. For parts the previous chord kept silent (no chord, Cancel, or a CASM chord-mute route) that the new chord plays, it starts every note-on the pattern still holds. It uses a fixed 32-entry buffer, with no allocation or locks. The new sim test `late_chord_after_no_chord_keeps_the_downbeat` runs over all corpus styles. It covers C -> Cancel -> F at bar 4 + 20 ms, and also Start with no chord -> F at bar 4 + 20 ms. The notes sounding at the moment F lands must equal those from an on-time F. With `catch_up` disabled, the test fails.
- **B2 (Chord NTT over 1+5 gave the 2nd and 4th): code changed to match the doc.** `chord_map` now snaps the 2nd to the root and the 4th to the 5th when the target is 1+5. `one_plus_five_is_quality_neutral` now checks Chord NTT with Root Fixed and Root Trans, for every source type 0–29 and keys 48–71, and asserts that only {root, 5th} come out. Your example, src 2 key 50/53 over G1+5, now gives D/G.

**Nonblocking**
- **N1: filed as #46** (duplicate same-key note-on in revoice), in the M2 milestone.
- **N2: done.** "Ending under Cancel plays rhythm only" is now a Decision in the PR body and noted in `docs/genos-features.md` §C.3.
- **N3: done.** The corpus 1+8/1+5 test now runs a G-major/D-major reference over the same bars. Any non-drum part whose Main A rules all allow type 30/31 and the root, and that plays in the reference, must also play under G1+8/D1+5. I checked this is not a trivial pass: blocking type 30 in `plays` makes the test fail. Only Main A's CASM is read for the Bypass exemption because the script only ever plays Main A, so it matches what is actually playing.
- **N4: declined here, passed on.** Chord NTT with Root Trans can move a voice 13 semitones. That comes from the "keep the movement small" step on the base transposer, not from this ticket. It belongs to the NTT-fidelity work (#11 / #14).
- **N5: noted.** This PR changes `scale`, `scale_map` (new `tgt_ty` argument), `chord_map` (now a wrapper around `chord_map_raw`) and adds `src_scale`. Whoever merges second among the NTT tickets should expect conflicts in `theory.rs` and should keep the `is_no_third` snap and the 1+5 Chord snap.

**Gates:** `cargo build --release` passes. `cargo test --release`: 27 passed, 1 ignored, with the corpus symlinked and the tests actually run. `cargo clippy --release` shows no new warnings in touched code.

ryangavin:

I checked out 284caae (detached from `origin/m1/4-cancel-1-5-1-8`) with the corpus symlinked. All 36 MOX_v2 styles load.

- `cargo test --release`: 27 passed, 1 ignored, and no "skipping" lines.
- `cargo clippy --release`: no new warnings. The warnings left in touched files were already on the base: engine.rs:576/975, and theory.rs:193/206/502. theory.rs:502 is the on-bass `if let` and is unchanged.

### Round-1 fixes, verified

- **B1 (late chord after no-chord loses the downbeat): fixed.**
  - I replaced the `catch_up` call in `set_chord` with a no-op. `late_chord_after_no_chord_keeps_the_downbeat` then fails (sim.rs:229), so the test really exercises the fix.
  - I also probed the case the new code quietly widens: CASM chord-mute routing. Cm then C, 20 ms late at bar 4, on every corpus style. The notes held once the chord lands are identical to an on-time C on every style, so there are no spurious duplicates from the routing path.
  - RT safety: fixed `[_; 32]` buffer, no allocation, no locks. Slice indexing is bounded by `end = ev_idx.min(len)` and there is no `unwrap`.
- **B2 (Chord NTT over 1+5 gave the 2nd/4th): fixed.**
  - I reran my round-1 sweep with a wider range than the unit test covers. It crossed every NTT (all 14 non-Bypass tables, including GuitarStroke/Arpeggio, MelodicMinor(5) and NaturalMinor(5)) with Root Trans, Root Fixed and Guitar NTR.
  - It used src_type **0–33**, including sources recorded over 30/31/32/33, which the unit test skips. It also used src_root 0 and 5, all 12 target roots, and keys 24–99.
  - Results: 1+8 gives only the root. 1+5 gives only {root, 2nd, 4th, 5th}, and Chord NTT gives only {root, 5th}. There were zero violations and no panics.
- **N2 (Ending under Cancel) / N3 (vacuous pitch checks):** done and verified. The reference-run check reads a set chord-mute bit correctly.
- **N1 (#46), N4 (declined, NTT fidelity), N5:** handled properly.

Fidelity: I re-checked OM p.46 (manual text around line 2249). It says Cancel creates "a state in which no chord is input", and the manuals say nothing more about per-channel behaviour. Treating Cancel as the engine's no-chord state is the most defensible reading.

### Blocking

None.

### Non-blocking

**N1. `catch_up` can start Fill/Break notes from before the section's entry point, so a late chord no longer matches an on-time one** (engine.rs `catch_up`, the `lo` computation and backward scan).

- Fills and Breaks start mid-bar: `queue_fill` sets `sec_start = bar_start` and `seek(at - start)` skips everything before the entry beat.
- `catch_up` scans back to `now - 40 ms` in section-relative ticks, and ignores the seek position. So when a chord lands within 40 ms of a Fill/Break entry, it also scans events the section never played.
- Probe: C at 0 → Cancel at bar 2⅓ → Break pressed early in bar 3 (enters on beat 2) → F at the entry vs. entry + 10/20/35 ms. I compared the notes held at entry + 35 ms. Late-only extras:
  - FunkinStyle: ch13 key 65.
  - GirlsFolk: ch12 key 53. The event is at tick 1884, the entry is at tick 1920 and `lo` is 1853.
  - JumpinJiveOrgan: ch11 key 51.
  - SunsetDriveStyle: ch13 key 51, at tick 952 with the entry at 960.
  - The on-time run never has these notes.
- They are humanised notes played slightly early for the entry beat, so musically they are harmless, maybe even welcome. But the late path now differs from the on-time path, which is exactly what `late_chord_after_no_chord_keeps_the_downbeat` claims cannot happen. The test only covers Main at a barline, so it does not see this.
- Suggested fix: clamp `lo` to the tick where playback in the current section began. Store the `seek` position in `transition`/`load`, or use `seek`'s partition index as a floor for `i`. Alternatively, decide that slightly early entry notes should play on both paths. Either way, add the Break case to the test.

**N2. Caught-up notes past 32 are dropped silently** (`n_buf < buf.len()`). The corpus never hits this: the late-chord test passes on all styles. But a dense guitar strum or pad stack across 8 parts in a 40 ms window could. Consider sizing the buffer from the part count × 8 group size, or add a comment saying the truncation is intended.

**N3. Coverage nit.** `one_plus_five_is_quality_neutral` loops `src_type in 0..30`, so sources recorded over sus4/sus2 (32/33) are never asserted. My sweep shows they are correct today. Extending the range to `(0..34).filter(|&t| t != 30 && t != 31)`, or including 30/31 too, would pin that.

**Verdict:** approve from my side. No blocking findings. N1 is a small consistency fix worth doing now, since it is new code from this PR.

ryangavin:
Round 2 review addressed in 9568c78.

**Addressed**
- **N1 (catch_up before the section entry):** the engine now records the tick each section pass enters at (`entry`, set in `seek` and `start`). `catch_up` clamps its window to that tick, so a Fill/Break that comes in mid-bar never starts notes from before its entry. `late_chord_after_no_chord_keeps_the_downbeat` gained a Break case: Cancel, then Break pressed mid-beat, then F arriving on the entry beat vs. 10/20/35 ms late. It runs over every corpus style. Without the fix it fails on FunkinStyle (ch11 key 53 held only by the late chord); with the fix it passes.
- **N2 (32-entry buffer):** the buffer now holds 64 entries (8 parts x 8 notes, the most a single `catch_up` can produce given its 8-note grouping). A comment records that anything past that is dropped instead of allocated.
- **N3 (test range):** `one_plus_five_is_quality_neutral` now sweeps src_type 0..34, so sus, 1+8 and 1+5 sources are pinned for Chord NTT too. It passes with no code change.

**Declined:** none.

Gates: `cargo build --release` passes. `cargo test --release`: 27 passed, 1 ignored, corpus tests executed. `cargo clippy --release`: no new warnings in touched code. No decisions changed.

ryangavin:
Integrator: rebased onto integration/m1-chord-recognition (after #1, #2, #3).

- **Conflict in src/sim.rs**: both sides appended tests at the same spot. Kept #2/#3's `corpus_every_chord_type` and all of this PR's tests (`chord_cancel_leaves_only_rhythm`, `chord_cancel_does_not_sync_start`, `corpus_one_plus_eight_and_one_plus_five`, `late_chord_after_no_chord_keeps_the_downbeat`) unchanged.
- **Build fix**: #3 added `Step::Release`, so the hand-written `Step` copy in `late_chord_after_no_chord_keeps_the_downbeat` no longer compiled. `Step` is `Copy`, so it is now `intro.clone()`. No change in behaviour.
- **theory.rs / engine.rs** merged cleanly. `plays`/`transpose` call `chord.casm()` (from #2) before this PR's Cancel check, so the display-only types (M7b5, (b5), mM7b5) still map first and Cancel still returns `None`.
- **Goldens regenerated** (`UPDATE_GOLDEN=1`). Only the `C1+5` bar (17) changed, plus the note it carries into bar 18:
  - BluesOrganTrio ch11 bar 17 beat 4: A0 -> G0. This PR's 1+5 rule sends the 6th to the 5th.
  - OrganCruise ch11 bar 17: the A0s at beat 1 and beat 4.144 become G0 (same rule). Beat 4 now has two G0 notes, so the listing pairs their note-offs in the other order (durations 480/532 appear swapped); this is the same-key collapse already filed as #46.
  - OrganCruise ch11 bar 18: the short G0~52 retrigger at the Bm7b5 downbeat is gone. The note held over from bar 17 is already G0, and the Bm7b5 revoice maps it to G0 too, so it keeps sounding and is not retriggered.
- No `#[ignore = "M1 #4"]` rows were pending (`PENDING` is empty).
- Gates: `cargo build --release` passes. `cargo test --release`: 135 passed, 1 ignored (synth loudness probe); the corpus is symlinked and no tests were skipped. `cargo clippy --release` shows the same warning set as the integration base.

PR #44, opened 2026-09-23
Recognizer: match the Data List Fingered chord table exactly (#2)

ryangavin:
## What

The Fingered / Fingered On Bass recognizer in `src/theory.rs` is now driven by the Data List p.45 table ("Chord Types Recognized in the Fingered Mode"). It covers all 37 rows plus Cancel.

- **`FINGERED` table:** each row lists its required intervals and its optional (parenthesised) intervals. A pitch-class set matches a row when it contains every required note and nothing outside the row, so any subset of the optional notes may be left out. The lookup table is still precomputed at load: 4096 masks × 12 lowest notes, with no allocation, locking or panics on the MIDI thread.
- **Chords with no MIDI code** (M7b5, (b5), mM7b5) get display ids 35, 36 and 37, named `maj7b5`, `(b5)` and `mMaj7b5`. `casm_type()` / `Chord::casm()` map them to a CASM type for chord mute, the NTT tables and Guitar voicings. `chord_tones()` (used by Stop ACMP) still returns the notes actually played.
- **1+8:** needs the key count. The new `Recognizer::recognize_keys(mask, low, keys, on_bass)` returns 1+8 when two or more keys share one pitch class. `live.rs` counts the held keys and calls it.
- **Signature:** `recognize(mask, low)` keeps its signature. It is Fingered On Bass with one key per pitch class.
- **Fingered vs On Bass:** `on_bass = false` gives plain Fingered, where the bass is the root and a stray bass note does not make a chord. This is a hook for #3's fingering-mode dispatch. The dispatch itself is not implemented here.
- **1+5 display:** 1+5 now reads `C1+5`, the Data List display, instead of `C5`.
- **Spec:** the recognition rules are written up in `docs/genos-features.md` §C.4a, with a note under the §D table.

Not in scope: what the band plays under Cancel, 1+5 and 1+8 (#4), and the fingering-mode enum (#3).

## How tested

- One `#[test]` per Data List row (`dl01_one_plus_eight` … `dl38_cancel`). Each one covers every listed voicing, full and with optional notes left out, in all 12 roots and in both Fingered modes, and also checks the display name.
- `table_sweep_inversions_are_consistent` runs every row, root and omission subset in every inversion. The recognised chord must hold exactly the played pitch classes, the bass must be right, and a root in the bass must always be read as that root.
- Ambiguity tests cover C6/Am7 (including omitted notes), m6/m7b5, dim7/aug symmetry in every inversion, 7b5 tritone symmetry, sus4/sus2 and 6(9)/m7(11).
- Other tests: two-note inputs, On Bass slash chords with a bass outside the chord, and the CASM mapping (chord-mute bit, pack/unpack of ids 35–37).
- New corpus test `sim::corpus_every_chord_type`: every corpus style (36 files, which ran) plays under all 38 chord ids. Each non-Cancel chord must produce notes, and each display-only type must give **identical** engine output to the CASM type it maps to.
- Gates: `cargo build --release` passes. `cargo test --release` gives 66 passed and 1 ignored (after review round 1: still 66 passed, new slash-fallback assertions added); the ignored one was already ignored, and the corpus tests ran. `cargo clippy --release` shows no new warnings in touched code; the `live.rs:163` collapsible-if was already there.

## Decisions

- Decision: a single key is not a Fingered chord, and no two-note set is either except 1+5 (fifth either way up) and 1+8. The previous chord stays. Because RM p.9 says the thing that sets AI Fingered apart is that "less than three notes" can indicate a chord, and the Data List's only two-note rows are 1+5 and 1+8. **Behaviour change:** before, C alone gave C, and C+E gave C.
- Decision: 1+8 means two or more held keys that all share one pitch class. Because the name "1+8" is root plus octave, and a lone key is covered by the decision above.
- Decision: ambiguous pitch-class sets are resolved in this order: (1) the root is the lowest note; (2) fewest omitted notes; (3) the lowest note's role, ranked root > 5th > 3rd > b5/#5/4th > 6th/7th > tension; (4) Data List table order. Because rule 1 is the only inversion rule the manuals imply (Fingered On Bass takes the lowest note as bass), and it makes dim7 and aug take the lowest note as root. Rules 2 and 3 pick the reading a player most likely means when the root is not in the bass. Examples: E G A C = Am7/E, G A C E = C6/G, Eb G A C = Cm6/Eb, G C F = Csus4/G. **Playtest:** E G A C = Am7/E (many would say C6/E) and C D G Bb = Bb6(9)/C.
- Decision (consequence of the two-note rule): a single key, or a two-note set other than 1+5/1+8, no longer triggers Sync Start, since Sync Start fires on the first recognised chord. Because OM p.46 ties Sync Start to playing a chord.
- Decision: in Fingered On Bass, a set that is not a table chord but whose notes above the bass form a **complete three- or four-note** table chord (no omitted notes) is read as that chord over the bass (F# C E G = C/F#, F# C Eb G Bb = Cm7/F#). A full-table reading always wins, so D C E G = Cadd9/D. Tension-laden sets do not fall back to an unrelated root: C E G B F and C E G Bb C# Eb are no chord (previous chord stays), not G13/C or Eb7b9/C, matching plain Fingered. Plain Fingered does not recognise the slash-only form. Because On Bass is defined as "the lowest note becomes the bass", slash chords are played as a triad or seventh over a bass, and plain Fingered only knows the table shapes.
- Decision: M7b5 maps to CASM M7(#11) (3), (b5) to 7b5 (21), and mM7b5 to dim (17). Because the rule is to take the smallest CASM type holding every played note, or else the largest one made only of played notes, so the style never drops a played note when a superset exists. It can add one: M7(#11) adds the natural 5th a semitone above the played b5, and 7b5 adds a b7; mM7b5 -> dim drops the M7. M7(#11) is the only CASM type containing C E Gb B. 7b5 is the smallest one containing C E Gb. No CASM type contains C Eb Gb B, and dim is the largest one inside it. These three are displayed as themselves; only the style follows the CASM type.
- Decision: the display name for 1+5 changes from `C5` to `C1+5`, and the new types are named `maj7b5`, `(b5)` and `mMaj7b5`. Because the Data List shows C1+5, and `(b5)` stays unambiguous for `yahaha sim`'s chord parser ("Cb5" would read as root Cb).

Closes #2


ryangavin:

**Verdict: no blocking findings.** I checked out `origin/m1/2-chord-table` (1ce87a0), symlinked the corpus (36 files), and ran `cargo build --release` and `cargo test --release` (66 passed, 1 ignored; `corpus_every_chord_type` and `corpus_full_performance_no_stuck_notes` both ran against the corpus). `cargo clippy --release` shows no new warnings in the touched hunks.

### What I verified
- **The `FINGERED` table against Data List p.45:** I checked all 37 rows and Cancel, including every parenthesised note. They match. In particular, m7(11) has both (2) and (b7) optional and the 5th required, M7(#11) and 7(#11) have (2) optional and the 5th required, and M7aug has (3) optional.
- **Table packing:** `EXTRA_BASS` (bit 15) cannot collide with `NO_CHORD` (0xFFFF), because the bass nibble is at most 11 (0x5800). `pack`/`unpack` hold ids up to 37 in 6 bits. `chord_mute & (1u64 << ty)` is only reached after `casm()`, so it never shifts by 35 or more.
- **RT safety:** `recognize_keys` does a mask, a popcount, a `trailing_zeros` and one table index. It does not allocate, lock or panic. The `unwrap` in `compute` runs only at table build time, with at least 3 upper notes guaranteed.
- **Every consumer of `chord.ty`:** `plays`, `transpose` and `guitar_voicing` go through `casm()`. `engine.rs:466` (Stop ACMP) uses the played tones, as the PR says. `main.rs` `parse_chord` handles `Cmaj7b5`, `C(b5)`, `CmMaj7b5` and `C1+5`. No docs or scripts in the repo still use the old `C5` name.
- **Probe voicings** (a throwaway test, not committed). In each pair below the first result is On Bass and the second is plain Fingered:
  - `F A C D` → F6 / F6
  - `G Bb C Eb` → Cm7/G / Cm7
  - `B C E G` → Cmaj7/B
  - `F# B C E` → Cmaj7b5/F#
  - `E Ab B C` → Cmaj7#5/E
  - `D C E G A` → C6/9/D
  - `C Eb F G` → Cm11
  - `C C G` → C1+5
  - `C# D C` → N.C.
  - `C C# D G` → none

  All of these follow the documented rules.

### Blocking
None.

### Non-blocking
1. **The fallback in `src/theory.rs` `compute()` (a table chord over a non-chord bass) can name an unrelated upper structure when the player simply plays an extension the table lacks:**
   - `C E G B F` (Cmaj7 add11) → **`G13/C`** (E F G B = G7(13) with no 5th).
   - `C E G Bb C# Eb` (C7(b9#9)) → **`Eb7b9/C`**.
   - Both give no chord in plain Fingered, so the two modes diverge a lot.

   This matches rule 6 as written. Still, a player would expect "keep the previous chord" or a C-rooted reading here, not a new root. Consider using the fallback only when the bass is **not** a chord tone of the result's root reading, or only when the bass is not a 3rd, 5th or 7th of the upper chord. At minimum, add these cases to the Decisions so the owner playtests them.
2. **The rationale for the CASM mapping is overstated:**
   - The docs say the mapping means "the band never adds a note that clashes with what you play" (`casm_type` doc comment and §C.4a rule 7). That is not true for M7b5 → M7(#11): TONES\[3\] contains the natural 5th (G) a semitone above the player's F#.
   - (b5) → 7b5 adds a Bb the player did not play.

   The choices are defensible, since no CASM type avoids adding something. The wording should say "smallest superset" rather than "never clashes".
3. **`sim::corpus_every_chord_type` passes vacuously when the corpus is missing.** It gives no skip message and has no assertion that it ran, unlike `corpus_full_performance_no_stuck_notes`. Add the same `eprintln!("no corpus; skipping")` guard, or `assert!(!files.is_empty())` behind an env flag.
4. **The `assert_eq!(out, play(casm_type(ty)))` equality holds largely by construction,** because `plays` and `transpose` call `casm()` on entry. It is a fine regression guard, but it does not show that the mapping is musically right. Just don't count it as fidelity evidence.
5. **Sync Start behaviour change.** A single key, or two-note sets like C+E or C+Bb, no longer arm or trigger Sync Start, because `engine.rs:445` only fires on a recognised chord change. OM p.46 says the style starts when "you play a chord with your left hand", so this is consistent. But it is a user-visible change from the current build and is not listed under Decisions. Please add it so the owner can playtest it.
6. **Some rule-3 results may surprise pianists:**
   - `E G A C` → Am7/E. Many players would call this C6/E.
   - `C D G Bb` → Bb6/9/C, not a C-rooted reading.

   Both are table-exact and documented, so this is only a playtest flag.
7. **Merge risk with #3:** `recognize_keys(mask, low, keys, on_bass: bool)` adds a mode parameter that #3's fingering-mode enum will probably replace. The `on_bass` bool and the `EXTRA_BASS` flag are small and easy to rebase. Just coordinate so #3 does not duplicate the Fingered/On Bass split.

ryangavin:
Review round 1 addressed in 1a95391.

**Addressed**
- Slash fallback producing unrelated roots: On Bass now only reads "chord over a foreign bass" when the notes above the bass form a *complete* three- or four-note table chord. C E G B F and C E G Bb C# Eb are now no chord in both modes (previous chord stays), not G13/C or Eb7b9/C. F# C E G = C/F# and F# C Eb G Bb = Cm7/F# still work. Tests added in `on_bass_slash_chords`.
- "Never adds a clashing note" wording: `casm_type` doc and spec §C.4a rule 7 now say the mapping is a smallest superset that can add a note (natural 5th for M7b5, b7 for (b5)); mM7b5 -> dim drops the M7.
- `corpus_every_chord_type`: now prints `no corpus; skipping` and returns on an empty corpus, like the other corpus test (it ran here on 35 styles).
- Regression-guard note: added a comment saying the equality check holds by construction and is a regression guard, not a fidelity check.
- Sync Start consequence: recorded in the PR Decisions and in spec §C.4a rule 2.
- Rule-3 readings (E G A C = Am7/E, C D G Bb = Bb6(9)/C): flagged for playtest in the PR Decisions and in the spec.

**Declined**
- Changing the `on_bass: bool` parameter now: #3 owns the fingering-mode enum. Changing it here would clash with #3. The merge note stands: #3 should replace `on_bass` with its mode enum in `recognize_keys` and not add a second entry point.

Gates: `cargo build --release` OK; `cargo test --release` 66 passed / 1 ignored (the ignored one was already ignored), and the corpus tests ran; no new clippy warnings in touched code.

ryangavin:
## Integration: rebased onto `integration/m1-chord-recognition` (after #1 / #45)

The rebase had no textual conflicts. #1's golden harness needed these follow-ups, all in `4ff638c`:

**Recognizer golden (`src/recognizer_golden.rs`)**
- Rows M7b5, (b5) and mM7b5 now use #2's display ids (`M7B5`, `FLAT5`, `MM7B5`) in place of `None`. `table_is_consistent` now expects types 0..=37.
- The harness's `recognize()` calls `recognize_keys(mask, low, key_count, on_bass = true)`, so 1+8 sees the key count.
- Un-ignored the tests this ticket makes pass: rows 1, 4, 6, 7, 11, 14, 21, 22, 30 and `off_table_inputs`. `PENDING` is empty. All of them pass. `ambiguous_voicings` and `octave_doublings_keep_the_chord` also pass under #2's priority rules.

**Style snapshots (`tests/golden/*.txt`)**
- #2 renamed the 1+5 type suffix from `5` to `1+5` to match the Data List display. `C5` in `chords.script` and in `script_grammar` stopped parsing, so both now say `C1+5`. The README example has the same change.
- Regenerated with `UPDATE_GOLDEN=1`. The only changed line group in all 6 snapshots is the bar 17 header, `bar 17  Main B  C5@1.0000` → `C1+5@1.0000`. No note output changed.

**Gates:** `cargo build --release` passed. `cargo test --release`: 115 passed, 0 failed, 1 ignored (the synth test, which was already ignored). The corpus tests ran and none were skipped. `cargo clippy --release` shows no warnings in the touched files.

Issue #10, opened 2026-09-23
RTR: real Pitch Shift and Note Generator


ryangavin:
- Pitch Shift should bend the sounding note (not note-off/note-on). Implement Pitch Shift to Root the same way, and add Note Generator. Currently all of these retrigger.

---
Milestone: **M2 · Chord-to-note conversion 1:1** · Spec: [docs/genos-features.md](../blob/main/docs/genos-features.md) · Backlog: [docs/backlog.md](../blob/main/docs/backlog.md)

ryangavin:
Decisions made in #63 (the owner playtests and may overrule):

- **Decision (updated after review round 1): Pitch Shift and Pitch Shift to Root bend with MIDI pitch bend. Each chord-following part (ch 11–16) gets an output bend range of max(style's own, 12, min(widest pattern bend + 12, 24)) via RPN 0. The pattern's own bends are rescaled from the style's range. A shift that would not fit on top of the pattern's widest bend falls back to a retrigger, so nothing is ever clamped. Reasons:**
  - bend is the only way to move a sounding note on a MIDI channel without a new attack;
  - 12 covers every to-Root move (±6) and nearly every Pitch Shift;
  - it is already the most common range styles set: RPN 0 in SInt is 12 ×70, 24 ×16, 8 ×8, 4/3 ×2 each, 0 ×3;
  - a Genos Style part takes RPN 0 up to 24 (DL p.98: Pitch Bend Sensitivity 00H–18H; MIDI Implementation Chart: 0–24 semi).

  An earlier version of this comment cited RM p.140 for "12 is the Genos maximum". That page is the keyboard parts' wheel range, so the citation is withdrawn. Instruments that ignore RPN need their range set by hand: 12, or 24 to be safe.
- **Decision: when the notes on one part need different shifts, the part bends by the shift the most continuing notes need (a held note votes for 0), and the other notes are retriggered at their new pitch, because:**
  - pitch bend is per channel;
  - Yamaha shifts each note inside its tone generator, and the manual doesn't say what MIDI would carry;
  - this keeps every pitch exact with the fewest new attacks.

  Notes started while a part is bent are sent compensated. The bend returns to centre at the part's next note once it has fallen silent, also when nothing continues through a change, so release tails keep their pitch. A note about to end has no vote. In a corpus-wide run, 15,797 Pitch Shift notes bent and 11,140 fell back to a retrigger. Attacks at chord changes fell from 42,303 to 27,119.
[...]

PR #45, opened 2026-09-23
Golden harness: Data List recognizer table and sim snapshots (#1)

ryangavin:
## What

This PR adds the golden test harness for M1, so any change to chord behaviour shows up as a diff that someone reviews.

**Recognizer table** (`src/recognizer_golden.rs`)
- One `#[test]` per Data List chord row (docs/genos-features.md §D, all 38 rows including Cancel). Each row is played in all 12 roots, in every inversion and with every combination of the optional "(…)" notes left out. The test asserts the root, type and on-bass note.
- A failure lists every wrong voicing, for example `row 21 m7(11) on C: [C2 D2 Eb2 F2 G2 Bb2]  want Cm11, got Ebmaj9/C`.
- Rows that fail today are `#[ignore = "M1 #2"]`, ready for #2 to un-ignore:
  - 1+8
  - 6 (C G A reads as Am7/C)
  - M7(#11) and 7(#11): `TONES` has no 9th
  - M7b5, (b5) and mM7b5 (no chord code yet; M7b5 C E F# B currently reads as Cmaj7#11)
  - M7aug without its 3rd
  - m7(11)
- `ambiguous_voicings` covers sets that could be several chords (for example E G A C). It passes today and only requires a valid reading, so #2 can tighten it.
- `octave_doublings_keep_the_chord` checks that doubling notes in other octaves does not change the chord. For every row that is not ignored it also checks that the doubled voicing gives the table's chord (`PENDING` lists the ignored rows).
- 1+8 is played in one, two and three octaves of the same key. Cancel is played only as the close 1+b2+2 cluster.
- `off_table_inputs` (ignored for #2) expects no chord for a single key, two-key intervals other than 1+5 and 1+8, clusters, and Cancel with an extra key.
- All test input goes through one `recognize(&Recognizer, &[notes])` helper, so #2 and #3 can change the recognizer API in a single place. For example, 1+8 needs the note count.

**Sim snapshots** (`src/sim.rs`, `src/golden.rs`, `tests/golden/`)
- `sim::parse_script` reads a small script grammar: `|` bars, chords shared evenly across the bar, `-` holds, `^` releases the chord (Sync Stop), and `[MainB]`-style buttons.
- `sim::snapshot` produces a stable listing, bar by bar and part by part. Each bar header shows the section and any section changes, followed by each part's notes as `beat.tick Note~length`.
- Parts that play as written whatever the chord only get a per-bar count (`N as written`): the drum parts, and any part that is Root Fixed/Guitar + Bypass for every key in that section. No drum or other untransposed style pattern is stored.
- `yahaha sim <style> <"C Am F G7" | script file>` prints this listing. The old one-chord-per-bar form still works.
- `golden_snapshots` plays `tests/golden/chords.script` on six styles: BluesOrganTrio, JackDoesItAgain, OrganCruise, SmoothItOver, TickingAway (SFF1) and AustinCityBlues (no CASM). The script covers:
  - major, minor, 7, dim/dim7, aug and 7#5, sus2, sus4 and 7sus4, 1+5, m7b5, mMaj7, m(add9), tension chords, and slash chords
  - Intro, Main A→B→C→D through fills, Break, and Ending
- `UPDATE_GOLDEN=1` rewrites the stored listings. A mismatch prints a line diff grouped by bar, with the changed notes called out (compared as a multiset).
- `| |` on one script line is an error, so an empty bar can no longer silently shift every later bar.
- The repo stores only our script's output for the chord-following parts, keyed by style file name. If the corpus is missing, the test skips.
- Usage is documented in `tests/golden/README.md`.

## Known behaviour pinned by the snapshots

- Zero-length notes such as `D3~0` (about 72 across the files) are chord-change retriggers that the pattern cuts on the same tick. This is an engine bug, tracked in #47. When it is fixed, the snapshots will change and must be regenerated.

## Merge note

- **Squash-merge this PR.** Commit 716ed4d still contains the first, larger snapshots, which include verbatim drum pattern data. Squashing keeps them out of the integration branch history.

## How tested

- `cargo test --release`: 58 passed, 11 ignored (9 recognizer rows and `off_table_inputs` for #2, plus 1 that was already ignored). The corpus was symlinked; the corpus tests pin exact section changes on real styles, so they cannot pass by skipping.
- `snapshots_do_not_list_parts_played_as_written` checks the stored files (even without the corpus) and a fresh OrganCruise run, where Phrase 1 is Root Fixed + Bypass in the fills.
- `cargo clippy --release --all-targets` shows no new warnings in the code I touched.

## Decisions

- Decision: expectations follow Fingered On Bass, where the lowest note becomes the on-bass note unless it is the root, because that is what the recognizer implements today. #3 adds plain Fingered, where the bass is always the root.
- Decision: when a pitch-class set spells more than one chord, the reading whose root is the lowest note wins. So C E G A = C6, A C E G = Am7, C G A = C6 (not Am7/C), and C Eb Gb A = Cdim7. I chose this because the Data List gives every shape in root position and says nothing about inversions. Yamaha's own Fingered On Bass treats the lowest note as the bass, so a root in the bass is the natural reading. Sets with no reading rooted on the lowest note, or with several, are "ambiguous" and left to #2.
- Decision: the 3 uncoded rows (M7b5, (b5), mM7b5) fail until #2 fills in the chord type they play as, because accepting any type with the right root let M7b5 pass as Cmaj7#11, a chord whose 5th is not played.
- Decision: off-table note sets (a single key, two keys other than 1+5/1+8, clusters, Cancel plus a key) should give no chord, because the Data List is the complete Fingered table and the Reference Manual (p.9) names "less than three notes" as the one thing AI Fingered adds over Fingered. The test is ignored for #2 to make pass.
- Decision: in snapshot scripts, buttons press one tick before their slot, because a press exactly on a beat or bar line would otherwise depend on float rounding in the engine ("this beat or the next?"), which made fills start a beat later on some styles. The listing shows the button at that press tick, so a button on a bar line shows at `4.1919` of the bar before, next to the section change it causes.
- Decision: parts that play as written (drums, Root Fixed/Guitar + Bypass for every key) are only counted per bar, because listing them would transcribe the style's own patterns and they do not follow the chord. A per-bar count still shows sections starting, stopping and switching.
- Decision: snapshots list note starts and lengths but not velocities or pitch-bend messages, because those would add noise before RTR Pitch Shift (#10) is real. Lengths already expose Stop and Retrigger behaviour.
- Decision: Cancel is tested only as the three adjacent keys 1+b2+2. The Data List gives only that shape, and whether spread or inverted clusters also cancel is for #4 to decide.
- Decision: `| |` on a single script line is an error. A line that ends in `|` followed by a line that starts with `|` is still the normal layout.
- Decision: `#` starts a comment only at the start of a token, because `F#m7b5` must stay a chord.

Closes #1


ryangavin:

I checked out `716ed4d` with the corpus symlinked and ran `cargo build --release` and `cargo test --release`. Everything passes: 55 passed, 9 ignored. `golden_snapshots` really compared all six styles and printed no "skipping" lines. The ignored rows fail as described when run with `--ignored`. The clippy warnings at `sim.rs:249` and `main.rs:138` were already there before this PR; they only moved lines.

I checked the chord table in `recognizer_golden.rs` against the Data List. All 38 rows match DL p.45 for tones and for which notes are optional, and all MIDI `ct` codes match DL p.111. I found nothing wrong there.

### Blocking

**B1. The golden listings commit third-party style data word for word, and the repo is public.**
- The drum parts are `ch9`/`ch10` (RootFixed/Bypass), and Bypass phrase parts behave the same way. Their lines copy the style's MIDI pattern note for note: every hit's position, key and length, across Intro, Main A–D, the fills and Ending, for six MOX styles.
- Measured: 100,489 of the 193,145 bytes in `tests/golden/*.txt` are `ch9`/`ch10` lines. That is 52% of the stored snapshots, and it is style data, not our script's output.
- It conflicts with the rule "never commit style data". It also adds nothing to chord coverage: drums do not follow the chord, and the bar headers already show section timing.
- Fix: leave the Rhythm1/2 channels out of `sim::snapshot`. A per-bar note count is enough if you want to catch section or timing regressions. Also consider leaving out any part whose CASM for the section is Root Fixed plus Bypass (the pass-through case in spec §C.5), because those lines are verbatim too.
- If the owner explicitly accepts committing this, downgrade B1 to nonblocking. But it should be a conscious decision, not a side effect.

### Nonblocking

**N1. The row 6 M7b5 test pins nothing about the chord type.**
- It "passes" today because C E F# B is recognised as `Cmaj7#11`, a different coded type (ct 3).
- For rows with no code, `check_row` only compares root and bass. So *any* type, including plain major, would pass.
- Suggestions:
  - List the candidate types (for example ct 3, or "no chord") so a regression to something absurd fails.
  - Or state in the Decision that row 6 passing is accidental.

**N2. There are no negative or off-table tests.** The spec says Fingered recognises only the Data List shapes, but the harness never checks what happens with other input. My throwaway probe against today's recognizer:
- `[C2]` (a single key) gives `C`, a major chord. That is not a Fingered shape: it is Single Finger or Multi Finger behaviour.
- `[C2 C#2]` gives `None`.
- `[C2 D2 G2 A2]` gives `D7sus4/C`.

Please add a small off-table table marked `#[ignore = "M1 #2"]` so #2 has to decide these cases:
- a single key
- two-note sets other than 1+5 and 1+8
- octave-doubled voicings (for example C2 G2 C3, which should be 1+5)
- clusters with four or more notes
- Cancel with an extra note

Doubled voicings will matter as soon as the helper passes a note count for 1+8. Today `cases()` only builds close-position voicings with no doubled notes.

**N3. The Jack snapshot bakes in a doubtful engine behaviour.**
- JackDoesItAgain has no Main D.
- Yet `JackDoesItAgain...txt` bar 32 plays `Main C > Fill In DD@4.0000` for `[MainD]` and then goes back to Main C.
- On Genos, a section button with no data is unlit and does nothing, so I would expect no Fill In DD here.
- The golden snapshot now blesses this. Please either raise it as an engine issue or note it in the PR, so nobody reads it as "verified".

**N4. There is no key-release token in the script grammar.**
- `[SyncStop]` can't actually be exercised: Sync Stop needs all keys released, and the script has no release step.
- Stop ACMP (§C.8) and chord memory on release can't be scripted either.
- An `N.C.` or `off` token, meaning "release all chord keys", would let #4, #5 and later tickets add golden coverage.

**N5. The one-tick-early button press hides a timing edge.**
- The Decision is defensible, but the PR says a press exactly on the beat made fills "start a beat later on some styles".
- That comes from `ns()` using `ceil`, which lands up to 1 ns after the engine's own beat deadline, so the engine has already started that beat.
- Suggestion: add one explicit script test that presses exactly on a beat, so the boundary behaviour is pinned rather than dodged.
- Also, in the listing, `[Stop]@1.0000` shows in bar 2 while the stop actually happens at bar 1 `4.1919` (`yahaha sim FunkyFinger "| C | [Stop] | C |"`). That is mildly confusing.

**N6. All six styles are from `corpus/MOX_v2`, the only pack available.**
- They are all 4/4, five of the six are ppq 1920, and most have 1-bar intros and endings.
- That's fine given the corpus, but the README's "diverse" claim should mention it. Add a 3/4 style or a Genos-native SFF2 style when one becomes available.

**N7. `snapshot_is_deterministic` asserts very little.**
- It runs the same thing twice in one process, so it can only catch hash-order or global-state leaks.
- It could be removed, or changed to compare a snapshot before and after an unrelated `Style::load` of another file.

**N8. The diff summary uses sets.**
- `diff()`'s `^` line compares notes as sets, so a note that appears twice (`3.0000 C2~0  3.0000 C2~0`) losing one copy produces no summary.
- This is minor, because the `-`/`+` lines still show the change.

**N9. Nits**
- The labels re-spell chords from the parsed value, so script `Dbmaj7` is listed as `C#maj7`. That makes it harder to match the listing back to the script.
- `[Start]` maps to `Button::StartStop`, so `[StartStop]` would be clearer.
- The PR body says both "Closes #1" and "Part of #1". Also, "Closes" won't fire when merging into a non-default branch.
- The branch `if i == 0 && t == 12` in `cases()` can never run, because `inversions == 1` for 1+8 means `i == 0` always has `t == 0`.

### Decisions check
- **Fingered On Bass as the baseline.** OK, and it matches RM p.9. Watch out that #3 will need a mode parameter in `recognize()`.
- **Root-in-bass disambiguation (C G A = C6).** OK. It follows the Data List, where the 3rd of the 6 chord is optional, and the spec (§C.4) says we must define priorities ourselves.
- **Unnamed types pin only root and bass.** OK, but see N1.
- **Buttons one tick early.** Acceptable, but see N5.

Verdict: **changes requested**, for B1 only. The harness itself is solid, and apart from B1 the remaining work is small.

ryangavin:
Review round 1 addressed in fa27598.

**Blocking**
- **B1 (fixed):** `sim::snapshot` now lists notes only for parts that follow the chord. The drum parts, and any part that is Root Fixed/Guitar + Bypass for every key in that section (`PSection::plays_as_written`, the same test `theory::transpose` passes through on), only get a per-bar count such as `ch10 Rhythm2 14 as written`. This also catches OrganCruise's Phrase 1 in the fills. The stored snapshots shrank from 193 KB to 98 KB. `snapshots_do_not_list_parts_played_as_written` checks the stored files (even without the corpus) and a fresh OrganCruise run. Note: the first commit on this branch still has the old files. Squash-merge keeps them out of the integration history; I can also rewrite the branch if you want them gone from the PR too.

**Nonblocking**
- **N1 (fixed):** The uncoded rows no longer pass on root and bass alone. Row 6 (C E F# B read as Cmaj7#11, which needs a G that is not played) now fails and is `#[ignore = "M1 #2"]`, like (b5) and mM7b5. #2 fills in their type.
- **N2 (done):** Added `off_table_inputs`, ignored for #2. It expects no chord for a single key, two-key intervals other than 1+5 and 1+8, clusters, and Cancel plus a key (96 of 168 inputs give a chord today). The basis: RM p.9 says "fewer than three notes" is what AI Fingered adds over Fingered. Added `octave_doublings_keep_the_chord`, which passes today. [C2 D2 G2 A2] is left out on purpose: D G A C is exactly D7sus4, so D7sus4/C is a correct On Bass reading.
- **N3 (declined):** The harness should record what the engine does, so a fix to the Main D fallback shows up as a reviewed snapshot diff. Changing the engine is outside #1. I'd suggest a separate ticket for "Main button with no Main section".
- **N4 (fixed):** New `^` token lets go of the chord (`engine.chord_released`) and takes a slot. Also added `[StopAcmp]`. `release_triggers_sync_stop` pins Sync Stop via `^` and the restart on the next chord.
- **N5 (partly):** The listing now shows each button at the tick it was pressed, so `[StartStop]@4.1919` sits next to `> stopped@4.1919` in the same bar. `buttons_count_for_their_beat` pins that a button written on a beat fires for that beat. I did not change the engine's `queue_fill` ceil at an exact-nanosecond boundary: that is engine timing, outside #1, and a live MIDI press never lands on the exact nanosecond.
- **N6 (declined):** MOX_v2 is the only pack in the local corpus, and style data can't be committed. `STYLES` is one line to extend when more packs arrive.
- **N7 (fixed):** `snapshot_is_deterministic` now loads the style twice and asserts that the intro, the fill's position, the ending, the stop, and the bass and chord parts are all present.
- **N8 (fixed):** The diff summary compares notes as a multiset, and an empty side prints `nothing`. Covered in `diff_is_readable`.
- **N9 (fixed):** Chord labels keep the script's spelling (`Dbmaj7`). `[Start]` is now `[StartStop]` (a toggle). The PR body no longer says "Part of #1". Removed the dead `i == 0 && t == 12` branch.

Gates: `cargo build --release` passes. `cargo test --release`: 58 passed, 11 ignored, with the corpus symlinked. `cargo clippy --release --all-targets` shows no new warnings in touched code. I updated the Decisions section of the PR body: the uncoded rows, off-table input, the button display tick, and counting parts played as written.

ryangavin:

I checked out `fa27598` (detached, with the corpus symlinked) and ran the gates.
- `cargo test --release`: 58 passed, 11 ignored.
- `cargo test --release golden -- --nocapture` printed no `skipping` lines, so all six styles really compared.
- `cargo clippy --release --all-targets` gives no warnings in `golden.rs`, `recognizer_golden.rs`, `sim.rs` or the new `PSection::plays_as_written`. Every remaining warning was already there before this PR.

I also rechecked the table against the Data List text myself (DL p.45), including the "Notes in parentheses can be omitted" footnote. All 38 rows match.

### Round 1 fixes: verified
- **B1 (fixed).**
  - Every `ch9`/`ch10` line in the six stored files is `N as written` with no `~`.
  - `plays_as_written` uses the same condition as `theory::transpose` (`Bypass && ntr != RootTrans`, drums always).
  - A part with no rules is vacuously "as written", but that is harmless: the engine only sounds notes that have a rule (`engine.rs:758`).
  - The "Guitar + Bypass" wording is dead in practice, because `decode_ntt_new` never returns `Bypass` for `Ntr::Guitar`.
  - The OrganCruise Phrase 1 case is pinned by a test.
- **N1 (fixed).** Uncoded rows now fail unless #2 supplies a `ct`. Running the ignored tests confirms M7b5 → `Cmaj7#11` is reported.
- **N2 (fixed).** `off_table_inputs` fails as intended (96 of 168 give a chord). The rationale matches RM p.9: only AI Fingered accepts "less than three notes". Excluding the D7sus4/C case is correct.
- **N4 (fixed).** `^` and `[StopAcmp]` parse, and `release_triggers_sync_stop` pins stop-on-release and restart-on-press, which matches Genos Sync Stop. I also checked by hand that `| C | ^ | F |` without Sync Stop keeps the bass on C through bar 2 (chord memory). That is correct.
- **N5, N7, N8, N9 (fixed as described).** The button now shows at its press tick, next to the section change it causes (`buttons_count_for_their_beat`).
- **N3 and N6:** the reasons for declining are fine, so I'm not re-raising them.

### Blocking
None.

### Nonblocking
**N10. The history still carries the verbatim drum listings.**
- `716ed4d` has the 193 KB snapshots, and about 100 KB of that is `ch9`/`ch10` pattern data.
- A merge commit or rebase-merge into `integration/m1-chord-recognition` would bring it into that branch's history, and the repo is public.
- Please **squash-merge** this PR. Better, force-push a rewritten branch now, so the old files are not the PR's reachable head history. They stay fetchable by SHA on GitHub either way, so this reduces exposure rather than removing it.

**N11. The snapshots now expose (and bless) zero-length notes from retriggering at a chord change.** This is an engine issue for a separate ticket, not a harness bug. Examples:
- `JackDoesItAgain` bar 37 `ch12 ... 1.0000 D3~0`
- bar 38 `ch13 ... 1.0000 Eb3~0  1.0000 Eb3~1920`
- 35 such notes in Jack, 24 in BluesOrganTrio, 8 in TickingAway and 5 in OrganCruise

What happens:
- `sim` applies the chord before `process()` at the same instant, which is realistic because a player hits the chord slightly early.
- The chord change retriggers a sounding note whose pattern note-off is due on that same tick.
- So the engine sends note-on followed immediately by note-off. On a real synth that is an audible blip, or two note-ons on the same key.
- Real Genos would not retrigger a note that ends at that instant.

Please raise a ticket, and mention in the PR body that the `~0` entries are known engine behaviour, not something the snapshot verified.

**N12. `octave_doublings_keep_the_chord` is a relative check, so it can't catch a wrong chord.**
- It compares close and doubled voicings with each other, not with the table.
- Rows that are wrong today pass it: M7(#11) reads as `Em9/C` both ways.
- That's fine as an invariant. Consider also asserting `doubled == c.expected()` for rows that aren't ignored, so it proves "doubled gives the right chord" rather than "doubled gives the same chord".

**N13. Cancel inversions are pinned without a Decision.**
- `cases()` inverts row 38, so `[C#2 D2 C3]` and `[D2 C3 C#3]` must read as Cancel.
- The Data List only gives `1+b2+2`, and a spread voicing is not three adjacent keys.
- The recognizer does this today, so nothing fails. It is still a behaviour choice #2 will inherit silently, so either add a Decision line or restrict row 38 to the close cluster.

**N14. Script edge case:** `| | C |` quietly drops the empty bar (bars 1). That is consistent with "a bar needs at least one chord or `-`" in the README. An error would be friendlier than silently shifting every later bar, but this is minor.

**N15. 1+8 is only generated as `[k, k+12]`.** Once #2 widens `recognize()` to take a note count, consider adding `[k, k+24]` and `[k, k+12, k+24]` so #2 has to decide those too.

### Decisions check
- The Fingered On Bass baseline, "rooted reading wins" (C G A = C6), the uncoded rows failing, off-table giving no chord, pressing buttons one tick early, and counting parts played as written: all are well reasoned and consistent with the RM and Data List.
- I found no fidelity errors in the stored expectations.

Verdict: **approve from the adversarial side**. There are no blocking findings. Please squash-merge (N10) and file the retrigger blip (N11) as an engine ticket.

ryangavin:
Round 2 review, addressed in c679278:

Blocking: none.

Addressed
- **N12:** `octave_doublings_keep_the_chord` now checks every passing row against the table (`doubled == expected`), not only against its own close voicing. The rows still ignored for #2 are listed in `PENDING`. As a check, I took row 7 out of `PENDING` and 60 doubled voicings failed, for example `want Cmaj7#11, close Em9/C, doubled Em9/C`.
- **N13:** Cancel now generates only the close 1+b2+2 cluster, so the test no longer requires spread shapes such as [C#2 D2 C3] to cancel. `table_is_consistent` asserts this, and I added it to Decisions.
- **N14:** `| |` on a single script line is now an error that suggests `| - |`. A line that ends in `|` followed by a line that starts with `|` still parses. There are tests for both cases, and the README is updated.
- **N15:** row 1 (1+8) now plays [k, k+12], [k, k+24] and [k, k+12, k+24]. The helper already passes full notes, so #2 needs no change here.
- **N11:** filed as #47 (the ~72 zero-length retrigger notes). The PR body now has a "Known behaviour pinned by the snapshots" section that points to it.
- **N10:** the PR body now asks for a **squash-merge**. I did not force-push a rewritten branch: GitHub keeps the old commits reachable through `refs/pull/45/*` either way, and squashing keeps them out of the integration branch.

Gates: build passes; `cargo test --release` gives 58 passed and 11 ignored, with the corpus symlinked and snapshots checked on all six styles; clippy shows no new warnings in the touched files.

Issue #47, opened 2026-09-23
Engine: chord change retriggers notes the pattern cuts on the same tick (zero-length notes)


ryangavin:
Found while reviewing the golden snapshots in #45.

When a chord change lands on the same tick as a pattern note-off (typically a bar line), the engine retriggers the sounding note at its new pitch and then the pattern's own note-off (or a new note-on for the same key) cuts it immediately. The snapshots record these as zero-length notes:

- `tests/golden/JackDoesItAgain.S930.STY.txt` bar 37 ch12 `1.0000 D3~0`, bar 38 ch12 `1.0000 C3~0`, ch13 `1.0000 Eb3~0  1.0000 Eb3~1920`
- About 72 such notes across the six snapshot files (JackDoesItAgain 35, BluesOrganTrio 24, TickingAway 8, OrganCruise 5).

On a synth this is a note-on/note-off pair on the same tick: at best wasted events, at worst an audible blip or a re-attack of the envelope. Genos RTR should not retrigger a note whose pattern note ends at the chord-change tick.

Fix direction: when applying RTR on a chord change, process pattern note-offs due at that tick first (or skip retriggering notes whose note-off is due on the same tick). Probably belongs with #10 (RTR). Once fixed, regenerate the snapshots with `UPDATE_GOLDEN=1 cargo test --release golden`; the `~0` notes should go.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

ryangavin:
Done in #101, merged to main via #113.

PR #48, opened 2026-09-23
Golden snapshots: commit digests only, keep readable listings local

ryangavin:
## Why

This repo is public. The golden snapshots from #1 (`tests/golden/*.txt`) were readable per-part note listings of six commercial Yamaha styles (S930/T162 from the git-ignored corpus). A bar-by-bar list of the notes each part plays amounts to a transcription of the style's bass and chord patterns, so committing it redistributes copyrighted content. This PR stops committing the listings but keeps the regression value.

## What changes

- **Committed:** `tests/golden/<style>.digest`. It keeps the style and bar header lines as they are, since those come from our own `chords.script` (sections and chords). Each part line is reduced to its key and a 64-bit FNV-1a hash of the line, e.g. `  ch11 Bass    24688a8ae747a169`. No new dependencies are added, and the hash is pinned by a reference-vector test.
- **Local only (git-ignored `tests/golden/local/`):**
  - `<style>.txt` is the full listing, written on every run.
  - `<style>.baseline.txt` is written by `UPDATE_GOLDEN=1`, or by the first passing run on a fresh checkout.
- **On a mismatch** the test names each changed bar and which parts changed (changed, gone or new), with the expected and actual header. When a local baseline exists, it adds a readable note diff of just those bars. It warns if the baseline is from another revision.
- `UPDATE_GOLDEN=1` regenerates the digests and the local baselines.
- The new `committed_digests_hold_no_notes` test checks that every committed part line is a key plus 16 hex digits, that no line contains `~`, and that no readable `.txt` listing is in `tests/golden/`.
- The README explains the design and the reason for it. Its examples now use made-up notes instead of fragments from the real styles.

## Verification

- `cargo build --release`, and `cargo test --release` with the corpus present: 156 passed, 1 ignored (as before). No clippy warnings in `src/golden.rs`.
- Checked with grep: no committed file under `tests/golden` contains a note with timing (`~N`).
- Perturbation check: I added +1 semitone on the Chord NTT path for Eb-rooted chords. The golden test failed for JackDoesItAgain, TickingAway and AustinCityBlues. The report listed exactly the Ebdim7 and Ebmaj7 bars (plus the held-note knock-on in TickingAway bars 11 and 13) and the affected parts, with a note-level diff from the local baseline. I then reverted the change.

## Caveat

Hashes of very short part lines, such as a single note in a bar, could in principle be brute-forced from a small search space. Longer lines cannot be recovered in practice. I kept one hash per bar × part, as the design specifies, so failures point at a single part.


ryangavin:
## Adversarial review

Reviewed at `144721f` with the corpus linked in. `cargo build --release` is clean. `cargo test --release` gives 156 passed, 1 ignored. `cargo clippy --all-targets` shows no warnings in `src/golden.rs`, and the warning count elsewhere is the same as before this PR.

**Verdict: approve. Nothing in this diff is blocking, so I pushed no fixes.** Finding 1 is serious, but no change to this branch can fix it. The owner needs to decide on it before `integration/m1-chord-recognition` goes into `main`.

### Findings

**1. The deleted listings are still public in history. Blocking for the PR's goal, but outside this diff and not fixed here.**
This PR stops committing listings from now on. It does not remove the ones already published. All six full `tests/golden/*.txt` listings are still reachable in three places:
- Commits `1034ed3` (#45), `e047f23` (#44) and `6c6c52a` (#43) on `origin/integration/m1-chord-recognition` and `origin/m1/7-rt-panic`.
- The "Files changed" views of merged PRs #45, #44 and #43.

They are not on `main` yet. To actually withdraw them:
- Rewrite `integration/m1-chord-recognition` and `m1/7-rt-panic` so the `.txt` files never existed, for example with `git filter-repo --path-glob 'tests/golden/*.txt' --invert-paths`, before they merge into `main`.
- Ask GitHub Support to purge the cached PR diffs and dangling commits.

Both steps are destructive or external, so they are the owner's call and I did not do them. PR #45's comments, PR #43's description and issue #47 also quote a few single notes each (for example `D3~0`, `G0~52`). That is de minimis, but they could be edited out while you're at it.

**2. Short part lines can be recovered from their hash in under a second. Nonblocking, but the caveat in the PR description understates it.**
Without using any knowledge of the chord, I brute-forced `TickingAway bar 2  ch11 Bass  24688a8ae747a169` over every beat, tick, key and length up to 4000. A single thread found it in 0.73 s after about 9.9M tries. The whole single-note search space is about 1e9, roughly 70 s. The baseline listings contain 35 single-note part lines and 14 two-note ones, out of about 470 melodic lines. The `N as written` drum counts can be recovered trivially.

These lines carry very little information (one held note, or a hit count), and lines with 3 or more notes are out of practical reach. So I accept the trade-off for the sake of per-part failures. Please reword the caveat from "could in principle be brute-forced" to "single-note lines are recoverable instantly; they reveal only one sustained note".

**3. A passing run never refreshes a stale local baseline. Nonblocking.**
The baseline is only written when none exists. Say someone pulls a commit that intentionally changes the digests. Their tests still pass, but `local/*.baseline.txt` stays at the old revision indefinitely. Every later failure then carries the "from another revision" warning, and the note diff mixes in someone else's changes. I reproduced this by editing one baseline note and running the test: it passed and the file stayed stale.

Suggested one-line change: in the `want_digest == got_digest` branch, rewrite the baseline when `base.is_none() || digest(base) != want_digest`.

**4. The comparison breaks on CRLF checkouts. Nonblocking, and it predates this PR.**
There is no `.gitattributes`. With `core.autocrlf=true` on Windows, the `.digest` files check out with CRLF, and `want_digest == got_digest` fails for all six styles. `compare()` splits with `str::lines()`, which strips the `\r`, so the report says **"0 bar(s) changed"** with an empty body, which is confusing.

The hash itself is fine. Two possible fixes:
- Add `tests/golden/*.digest text eol=lf` to a `.gitattributes` file.
- Compare `want_digest.lines().eq(got_digest.lines())`.

**5. The `corpus` symlink is not ignored. Nonblocking, and it predates this PR.**
The setup (`ln -sfn …/corpus corpus`) creates a symlink, and `.gitignore`'s `corpus/` only matches directories, so `corpus` shows up as untracked. A `git add -A` would commit a symlink to `/Users/…`. That leaks a local path but no content. Change the entry to `/corpus`.

**6. Velocity and CC are not covered. Nonblocking, and it predates this PR.**
Listings have no velocity, so I tested a velocity perturbation: `vel.saturating_sub(20).max(1)` on every note-on passes the golden test. Only velocity 0, which is treated as a note-off, is caught. Volume, CC, program change and bend are not snapshotted either. This is a known limit of the listing format and not this PR's job. It is noted here because "the harness still catches regressions" is only true for pitch and timing.

### What held up

- **Leak sweep of the tree at `144721f`:** I grepped all 31 tracked files for `Note~len` and `beat.tick Note`. The only matches are the synthetic fixtures in `src/golden.rs`, which use a 4-digit tick format with `~460`, and the made-up README examples. I compared them with the real local listings and none of them match. Nothing matched in `src/recognizer_golden.rs`, `src/sim.rs`, the docs, `PLAN.md` or the top-level README. The `.digest` files contain only headers from our own script plus `key + 16 hex` lines, and `committed_digests_hold_no_notes` enforces that even without the corpus.
- **Regressions are caught, with useful reports:**
  - **Single part, single chord:** Bass +1 semitone, only on `Cdim`. The test fails for TickingAway (bar 35, plus the knock-on in bar 34, both `ch12 Chord1 changed`) and for AustinCityBlues (bar 35 `ch12 Chord1 changed`). The note-level diff from the baseline shows exactly C→C#, F#→G. Other styles are unaffected because their Cdim bass is muted or bypassed, which is correct.
  - **Root Fixed octave tie-break flipped:** JackDoesItAgain fails with 8 bars (`ch13 Chord2`) and AustinCityBlues with 1 bar (`ch12 Chord1`).
  - **Velocity driven to 0:** 5 styles fail.
- **Fresh-checkout behaviour:**
  - With `local/` deleted and the code good, the first run passes and writes all six baselines.
  - With `local/` deleted and a regression present, no baseline is written for the failing styles, so a bad baseline can't be captured. The report points to `UPDATE_GOLDEN=1` on the last good revision.
  - `UPDATE_GOLDEN=1` on unchanged code leaves the committed digests byte-identical.
- **Hash stability:** the hash is a hand-rolled 64-bit FNV-1a over bytes. `DefaultHasher` is not used, there are no dependencies, and the output is formatted with `{:016x}`. The reference vector `fnv1a("a") = 0xaf63dc4c8601ec8c` is correct and pinned in a test, along with a fixed digest line. I brute-forced with an independent implementation and it produced identical hashes.
- **Corpus missing:** with the symlink removed and `local/` deleted, the suite gives 156 passed and 1 ignored. `golden_snapshots` skips every style, no `local/` directory is created, and the corpus-dependent tests return early.

No fixes pushed. The branch is untouched.

Issue #50, opened 2026-09-23
live.rs: raw MIDI data byte indexes held[] on the input thread; .gitignore misses corpus/soundfonts symlinks


ryangavin:
- `live.rs` indexes `held[k]` with a raw CoreMIDI data byte, so a malformed or running-status byte of 0x80 or more could panic the MIDI thread. Mask it with `& 0x7F` or validate it. (Flagged in the review of #39.)
- The `corpus/` and `soundfonts/` patterns in `.gitignore` don't match symlinks, so worktree symlinks show as untracked and `git add .` would commit them. Use `/corpus` and `/soundfonts` without the trailing slash.

---
Milestone: **M2 · Chord-to-note conversion 1:1**

ryangavin:
Decision (in #59): validate in the stream splitter (`midi::for_each_message`) instead of only masking at each use. A status byte in a data position ends the message, as the MIDI spec says: the cut message is dropped and reading continues from the status byte. Realtime bytes inside a message are allowed. As a result, no consumer (keys, pads, synth) ever sees a data byte of 0x80 or more. `key_msg` keeps its `& 0x7F` masks as a second guard.

PR #51, opened 2026-09-23
Launchkey: pad pages for chord/setup and OTS/parts, Track buttons change style

ryangavin:
More of the arranger on the Launchkey MK4: the 16 pads now have three pages, and the Track buttons change style.

## Page switching: Pad Bank ▲ / ▼ (CC 106 / 107)

- **▼** goes to the next page and **▲** to the previous one. They stop at the ends, so a few presses of ▲ always take you back to page 1.
- The arrows light in the current page's colour, and only in a direction where there is another page.
- **Why these buttons:** Novation labels them "Pad Bank up/down" and uses them to move the pads through banks, which is what a page switch is. They sit right next to the pads, need one hand, and have no Shift function in the firmware.
- **Why not Shift + pads:** in DAW mode the firmware uses Shift + pad to change encoder and pad modes (user guide Table 3).
- **Tempo +/- didn't move.** It stays on Scene Launch > / Function (104/105).
- **The old toggles moved.** These two buttons used to toggle Left voice and OTS Link. Those are now **Shift + ▲ / ▼**, pads on page 3, and `l` / `F10` as before.
- **Terminal:** `Tab` / `Shift+Tab` cycle through the pages. The pad map shows `page n/3 <name>` and mirrors the current page.

## Pages

**Page 1 · Sections**: the layout is unchanged, with per-section colours.

| pad | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 |
|---|---|---|---|---|---|---|---|---|
| top | Intro I | Intro II | Intro III | Sync Start | Ending I | Ending II | Ending III | Auto Fill |
| bottom | Main A | Main B | Main C | Main D | Break | Tap | Sync Stop | Start/Stop |

**Page 2 · Chord/Setup**: all pads cyan.

| pad | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 |
|---|---|---|---|---|---|---|---|---|
| top | Single Finger | Fingered | Fingered On Bass | Multi Finger | AI Fingered | Full Keyboard | AI Full Keyboard | Upper/Lower (`d`) |
| bottom | Manual Bass (`D`) | Stop ACMP (`h`) | Split − (`[`) | Split + (`]`) | Kbd transpose − (`;`) | Kbd transpose + (`'`) | Transpose reset (`/`) | — |

- The lit fingering pad is the active type.
- Manual Bass is dark in Lower, where it isn't available.
- The transpose pads light while the transpose is down, up, or not zero.
- No tempo pads: tempo stays on its buttons.

**Page 3 · OTS/Parts**: all pads magenta.

| pad | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 |
|---|---|---|---|---|---|---|---|---|
| top | OTS 1 | OTS 2 | OTS 3 | OTS 4 | OTS Link (`F10`) | Left voice (`l`) | Left voice − (`(`) | Left voice + (`)`) |
| bottom | Rhythm 1 (`z`) | Rhythm 2 (`x`) | Bass (`c`) | Chord 1 (`v`) | Chord 2 (`b`) | Pad (`n`) | Phrase 1 (`m`) | Phrase 2 (`,`) |

- The lit OTS pad is the last one recalled.
- OTS pads the style lacks are dark. So are the OTS and Left pads when the synth is off.
- Part pads are lit while the part plays. Bass shows off while Manual Bass mutes it, the same as on screen.

## Buttons and CC numbers

Every number comes from the **Launchkey MK4 Programmer's Reference Guide v3.0**, page 9 ("The surface in DAW mode", Figure 3, decimal). Button names come from the Launchkey 49 MK4 User Guide v3.0, hardware overview (pp. 9–11). No MK3 fallback was needed.

| control | CC | source | use |
|---|---|---|---|
| < Track | 103 | Programmer's Guide p.9, Fig. 3 | previous style |
| Track > | 102 | Programmer's Guide p.9, Fig. 3 | next style |
| Shift | 63 | Programmer's Guide p.9, Fig. 3 (also the p.22 table: 3Fh Shift, "always enabled in DAW mode") | modifier |
| Pad Bank ▲ | 106 | Programmer's Guide p.9, Fig. 3 | previous page · Shift: Left voice |
| Pad Bank ▼ | 107 | Programmer's Guide p.9, Fig. 3 | next page · Shift: OTS Link |
| Scene Launch > | 104 | Programmer's Guide p.9, Fig. 3 | tempo + (unchanged) |
| Function | 105 | Programmer's Guide p.9, Fig. 3 | tempo − (unchanged) |
| Play / Stop | 115 / 116 | Programmer's Guide p.9, Fig. 3 | unchanged |

These match what the code already used: 104/105 are the buttons right of the pads, and 106/107 the buttons left of them.

## Style select

- < Track / Track > send `Action::Style(-1/+1)`, the same action `←/→` now produce, and run the existing style-load path.
- That path prepares the style on the UI thread and hands it to the engine through `style_tx`. It works the same whether the band is playing or stopped.
- Both Track buttons light while there is more than one style.

## How it's wired

- `launchkey::Action` covers every control the Launchkey reaches.
- The keyboard shortcuts now map to the same `Action`s (`ui::key_action`). One executor in the UI loop runs them, whether they come from a key or from the Launchkey.
- **Engine buttons** go straight from the CoreMIDI thread to the engine ring, as before: page 1, Stop ACMP and the part toggles.
- **Everything else** goes over a new 64-slot rtrb ring to the UI thread. That covers settings, OTS, Left voice and style.
- **Real-time rules hold:** the input thread only does atomic loads/stores and try-pushes. There is no allocation, locking or panicking.
- **Page state** is an `AtomicU8` in `Shared`, read by the input thread (pad → action) and the UI (LEDs, screen).
- **LEDs:**
  - Pages 2 and 3 use the existing RGB `Look` model, and `--palette-leds` gets palette equivalents (cyan 37/39, pink 57/59).
  - Every page lists the pads in the same order, so the existing per-index LED cache resends only the pads that changed.
  - Nav button LEDs follow the old side-button pattern: palette on ch 1 plus brightness on ch 4.
- **Diagnostics:** the last DAW-port CC or note that nothing is mapped to is recorded (`Shared::last_unmapped`). It shows on the status line as e.g. `unmapped CC 103 = 127` or `unmapped note 36 = 90 (ch 10)`. Blank pads on a page don't count, since they are known pads.

## Also

- `.gitignore`: `corpus/` and `soundfonts/` become `/corpus` and `/soundfonts`, so symlinks match too (#50, second bullet). Added `/.claude/worktrees/`.
- README Controls is rewritten. It now has the three page tables, how to switch pages, a button table, and the full list of keyboard shortcuts, which was missing `f d D h ; ' : " / [ ] l ( ) F1-F10 tab`.

## Tests

- `cargo test --release`: 175 passed, 1 ignored.
- Clippy: no new warnings. `live.rs` went from 5 to 3 and `ui.rs` stayed at 11, all pre-existing.

New tests:
- `launchkey`: note → action on each page (page 1 matches the old `pad_button` for all 128 notes), CC → control with and without Shift, page stepping and wrapping, LED state per page in RGB and palette, page 1 LEDs independent of the panel state, nav button LEDs.
- `live`: through `Input`: ▲/▼ page changes, pads following the page, engine vs UI routing, Shift + ▲, Track buttons, unmapped recording.
- `ui`: every page 2/3 action and Track action is also a keyboard shortcut's action; the unmapped readout format.

## To verify on hardware

1. Pad Bank ▲/▼ switch pages and light in the page colour (white / cyan / magenta).
2. Shift + ▲/▼ toggle Left voice / OTS Link, and the Launchkey firmware doesn't swallow them.
3. < Track / Track > change style, and their LEDs light.
4. Pad Bank arrows lit in colour: they might be single-colour LEDs, in which case the ch 4 brightness applies and all pages look the same.
5. If anything does nothing, read the `unmapped CC …` line at the bottom of the screen.


ryangavin:
## Adversarial review of #51 (head d7f80a7)

**Verdict: approve. No blocking issues.** All the findings below are nonblocking.

### What I ran
- `cargo build --release` passes. `cargo test --release`: 175 passed, 1 ignored.
- Clippy compared with `integration/m1-chord-recognition`, per file and lint: no new warnings. `live.rs` has two fewer collapsible-if warnings.
- I added three throwaway tests to `live.rs` that drive `Input::pad_msg`, ran them, then reverted them (they are not in the PR):
  - `held_pad_release_after_page_switch_is_inert` passes. It presses Start/Stop on page 1, then ▼, then sends the pad's release as `90 77 00`, as `80 77 00` and as aftertouch `A0 77 28`. Nothing reaches the engine or the UI ring.
  - `channel_7_feature_reply_moves_page` passes, which confirms N1 below.
  - `stuck_shift_hijacks_page_buttons` passes, which confirms N2 below.

### Checks that passed
- **Real-time safety on the input thread.** The new path does only atomic loads and stores, `rtrb::push`, and index or subtract operations whose bounds are guarded by the match arms. `Page::from_u8` clamps, `Fingering::ALL[note-96]` is only reached for notes 96–102 (7 entries), and `TogglePart(note-112)` for notes 112–119. There is no allocation, lock or panic.
- **Stuck notes or state on a page switch.** Pads act on note-on only, and every release form is ignored, so a release that arrives after a page change can't trigger anything. Pads never produce output notes. A style change from < Track / Track > runs the same `style_tx` path as `←/→`, so it adds no new stuck-note risk.
- **Page 1 is byte-for-byte unchanged.** `section_leds` and `section_looks` are pure renames. `pad_action(Sections, n)` equals `pad_button(n)` for all 128 notes. The CC paths for 104, 105, 115 and 116 behave as before, including matching on any channel. The only difference is intended: 106 and 107 now change page.
- **CC numbers.** I rendered Figure 3 (p.9) of the Programmer's Reference Guide. It shows Shift 63, < Track **103**, Track > **102**, 106/107 left of the pads, and 104/105 right of them. All match the constants.
- **Page 2/3 actions match their keys.** Each pad runs the same `Action` as its key, and the UI executor runs the old key bodies unchanged. I checked `[` `]` (clamped 24..=96 as before), `;` `'` `/`, `d`/`D` with `sync_manual_bass`, `!@#$`, `F10`, `l`, `(` `)`, `h` (`Cmd::Button(StopAcmp)` in both paths) and `z…,`.
- **README.** It matches the code: the three page tables, the button table, Shift + ▲/▼, and the new key list.

### Nonblocking findings
1. **N1 · Button CCs match on any channel, so a channel-7 feature-control reply could change the page or run an action.** According to the guide (p.21), the Launchkey sends feature-control replies on channel 7 (`B6h`). That table reuses 66h–6Bh (102–107) for MIDI channel, velocity and arp settings. `pad_msg` masks off the channel, so `B6 6B xx` pages down (proven by the test), and `B6 66/67` would change style. In DAW mode those (*) features supposedly don't reply, so the risk is low. The same weakness already existed for 104 and 105. Suggestion: skip `st == 0xB6` before `cc_control`, or match only the channel the buttons report on. Shift 3Fh is (#), always reported on ch 7, and is harmless because it only sets `shift`.
2. **N2 · Shift can stick.** `shift` only changes on CC 63. If the release is lost, for example because Shift was held across DAW-mode entry or the firmware's Shift menu swallowed it, ▲/▼ keep toggling OTS Link or Left voice instead of changing page until Shift is pressed again (proven by the test). Pressing Shift again recovers, but it is surprising the first time.
3. **N3 · A page change can be lost.** Both `▲/▼` on the input thread and `Tab` on the UI thread do a `load` then a `store` on `shared.page`, so if they coincide one change is lost. `fetch_update` would close this. It is cosmetic.
4. **N4 · The lit OTS pad goes stale after a style change.** `ots_applied` is not reset when the style changes. With OTS Link off, page 3 keeps "OTS n" lit for a style where it was never recalled. The status-line text already had this behaviour, but a lit pad makes it more visible.
5. **N5 · Track with only one style reloads that style.** The Track LEDs are dark, yet a press still runs `load()` and pushes the same style to the engine, mid-play if the band is playing. `←/→` already did this. Suggestion: return early when `styles.len() <= 1`.
6. **N6 · Launchkey actions can wait a frame.** A key event that isn't a press hits `continue` before the `act_rx` drain, and a style `load()` blocks the drain. Moving Left voice and OTS Link from the input thread to the UI thread adds up to one frame of latency. That's fine, but worth knowing.
7. **N7 · Nav button LEDs aren't cleared on exit.** Exit clears only the 16 pads and relies on `EXIT_DAW` to reset the button LEDs. This was already true for the side LEDs.
8. **N8 · Nit.** The on-screen hint for the page-2 fingering pads is `f`, but `f` steps through the types rather than selecting that one.

### Still to verify on hardware
These are the same items as in the PR description. Most important: check that Shift + Pad Bank ▲/▼ actually reach the port. If they don't, the only Launchkey path to Left voice and OTS Link is page 3.

ryangavin:
Review follow-up, in 3aea2f5.

**Addressed**
- **N1, channel 7:**
  - Every CC with status `B6` (mode reports and feature-control replies) is now ignored before the button and fader checks. That covers the fader range too, since the Arp CCs 2–7 on ch 7 would otherwise have set part volumes. Nothing on ch 7 is reported as unmapped.
  - Shift (63) is still read on any channel, since the guide says it is reported on ch 7 in DAW mode.
  - Test: `channel_7_replies_are_not_button_presses`.
- **N2, sticky Shift:**
  - Shift now clears on two signals.
    - **Any pad note-on:** the firmware keeps Shift + pad for itself, so a pad note means Shift is up.
    - **Any pad-mode report (`B6 1D xx`):** the firmware sends one after its Shift menu is used or when DAW pad mode comes back.
  - I chose this over a timeout because a timeout would also break a Shift that is really being held.
  - I didn't add "clear on page switch": with Shift stuck, ▲/▼ never switch the page, and Tab runs on the UI thread, which doesn't own the Shift state.
  - Test: `shift_does_not_stick`.
- **N3, page race:** Pad Bank and Tab both go through `Shared::step_page`, a `fetch_update` compare-and-swap. It is lock-free and doesn't allocate. Test: `page_steps_from_either_side`.
- **N4, stale OTS pad:** a style change sets `ots_applied` to 0. OTS Link, if on, recalls one on the next pass as before.
- **N5, single style:** `Action::Style` does nothing when only one style is loaded. This applies to both the Track buttons and ←/→.
- **N7, exit:** exit now turns off the Pad Bank, Track and fader button LEDs (palette on ch 1 and brightness on ch 4) before `EXIT_DAW`. The nav LED test now covers this too.
- **N8, hint:** the page-2 fingering pads now show `[pad]`. No key selects a single fingering type.

**Left as is**
- **N6, latency:** Left voice and OTS Link now run on the UI thread, so they can wait up to one frame (~16 ms), or longer while a style `load()` is running. I haven't changed this. They are panel toggles, not timing-critical controls. Engine buttons still go straight from the MIDI thread.

**Gates**
- `cargo test --release`: 178 passed, 1 ignored.
- Clippy: no new warnings, with the same per-file counts as before (`live.rs` 3, `ui.rs` 11).

Issue #20, opened 2026-09-23
Browser UI: categories, search, favorites, recents


ryangavin:
- Genos-style category pages plus instant search. Use keyboard/pad navigation for live use.

---
Milestone: **M4 · Style browser** · Spec: [docs/genos-features.md](../blob/main/docs/genos-features.md) · Backlog: [docs/backlog.md](../blob/main/docs/backlog.md)

Issue #19, opened 2026-09-23
Style library index


ryangavin:
- Scan folders and cache each style's name, category, tempo, time signature, available sections, OTS names, source file and SFF version. Rescan incrementally.

---
Milestone: **M4 · Style browser** · Spec: [docs/genos-features.md](../blob/main/docs/genos-features.md) · Backlog: [docs/backlog.md](../blob/main/docs/backlog.md)

PR #52, opened 2026-09-23
TUI style browser: recursive library index, filter, load while playing

ryangavin:
Early, simple version of #19 / #20: a style browser inside the ratatui TUI, as a stopgap before the Tauri app.

## What it does
- **Library scan** (`src/library.rs`): `yahaha play <folder>` already walked folders recursively. The walk moved here and now also skips hidden/AppleDouble files (`._Foo.sty`) and guards against symlink loops. It matches every style extension (`sty prs sst bcs pcs pst fps`, any case).
- **Index**: a background thread (`yahaha-index`, normal priority) reads each file's display name (SFF name marker, falling back to the file stem), tempo, time signature and sections. It uses `sff::Summary::load`, which reads only the MThd header and the style track, so no CASM/OTS parsing and no section building. Results reach the UI thread through an mpsc channel that's drained between frames, and the list fills in progressively. A file that fails becomes a red error row. `catch_unwind` guards the indexer as well.
  - Timing: 3,600 styles (81 MB) scanned and fully indexed in about 0.5 s wall time, warm cache. Startup itself only does the folder walk plus the first style load.
- **Browser overlay**: opens with `Enter`. Close it with `Esc`, which only quits when the browser is closed.
  - Why `Enter`: it was unbound, and "Enter to browse" is easy to remember. `b` is taken (part mute, Chord 2 in `zxcvbnm,`).
  - The list shows name, folder (category), tempo and time signature, plus the sections. The loaded style is marked `▶` and the cursor starts on it.
  - Type to filter (case-insensitive substring on name or folder), `Backspace` edits, `↑/↓` `PgUp/PgDn` `Home/End` move, and `Enter` loads and closes.
  - While the browser is open, every key goes to it, never to the performance shortcuts. `Ctrl+C` still quits.
- **One load path**: `switch_style` is shared by ←/→, Track ◀/▶ and the browser. It's the same `style_tx` ring as before, so it works while the band plays and the band follows the next chord.
- **Folder order**: ←/→ and Track ◀/▶ step through the browser's unfiltered order (folder, then name). They skip styles known not to load, so a broken file can no longer trap you.
- **Real-time safety**: the engine, MIDI and synth threads are untouched. Indexing runs on its own thread, and loading stays on the UI thread as before.
- `sff.rs`: the meta-event scan (name, tempo, time sig, markers) is now shared by `Style` and `Summary`. Names end at the first NUL, since many styles NUL-pad the name (for example `AustinCityBlues.S930.STY\0\0…`).
- `yahaha screen <folder> out.html` renders the browser open over the panel, for checking the layout without a terminal.

## Not done
- **Launchkey browsing (skipped)**. Play is start/stop, and there's no free "enter" button to load with. Moving a cursor from the hardware with nothing to load it would be half a feature. Track ◀/▶ keep stepping styles while the browser is open, and the `▶` marker follows.

## Tests
- `library.rs`: recursion and every extension in mixed case (non-styles and `._` files ignored), the index fields, the folder-then-SFF-name sort and the matching `step` wrap, error rows (garbage, empty, no sections, truncated, missing file) and `step` skipping them, several roots, and the filter (case, name vs folder, backspace widening, re-match once the index names arrive). The corpus check confirms the light index agrees with the full parser on every file.
- `ui.rs`: browser key handling (typed letters go to the filter, move/End/Enter/Esc, empty result).
- `cargo build --release` and `cargo test --release` pass (186 tests). Clippy has no new warnings (31, down from 32).


ryangavin:
## Adversarial review

I checked out `tui-browser` at 7e8106b and linked the corpus. `cargo build --release` and `cargo test --release` pass (186). I also wrote throwaway tests (not committed) for fuzzing, filesystem edge cases, FIFOs, scale and order stability.

### Blocking

**1. With one Esc too many, the app quits mid-performance.** Esc is now the routine "close browser" key, and `Enter` also closes the browser when it loads. A player who presses Enter to load and then Esc out of habit, or who double-taps Esc to close the browser, gets `return Ok(())`: the app quits and the band stops. Before this PR Esc also quit, but nothing trained you to press it. Now it's the most-pressed key in the browser flow, so quitting by accident is clearly easier than before. **Fixed in 7f3ea3b** (see the next comment).

### Nonblocking

2. **A filter match can vanish while indexing and move the cursor.** The filter matches `name_lc`, which switches from the file stem to the SFF name when the index arrives. If the cursor row matched only by its stem (for example, you typed part of a file name), it drops out of the list. The highlight then falls back to `v[0]`, and Enter loads that row. Enter always loads what the last frame highlighted: `apply` runs before `draw` and never between draw and key, so this is not a data race. It can still surprise you in the first ~0.5 s, or for longer on a cold USB stick. It also means a file name stops matching once the index lands. Suggestion: match the stem *and* the SFF name. The stem never changes, so this case goes away.
3. **←/→ order shifts while indexing.** `step` walks `order`, which re-sorts as SFF names arrive. My test copied three corpus styles as `aa/mm/zz.sty`: before the index the order is `[aa, mm, zz]`, and after it's `[AustinCityBlues…, BluesOrganTrio…, NewSong3]`. A →-press during indexing can therefore revisit or skip a style. The loaded entry itself is tracked by id, so the wrong style never loads. The order is stable once indexing is done. Fine as a known limitation.
4. **Hidden *folders* are indexed.** `is_style` skips `.foo` files, but the walk still descends into `.Trashes/501/`, `.git`, and so on. On a USB stick, deleted styles in `.Trashes` show up in the browser and in ←/→. Test: `.Trashes/501/deleted.sty` was listed with folder `.Trashes/501`. The fix is one line: skip dirs whose name starts with `.` in `scan`.
5. **A FIFO named `*.sty` blocks the index thread forever.** `File::open` blocks until a writer appears. My test: with `mkfifo a_pipe.sty` + `b.sty`, no index result arrived in 2 s. The UI stays responsive, but every later row stays `…` and the title says "indexing, N to go" forever. If the FIFO sorts first, the startup `load()` on the UI thread also hangs, which was true before this PR too. Suggestion: skip anything that isn't `is_file()` in `scan`.
6. **Unreadable folders are skipped silently.** `read_dir` errors are flattened away. That's fine, but a missing category gives the user no hint.
7. **Summary and full parse can disagree on OK-ness.** `Summary` doesn't read CASM, so a style with a broken CASM shows as a healthy row, and loading it then fails. `switch_style` turns it red, so the fallout is harmless. The reverse can happen too: an index `Err` for a style that loads, when the MThd length is over 64 KiB (the clamp in `Summary::load`). That's theoretical only.
8. **Ctrl+letter is typed into the filter.** For example, Ctrl+U adds `u`. Cosmetic.
9. **The Esc fix may overwrite an error message.** The "press esc again" prompt replaces the current error message (such as a synth error) for its 1.5 s window. Minor.

### Verified OK
- **Cursor and threads:** `Browser.cursor` is an entry id, so re-sorting never changes what Enter loads. `apply` runs only on the UI thread, before `draw`. The indexer exits when the receiver drops.
- **Key leakage:** with the browser open, the performance match gets `KeyCode::Null`, so `key_act` stays `None` and no shortcut fires. Ctrl+C still quits. Launchkey and MIDI paths are unchanged.
- **Malformed files:** I fuzzed 4,935 cases (every ~1/60 truncation, random byte mutations in the first 4 KB, and random tracks behind a valid MThd/MTrk) through both `summarize` and `parse`. No panics, and each case finished well under 2 s. Where both parsers succeeded they agreed on name, tempo and time signature every time. A claimed `tlen = 0xFFFFFFFF` or `hlen = 0xFFFFFFFF` returns an error quickly, with no big allocation.
- **Header vs full parser:** on the whole corpus (35 files), name, bpm and time signature match `Style::load` exactly. The existing corpus test doesn't check bpm; it might be worth adding.
- **Filesystem:** a dir symlink to the root and a dir symlink to a parent are each visited once, and a file symlink loop becomes an error row. APFS refuses non-UTF-8 names, and `to_string_lossy` covers other filesystems.
- **Scale:** 3,500 files (100 folders × 100 styles) scanned in 24 ms and indexed in 257 ms. The worst UI-frame work (`apply` + re-sort + `filter` + `position` + `pending`) was 0.38 ms.
- **Real-time:** the engine, MIDI and synth threads are untouched. Style loads use the same `style_tx` ring as before.
- **README:** accurate. I updated the Esc lines for the fix.

**Verdict: approve once the Esc fix (7f3ea3b) is in.** Items 2, 4 and 5 are cheap follow-ups worth doing.

ryangavin:
## Fix pushed: 7f3ea3b "Esc quits only when pressed twice within 1.5 s"

- Adds `QuitGuard` in `src/ui.rs`. With the browser closed, one Esc shows "press esc again to quit" and a second Esc within 1.5 s quits. Any other key, or an Esc that goes to the browser, disarms it. Ctrl+C still quits at once. The prompt clears itself when the window runs out.
- Added the unit test `esc_quits_only_when_pressed_twice_quickly`.
- Updated the help line and README (`esc` twice quits, or `ctrl+c`; one `esc` closes the browser).
- `cargo build --release` and `cargo test --release` pass (187). Clippy is unchanged at 31 warnings, none of them new.

ryangavin:
Addressed the nonblocking items in 9826933, on top of 7f3ea3b. Each fix has a test.

- **#2 (a filter match vanishes while indexing): fixed.** `Entry::matches` now checks the file name as well as the style name and the folder. The file name never changes, so a row that matched by it stays listed after the index lands. The test is in `filter_is_case_insensitive_on_name_and_folder`: after the SFF name arrives, the file name still matches.
- **#4 (hidden folders are indexed): fixed.** The scan no longer descends into dot-folders such as `.Trashes` or `.git`. Test: `hidden_folders_and_non_regular_files_are_skipped`.
- **#5 (a FIFO blocks the index thread): fixed.** The scan lists only regular files. That covers folders named `x.sty` and files named on the command line, and a missing named file still becomes an error row. `index_one` also refuses anything that isn't a regular file, without opening it. Because FIFOs never make the list, the startup `load()` can't hang on one either. Test: the same test as #4, which uses `mkfifo`.
- **#8 (Ctrl+letter types into the filter): fixed.** `Browser::key` now takes the modifiers and ignores Ctrl/Alt + a character. Shift still types a capital. Test: `browser_ignores_ctrl_and_alt_letters`.
- **#3 (←/→ order shifts while indexing): declined.** `step` already goes from the loaded style's id, not a stored position, so the wrong style never loads. What moves is its neighbours, because the order uses SFF names as they arrive. Freezing the order would make ←/→ disagree with the browser, and a second provisional order would change anyway once the index finished. The window is about 0.5 s on a warm cache, so I left it as a known limitation.
- #6, #7 and #9 weren't in scope for this pass and are unchanged.

Gates: `cargo build --release` passes, `cargo test --release` passes (189), and clippy has no new warnings (31, the same as after 7f3ea3b).

PR #53, opened 2026-09-23
Mixer: style part levels drive the faders; unity synth gain

ryangavin:
## Principle

**A part fader is that channel's CC7.** CC7 is the universal volume message. Moving a fader sends its value unchanged, as CC7 on that part's channel, to every destination: the `yahaha` MIDI port and the built-in SoundFont synth. There is no separate internal gain, multiplier or scaling stage in the engine or the synth, so an external VST/AU driven from these channels will come out at the same loudness from CC7 alone. Style CC7 events set the fader value, which is then sent as CC7. They are never multiplied by the fader. The synth answers CC7, CC11 and velocity on the standard GM curves only. The master fader is the only gain that is not a MIDI message: it sets the synth's output level. After it there is only a safety soft clipper, which leaves everything below -1 dBFS unchanged. Voice slot, OTS and Left volumes follow the same rule: each is sent as CC7 on its own synth channel, and also on the port (see below).

## What was wrong (verified)

- `Engine::volume` sent `style_CC7 × fader / 127`, and the faders started at 127. So the faders scaled the style's levels instead of being those levels, and the screen showed 100% for every part.
- a4dab2d then put a square-root curve on CC7/CC11 and a compressed velocity curve with a floor of 48 into the built-in synth, to make quiet parts audible. That was a workaround for this imbalance, and it also broke the author's balance, because it flattened every level difference the style sets.

## Changes

- **Initial fader positions.** `Prepared::mix` holds, for each of the 8 accompaniment parts, the last init (SInt) CC7 routed to that part, or the GM default of 100. Loading a style, or starting the engine, sets the faders to these values. `send_init` sends the mixer's value for every part instead of the style's init CC7, so every part gets exactly one CC7, including the parts that use the default.
- **Section changes and restarts.** Sections never re-apply SInt. A start (Start/Stop) re-sends the init. Faders the player moved keep their value, and the mixer wins over SInt for them. Only untouched faders go back to the style's SInt level (89b4f59), so an Intro or Ending pattern's CC7 doesn't stick.
- **Style change.** Resets all 8 faders to the new style's levels and clears the record of which faders the player moved. This matches a Genos with no parameter lock.
- **Launchkey soft takeover.** One rule, `engine::Takeover`, used by the part faders (in the engine) and the master fader (in the input thread). Takeover is re-armed whenever software moves a value: a style load, a pattern CC7, or a start resetting an untouched part. After that, the hardware fader is ignored until it comes within 2 of the value or crosses it; then it follows directly. A fader that has never reported must also pick up first. The TUI shows `↕` next to a waiting part and next to the master on the synth line. A fader that has never reported is not marked.
- **Your parts' CC7 on the port.** The port carries your playing on two channels, so it gets two CC7s. Ch 1 gets the lowest layered slot's volume (Right 1 when it is on; one channel can't carry per-slot levels for a layer). Ch 2 gets the Left volume, or the Style's Bass fader under Manual Bass. Both are sent when they change. They go to the port only, because the built-in synth already has per-slot CC7, and a CC7 on ch 1 there would set every slot to the same level.
- **Manual Bass** now plays the Left part at the Style's Bass part fader (its CC7), not at the Left volume.
- **Display.** The TUI parts panel shows the real 0-127 volume next to the bar, and the synth line shows the raw master value.
- **Unity synth.** `soften_level` and `soften_velocity` are gone, so CC7, CC11 and velocity reach rustysynth unchanged: gain = (vel/127)² · ((CC7/127)·(CC11/127))², which is the GM curve. The fixed 0.6 / 0.8 master trims are replaced by `master_gain(m) = m / 100`, so the default master of 100 is unity.
- **Headroom.** The ignored `loudness_probe` now also renders the whole band at master unity, with reverb and chorus. Over its 4 bars, the 35 corpus styles peak between -3.2 dBFS (HoldOnAMinute) and -20.8 dBFS. Review measured about -1.7 dBFS over longer playing. Master 127 adds 2.1 dB, and the right hand adds more, so the output can exceed full scale. A safety soft clipper (`synth::soft_clip`) now runs on the final output only. It is identity up to -1 dBFS; above that it bends smoothly (tanh, slope 1 at the knee) towards full scale and never exceeds it. It costs one comparison per sample, with no allocation.
- OTS, Left and voice-slot volumes were already sent as plain CC7 on their player-synth channels, with no softening. They now also go out on the port (above).
- **A lost fader move (review #4).** The part faders' takeover state lives in the engine and only changes when a move reaches it. Crossing is judged from the last report received, so a move dropped because the command ring was full can't leave takeover inconsistent: the next move still counts as a crossing.

## Decisions

**Corpus survey (35 styles):**
- **In-pattern CC7:** found in 5 styles, 107 events in total.
  - Every one of these events is at tick 0 of a section: a per-section level setup, not a volume ride.
  - In JumpinJiveOrgan and SmoothLeadJam they only repeat the init value.
  - In SmoothItOver and NightCruiser they are the only level some CASM-routed parts get, and they differ by section (e.g. 47 vs 61, 79 vs 94).
  - In TickingAway they set 76 against an init of 90, in Intro B and Ending C.
- **In-pattern CC11:** found in 32 styles, 4,572 events: real expression moves.
- **Parts without a CC7:** 126 of the 280 part slots have no init CC7, so they use the GM default of 100.

**Decision: in-pattern CC7 updates the part's fader (it is sent, and the fader shows it) until the player moves that fader. From then on, the player's value wins until the next style load.** Because: a fader is the channel's CC7, so the displayed fader must always equal the last CC7 the channel received. Ignoring pattern CC7 would lose the only level that some parts in SmoothItOver and NightCruiser get. Always obeying it would let a section change overwrite the player's mix. With this rule the author's per-section levels play as written, and the player's own moves stick.

**Decision: CC11 passes through untouched.** Because: CC11 is the pattern's expression inside the level the mixer sets. That is what the Genos does, and the GM curve multiplies CC11 with CC7.

**Decision: master fader 100 = unity output gain (linear, 127 gives about +2 dB).** Because: 100 is the existing default, and the master fader is the only gain allowed outside CC messages. Unity leaves no guaranteed headroom (review measured about -1.7 dBFS), so the safety soft clipper covers the rest without changing any level below -1 dBFS.

**Decision: part-fader takeover in the engine; master takeover in the input thread; one shared rule.** Because: each value's owner holds its takeover state. The engine owns the part levels, and style loads, pattern CC7 and starts change them there. The master is a synth atomic that only the Launchkey moves. It is a few bytes of state with no locks or allocation. The snapshot carries the part pickup mask to the UI, and an atomic carries the master's.

**Decision: on the port, ch 1 gets the lowest layered slot's volume.** Because: the right hand is one port channel, so a layer can't carry per-slot levels, and Right 1 is the lead voice. An external instrument on ch 1 plays the layer as one part.

## Tests

- `sim::mixer::style_load_sets_faders_from_style_cc7`: faders equal the style's init CC7, 100 where there is none, and exactly one CC7 per part.
- `sim::mixer::fader_sends_its_value_as_cc7`: 0/1/64/100/127 go out unchanged, and a restart sends the mixer's value.
- `sim::mixer::section_change_keeps_user_fader`: SmoothItOver. Pattern CC7 moves an untouched fader. After the player sets it, 8 section changes send no CC7 on that part, and a restart sends the player's value.
- `sim::mixer::style_change_resets_faders`
- `sim::mixer::hardware_fader_soft_takeover`: far away ignored and marked, crossing picks up, within 2 picks up, and a style load re-arms takeover.
- `synth::curve_tests`: renders through `apply` and pins CC7 64/100/32, CC11 64, velocity 64 and a combined case to 40·log10(v/127) dB (±0.2 dB). Master default = unity.
- `sim::mixer::takeover_rule`: the shared rule, including a jump across the value (lost moves) and re-arm by a software move. `sim::mixer::start_keeps_moved_fader_under_hardware_control`. `sim::mixer::restart_restores_untouched_style_levels` (review fix).
- `synth::curve_tests::soft_clip_is_transparent_below_minus_1_dbfs`: identity up to the knee, monotonic, odd, below full scale, slope 1 at the knee.
- `synth::curve_tests::port_volumes_follow_slots_left_and_manual_bass`
- `cargo build --release` and `cargo test --release` pass: 200 passed, the golden digests are unchanged, and the ignored loudness probe is kept. No new clippy warnings in the lines this PR touches.

## What to listen for

- Parts should now sit where the style author balanced them. Parts the old curves had lifted will be quieter. For example, CoolRevibed's CC7 30-66 parts and velocity-1 parts now play at their real level. T552 styles (CoolRevibed, StreetGenos, DownTime) are noticeably quieter overall than S930 ones. That is their authored level on GM voices, not a bug.
- Some voices may sound too quiet or too loud because GeneralUser's samples differ from the Yamaha voices the styles were written for. With this PR the fix for that is the fader, and the fader shows the actual CC7.


ryangavin:
## Adversarial review

Built and tested at 654fc08: `cargo build --release` and `cargo test --release` pass (184 passed, 1 ignored). The golden digests are unchanged.

### What holds up
- **No hidden per-part gain remains.** I grepped engine, synth, live and ui for multipliers. The old `style × fader / 127` is gone. Pattern CC7, `set_volume` and `send_init` all send the raw fader value. The band synth's `apply` passes CC7, CC11 and velocity through unchanged. Stop Accompaniment uses fixed velocities (90/70) on the real Bass/Pad channels, so its level comes from their CC7. OTS, Left and voice-slot volumes are plain CC7 on the player synth.
- **The CC7 sent always equals the fader shown.** I wrote a throwaway sim test for all 35 corpus styles. After every step it compares the last CC7 sent on each part with `snapshot().volumes`. The steps were Main A-D, stop, start, Intro B, Ending C, restart, hardware fader moves (0 → 127 → 33) and a style change, then 4 more bars. There were no mismatches.
- **The GM curves are real.** In rustysynth 1.3.6, `voice.rs` computes velocity gain as `2·linear_to_decibels(vel/127)` (40·log10) and channel gain as `(vol·expr)²`, so CC7 and CC11 are each 40·log10(v/127). `curve_tests` checks this at ±0.2 dB. The 14-bit volume (127<<7)/16383 costs only -0.07 dB.
- **Real-time safety is fine.** `set_mixer`, `pattern_volume`, `hw_fader` and `pickup_waiting` don't allocate, lock or index outside bounds (`part & 7`, `min(127)`). The master fader is still a single relaxed `AtomicU8`.
- **Pattern CC7 is latched per part.** `user_set` has one bit per part. It is set on every user move (software or a hardware fader after pickup) and cleared only on a style load. A user move can't be overwritten by a pattern.
- **Soft takeover edge cases work.** A fader already within 2 of the target picks up at once. A jump past the target picks up by crossing, provided the hardware position was known before the jump. After startup, before any hardware report, nothing is marked.

### Findings

**1. BLOCKING (fixed in 89b4f59): a restart kept a level an Intro/Ending pattern set, on parts the player never touched.**
`start()` sent the mixer's current value for every part. On an untouched part, that is the last pattern CC7. TickingAway sets part 2 (ch 10) to 76 in Intro B and Ending C, against a SInt of 90, and no other section sets it. After Intro B → Main A → Ending C, a restart with Intro A sent 76, not 90, and the part stayed at 76 through every later run until the next style load. Before this PR, a start re-sent the SInt level. A Genos also re-plays SInt on a start.
Fix: at a start, every part not in `user_set` goes back to `style.mix` (through `set_mixer`, so soft takeover re-arms) before `send_init`. Faders the player moved keep their value, as before.
Test: `sim::mixer::restart_restores_untouched_style_levels`. It fails on 654fc08 with `left: Some(76) right: Some(90)`. It also checks that a part the player set (33) survives the restart.

**2. Nonblocking: I could not reproduce the headroom claim.** I ran a throwaway render of the full band: 35 styles, 24 bars each, Main A-D and 4 chords, default settings with reverb and chorus on, master at `master_gain(100)`. The worst peak was **-1.7 dBFS** (HoldOnAMinute), then -2.2 (SunsetDrive) and -3.1 (ClubJazz1). The PR says -3.2 dBFS at worst and "at least 3 dB of headroom". Also:
- The committed `loudness_probe` never calls `set_master_volume`. It runs at rustysynth's default of 0.5 (-6 dB) and measures RMS per part, not peak. So it can't back the claim.
- Master 127 (+2.1 dB) puts HoldOnAMinute over 0 dBFS.
- The player synth is summed on top with no limiter or clamp in the callback, so a right-hand chord adds to that.
- There are 35 `.sty` files in `corpus/MOX_v2`, not 36.

Suggested fixes: a soft clipper or limiter on the final sum (it is output gain, so allowed), or unity somewhat below 1.0.

**3. Nonblocking: the master fader jumps on its first touch.** Software starts at 100, and the hardware master has no takeover. The first touch jumps the master to wherever the physical fader is: 0 is silence, 127 is +2 dB. The PR says software never moves the master, but its startup value is itself a software value. Consider the same pickup logic, or at least note it.

**4. Nonblocking: a lost fader command can't be recovered.** `Cmd::PartVolume` is dropped silently if the command ring is full (`push(...).is_ok()`). This was already true, but a dropped message now also leaves `user_set` and `hw` stale. It is unlikely with a 1-in, 1-out ring. Mentioned for completeness.

**5. Nonblocking: the README and PR text are slightly off.**
- The README says soft takeover applies "after a style load moves a level". Pattern CC7 on an untouched part, and now a restart, also move the level and re-arm pickup. Something like "whenever the software moves a level (style load, the style's own CC7)" would be accurate.
- The PR says "Start/Stop re-sends the init with the mixer's CC7 values". After the fix, that is true only for parts the player moved.

**6. Nonblocking and out of scope: OTS, Left and slot volumes don't follow the principle for external VSTs.** They are CC7 on the internal player synth only. The `yahaha` port gets raw ch 1/2 notes and no CC7 for them, so an external VST on those channels won't follow OTS volumes. The same goes for Manual Bass: the left hand plays the Bass voice at `left_vol`, not at the Bass fader's CC7. Worth a backlog item.

### Gates after the fix
- `cargo test --release`: 185 passed, 1 ignored. The golden digests are unchanged.
- Clippy: no warnings in the lines I touched. All remaining warnings were already there.

**Verdict:** approve once 89b4f59 is in. Findings 2-6 are follow-ups. I would still fix the headroom/limiter issue (2) soon, because -1.7 dBFS on the band alone, with master and the right hand on top, will clip.

ryangavin:
Fix pushed to style-mixer as 89b4f59 (finding 1).

`Engine::start` now resets every part the player has not moved (not in `user_set`) to the style's SInt level (`style.mix`, through `set_mixer`, so soft takeover re-arms) before `send_init`. Faders the player moved keep their value.

New test: `sim::mixer::restart_restores_untouched_style_levels` (TickingAway, Intro B → Main A → Ending C → restart). It fails without the fix (76 instead of 90).

Gates: build passes; tests: 185 passed, 1 ignored; golden digests unchanged; no new clippy warnings.

ryangavin:
Review follow-up, now in 2ac7bcc. The branch is rebased onto `integration/m1-chord-recognition`, which includes #52, the style browser. The rebase had no conflicts, and the browser and the mixer changes to `ui.rs` both remain. Your fix 89b4f59 is kept.

**Addressed**
- **#1 (89b4f59):** kept as is. Its test passes.
- **#2 Headroom:** added `synth::soft_clip` on the final output only, with no per-part gain. It leaves everything up to -1 dBFS unchanged. Above that it bends smoothly (tanh, slope 1 at the knee) towards full scale and never exceeds it. It costs one comparison per sample, with no allocation.
  - The ignored `loudness_probe` now also renders the whole band at master unity, with reverb and chorus, and reports its peak. The worst of the 35 styles is -3.2 dBFS over its 4 bars; your longer run measured about -1.7 dBFS.
  - The PR body's headroom claim and style count are corrected: 35 styles, not 36, and 280 part slots.
- **#3 Master fader:** now has soft takeover, using the same rule as the part faders. That rule is now a single `engine::Takeover` type. The synth line shows `↕` next to the master while it waits.
- **#4 Lost moves:** no change needed. The part faders' takeover state only changes when a move reaches the engine, and crossing is judged from the last move it received. A move dropped because the ring was full therefore leaves the state consistent, and the next move still counts as a crossing. A test covers a jump across the value.
- **#5 Text:** the README and the PR body now say that takeover is re-armed by a style load, a pattern CC7, or a start resetting an untouched part. They also say that Start/Stop resets only untouched parts.
- **#6 Player volumes on the port:** voice slot, OTS and Left levels now go out as CC7 on the port.
  - Ch 1 gets the lowest layered slot's volume, because a layer on one channel can't carry per-slot levels.
  - Ch 2 gets the Left volume.
  - These go to the port only, because a CC7 on ch 1 in the built-in synth would set every slot to the same level.
  - Manual Bass now plays at the Style's Bass part fader. That fader is also what ch 2 gets under Manual Bass.

**Gates:** `cargo build --release` passes, and `cargo test --release` passes (200 tests, 1 ignored). There are no new clippy warnings in the lines this PR touches.

Issue #13, opened 2026-09-23
Note limits narrower than an octave; per-zone Bass On


ryangavin:
- `fold_into` ignores ranges under 11 semitones. Work out and implement the hardware behaviour.
- Track Bass On per SFF2 zone instead of ORing it across zones.

---
Milestone: **M2 · Chord-to-note conversion 1:1** · Spec: [docs/genos-features.md](../blob/main/docs/genos-features.md) · Backlog: [docs/backlog.md](../blob/main/docs/backlog.md)

ryangavin:
These decisions are implemented in #57. The owner may overrule any of them after playtesting.

- **Decision:** a note with no octave inside a Note Limit narrower than an octave goes to the octave nearest the range, the lower one on a tie. **Because:**
  - the manual only says "every note is transposed automatically to fit into the range";
  - nearest-octave is the smallest possible violation of the author's range;
  - it reduces exactly to the old folding for ranges of an octave or more;
  - on a tie, the lower octave keeps a bass from being pushed up;
  - the corpus has 0 such ranges (208 styles, 7,529 rules; the narrowest in use is exactly 11 semitones).
- **Decision:** a reversed Note Limit (Low > High) is read as High..Low, **because** the Style Creator shouldn't produce one and swapping is the least surprising reading. The corpus has none.
- **Decision:** Cntt's single Bass On bit applies to all three zones, **because** Cntt already sets NTT for all three zones. Cntt vs Ctb2 precedence stays with #14.
- **Evidence for per-zone Bass On:** 582 Ctb2 records have different Bass On bits across zones, and 268 of them have a real Mid Low/High split. The typical case is a piano or guitar with a left-hand zone (Melody, Bass On) below Mid Low 50–54 and a chord zone (Chord, Bass Off) above it.
- **Follow-up for #12:** `guitar_voicing` applies the slash bass whether or not the zone has Bass On. `Zone::bass_on` is now there to gate it.

ryangavin:
Update from PR #57 round 1: I've withdrawn the Decision on Cntt Bass On ("applies to all three zones"). #57 now just ports the existing Cntt handler mechanically from the rule to its zones. Replace vs OR, and Cntt vs Ctb2 precedence, are decided in #14 / PR #56. The two narrow Note Limit decisions still stand, with one addition: at the MIDI range edges, where the nearest octave doesn't exist, the other octave is used so the pitch class is kept.

ryangavin:
[...]

PR #56, opened 2026-09-23
SFF1 NTT codes and Cntt precedence (#14)

ryangavin:
## What / why

This fixes how SFF1 (`Ctab`) NTT bytes and the optional `Cntt` override are read (#14).

**Bug fixed:** the `Cntt` handler *replaced* `bass_on` with the Cntt's bit 7. Every corpus style that has a `Cntt` writes its Bass channel's Cntt as `01H`: Melody with bit 7 clear. Its Ctab says `03H` Bass. So in all 7 of these styles the Bass part silently stopped following slash chords (Fingered On Bass). Now Bass On = Ctab "Bass" code **OR** the Cntt bit 7.

Changes in `src/sff.rs`:
- `decode_ntt_old` now decodes all 256 values and returns `(Ntt, bass_on)`. Before, anything above 5 fell back to Melody.
  - `00H`–`05H`: as documented by Wierzba/Bedesem.
  - `06H`–`0AH`: the Cntt/Ctb2 tables with the same numbers.
  - Anything else, `80H`–`FFH` included: Melody without Bass On. Ctab carries Bass On only as the `03H` "Bass" code.
- `Cntt` overrides the Ctab *table* in every zone. It ORs in Bass On, and it is ignored for `Ctb2` rules.

Documented under "SFF1 encoding (Ctab + Cntt)" in `docs/genos-features.md`. The `Cntt` item is removed from the open questions in `PLAN.md`.

## Corpus evidence (210 files)

- **Formats:**
  - 194 files are Ctb2 only.
  - 5 are Ctab only.
  - 7 are Ctab + Cntt (all `MOX_v2/*.T552.sty`).
  - 0 files mix `Cntt` with `Ctb2`. This matches Wierzba/Bedesem: "Cntt structures are not used if the style file contains Ctb2 structures."
- **Ctab NTT byte (78 records):** values 0/1/2/3/5 only. None is above 5, and none has bit 7 set.
- **All 25 Cntt records are 2 bytes**, and each one follows the Ctab for its channel. Cntt is written for every non-Bypass channel and never for drums. Pairs (Ctab NTR, Ctab NTT → Cntt):

  | Ctab NTR, NTT | Cntt | Count |
  |---|---|---|
  | RootTrans, Bass | `01` Melody (bit 7 clear) | 7 |
  | RootTrans, Melody | `01` | 7 |
  | RootTrans, Chord | `02` | 7 |
  | RootTrans, Harmonic Minor | `06` Harmonic Minor 5th | 4 |

[...]

Issue #12, opened 2026-09-23
Guitar NTR: All-Purpose, Stroke and Arpeggio tables


ryangavin:
- `guitar()` ignores the NTT value. Implement per-table voicing behaviour, and replace the string/position heuristic with a documented model.

---
Milestone: **M2 · Chord-to-note conversion 1:1** · Spec: [docs/genos-features.md](../blob/main/docs/genos-features.md) · Backlog: [docs/backlog.md](../blob/main/docs/backlog.md)

ryangavin:
Decisions for the Guitar NTR tables, from PR #60. The manuals give only each table's purpose (RM p.29–30), so these are our choices. The owner can overrule any of them after playtesting.

- **Decision: the source of a Guitar part is always read as C / CM7, because** RM p.29 says Source Root/Chord is ignored when NTR = Guitar, and all 43 Guitar rules in the corpus are recorded in C / M7 with CM7 guitar fingerings.
- **Decision: each note maps on its own (per note, no group voice leading), because** about half the corpus strums spread their notes over several ticks. A group-based mapping would voice the same strum differently depending on timing.
- **Decision: a note's chord role maps with the Chord table, and tensions and chromatic notes follow the Melody scale, because** this plays the recorded fingering unchanged over CM7. It also turns the common CM7 shapes into the standard guitar shapes for other chords: x32010 for C, the 32000x G, and C E B♭ D E for C9.
- **Decision: the shape slides by the root interval folded to −5…+6 semitones, with the tritone going up, because** a guitarist keeps a shape within about six frets of where it was recorded. It also keeps the part's register, as High Key does for Root Trans.
- **Decision: Guitar output never goes below E1 (MIDI 40), because** that is the open low E, and corpus authors set Note Limit Low = 40 on 15 of 43 Guitar rules. All Purpose and Arpeggio fold such a note up an octave. Stroke mutes it.
- **Decision: Stroke mutes only strings below the chord's bass, where the bass is the lowest root, or slash bass, at or above low E. It does not also mute strings over 1+8 / 1+5, because** the manual says "some notes may sound as if they are muted" to simulate strumming voicings, and below-the-bass is the muting every strummer does (x32010, xx0232). A power-chord mute would silence an upper-string strum over 1+8 even though the style enables the part for 1+8, and the existing corpus guard caught exactly that.
[...]

PR #57, opened 2026-09-23
Narrow note limits; per-zone Bass On (#13)

ryangavin:
## What

**Narrow note limits.** `fold_into` used to skip octave folding entirely when `hi - lo < 11`, so a note outside a narrow Note Limit just passed through unchanged. Now:
- A range of an octave or more works as before: plain octave folding, and notes inside the range are untouched.
- In a narrower range, every pitch class that has an octave inside the range lands there.
- A pitch class with no octave inside the range goes to the octave nearest the range, the lower one on a tie. The result is never more than a tritone outside the limits.
- A reversed range (Low > High) is read as High..Low.
- At the MIDI range edges the nearest octave may not exist (e.g. Ab into 120..127 would be 128). The other octave is taken then, so the pitch class always survives; the final clamp never changes a note's pitch class.
- Still branch-light integer math, no loops or allocation, so it's safe on the engine thread.

**Per-zone Bass On.** `bass_on` moves from `ChannelRule` to `Zone`:
- SFF2: read from bit 7 of each zone's NTT byte (low / mid / high).
- SFF1: set on every zone from `Ntt::Bass`.
- Cntt: ported mechanically. The existing handler's single bit now lands on every zone instead of on the rule; its semantics (replace vs OR, Cntt vs Ctb2 precedence) are unchanged here and belong to #14 / PR #56.

`transpose` now checks the zone the note falls in. Before, the three SFF2 bits were ORed together. So a piano whose left-hand zone follows slash chords also moved the roots in its chord zone to the bass note.

The spec (docs/genos-features.md) gains one line for each behaviour.

## Corpus evidence

Scanned 208 styles, 7,529 CASM channel rules (7,451 of them Ctb2):
- **Narrow limits: 0.** No active zone that follows the chord has a Note Limit narrower than 11 semitones, and none is reversed. Drum channels and Root Fixed + Bypass were excluded, and low/high zones only counted when Mid Low/High actually splits. The narrowest in use is exactly 11 (e.g. 82..93), which already folded correctly. The narrow-range rule is therefore a robustness choice and changes no corpus output.
- **Bass On differs between zones in 582 Ctb2 records.** 268 of them have a real split (Mid Low > 0 or Mid High < 127). The audible count is **240 records in 82 styles**: those where a reachable zone (low with Mid Low > 0, high with Mid High < 127) differs from the mid zone. (Corrected in round 1; the earlier 268 overstated it.) The common pattern, across many SX900 styles (6-8ChartBallad, 70sSpanishDisco, 90sAussiePop, ...), is a piano or guitar channel with Mid Low 50–54:
  - low zone: Melody with Bass On;
  - mid zone: Chord with Bass Off.
  With the old OR, the chord voicing's root was replaced by the slash bass. SmoothItOver (a golden style) has Mid High 95 with the high zone Bass Off.

## Golden

No digest changed. The golden script's slash chords (C/E, Am/G, C/G) don't hit a split zone whose Bass On differs from its neighbour's in the six golden styles.

## Decisions

- Decision: a note with no octave inside a Note Limit narrower than an octave goes to the octave nearest the range, the lower one on a tie, because:
  - the manual only says "every note is transposed automatically to fit into the range";
  - nearest-octave is the smallest possible violation of the author's range;
  - it reduces exactly to the old folding for ranges of an octave or more;
  - on a tie, the lower octave keeps a bass from being pushed up, which is the manual's stated purpose ("prevents high bass sounds");
  - no corpus style uses such a range, so there is no author intent to contradict.
- Decision: a reversed Note Limit (Low > High) is read as High..Low, because the Style Creator shouldn't produce one and swapping is the least surprising reading of corrupt data. The corpus has none.
- No decision on Cntt Bass On in this PR (withdrawn in round 1). The handler is a mechanical port of the existing replace; whether it should OR and how it interacts with Ctb2 is decided in #14 / PR #56, which will conflict textually with this PR in `parse_ctab`'s SFF1 branch and the Cntt handler (`rule.bass_on` vs `zone.bass_on`).

## Not changed (noted for #12)

`guitar_voicing` applies the slash bass to the bottom string whether or not the zone has Bass On. The spec says only Bass On guitar zones follow the bass. I left `guitar()` alone to avoid colliding with the Guitar NTR rework in #12. With `Zone::bass_on` now available, the fix is a one-liner there.

## How tested

- New unit tests:
  - `theory::fold_into_wide_ranges`, `fold_into_narrow_ranges` (including an exhaustive check over every lo in 0..=127 and every width 0..10 that the pitch class is kept, the result stays in 0..=127, and it is within a tritone of the range unless that octave doesn't exist), `fold_into_reversed_range`, `fold_into_midi_edges` (Ab into 120..127 -> 116, Bb into 0..5 -> 10, and C8 under Ab major through `transpose`);
  - `narrow_note_limit_transpose` (a bass locked to E1..G1 over all 12 roots);
  - `bass_on_per_zone` (low zone follows C/E, mid zone keeps C).
- `sff::bass_on_per_zone` parses 6-8ChartBallad and checks Piano ch 13: low zone Bass On, mid and high zones off. It then runs the rule through `transpose` over C/E: the left-hand C1 moves to an E, the chord-zone C3 plays exactly as over plain C.
- `cargo test --release`: 207 passed, 0 failed, 1 ignored. The corpus was symlinked and the corpus and golden tests ran; nothing was skipped.
- `cargo clippy --release`: no new warnings in touched code. The collapsible-if at `transpose`'s Bass On check was already there.

Closes #13


ryangavin:

Checked out `origin/m2/note-limits-bass-on` (3221c88) detached and symlinked the corpus.
- `cargo test --release`: 206 passed, 1 ignored. The golden and corpus tests really ran: no "skipping" lines, and the 6-8ChartBallad `.prs` is present.
- `cargo clippy --release`: nothing new in touched code. The `theory.rs:605` collapsible-if was already there.

I reran the corpus counts with a throwaway scanner over all 208 loadable styles:
- 7,529 rules, 7,451 of them Ctb2. That matches.
- No active narrow or reversed Note Limits. That matches.
- 582 records have Bass On differing between zones, and 268 of those have Mid Low > 0 or Mid High < 127. That matches.
- The per-zone Ctb2 layout (bit 7 of each zone's NTT byte) matches the SFF2 format. Zone selection by source key (`zone_for(key)`) is right.

### Blocking

1. **`fold_into` changes the pitch class near the MIDI range edges (`src/theory.rs:528-542`).** The new nearest-octave rule can pick an octave outside 0..=127. The final `n.clamp(0, 127)` then moves that note to 0 or 127, which is a different pitch class: a wrong note, not an octave shift. I proved it with a throwaway test:
   - `fold_into(68, 120, 127)` returns **127 (G)**. Expected **116 (Ab)**. The Ab above the range, 128, is 1 semitone out and wins, then gets clamped to G.
   - Through `transpose`: RootTrans/Melody, Note Limit 120..127, source C8 (120), chord Ab major gives **`Some(127)` = G**. So the root of an Ab chord plays a G.
   - `fold_into(10, 0, 5)` returns **0 (C)**. Expected 10 (Bb). The Bb at -2 wins and is clamped to C.
   - Over every range narrower than an octave with lo in 0..127, **3,686** (n, lo, hi) combinations come back with a different pitch class.
   - Before this PR, narrow ranges passed through unchanged. Wide ranges can't reach the clamp because they always land inside lo..hi. So this is a regression introduced by the new path.
   - The Style Creator lets Note Limit go from C-2 to G8, so these are valid inputs even though no corpus style uses one today.
   - The fix: when the nearest octave is outside 0..=127, take the other one. It is always inside, because lo..hi is inside 0..127.

2. **The "exhaustive tritone-bound check" isn't exhaustive, and it misses bug 1 (`src/theory.rs` test `fold_into_narrow_ranges`, the `for lo in 30..90u8` loop).** The PR body calls it "an exhaustive tritone-bound check over every range width 0..10". But lo only runs from 30 to 89, which skips exactly the edges where the result gets clamped. Please extend it to `lo in 0..=127` with `hi = (lo + w).min(127)`. It must also assert `(out - n) % 12 == 0`, which it already does. With those bounds it fails today, and it should pass after the fix.

### Nonblocking

3. **The Cntt Decision contradicts sibling PR #56, and the two PRs will conflict (`src/sff.rs:651-655`, PR body "Decisions" item 3).** This PR records "Cntt's Bass On bit applies to all three zones", and its code keeps the replace: `z.bass_on = d[1] & 0x80 != 0`. #56 shows that the replace is a bug. All 7 Cntt styles (`MOX_v2/*.T552.sty`) write the Bass channel's Cntt as `01H`, so with the replace their Bass parts ignore slash chords. #56 changes it to `|=` and skips Ctb2 rules.
   - Please don't post this as a Decision on #13. Drop it, or say that #14 owns it.
   - Expect textual conflicts with #56 in three places: `parse_ctab`'s SFF1 branch (#56 changes `decode_ntt_old` to return `(Ntt, bool)`), the Cntt handler, and `rule.bass_on` against `z.bass_on`.
   - Whichever PR merges second has to end up with `z.bass_on |= bit` on every zone, and only for rules where `!r.sff2`. #56's `cntt_overrides_ctab_table_and_keeps_bass_on` will catch a naive resolution, but only if it's ported to `zones[..].bass_on`.
   - None of the 7 Cntt files has zones with differing Bass On, so the resolution doesn't interact with this PR's per-zone logic.

4. **The corpus evidence overstates "audible".** The PR says "268 of them have a real split ..., so the difference is audible". But a split only makes the difference audible when the zone on the split's side is the one that differs. Only **240** records (in 82 styles) have a reachable zone whose Bass On differs from the mid zone's. The conclusion stands; the number should be corrected.

5. **The narrow-limit Decision is reasonable.** Both manuals are silent beyond "Every note is transposed automatically to fit into the range". The corpus has 0 uses. Nearest octave with ties going down is the least-surprise choice, and it reduces exactly to the old folding for ranges of an octave or more. I have no objection beyond bug 1.

6. **No behavioural corpus coverage of the per-zone change.** No golden digest changed, and `sff::bass_on_per_zone` only checks the parsed bits. Consider an end-to-end corpus assertion through `transpose`: on 6-8ChartBallad ch 13, under C/E a source C below Mid Low moves to E, and a C in the chord zone stays C. The unit test `theory::bass_on_per_zone` covers the same thing on a synthetic rule, so this isn't required.

7. **The #12 note about the guitar slash bass is right.** Leaving `guitar()` alone is the right call to avoid colliding with #12.

### Also checked
- **Real-time safety:** `fold_into` is branch-light integer math with no loops, allocation or panic paths. The integer divisions are by constants.
- **Mixer principle:** nothing here touches CC7, gain or the synth.
- **Public-repo hygiene:** no style content was committed. The new corpus test asserts only Mid Low and three Bass On bits, not note listings. No golden digests changed.
- **State:** there is no new runtime state, so there's nothing to reset on style, section or stop changes.

**Verdict: changes requested.** Blocking items 1 and 2 are a small fix plus widening one test loop.

ryangavin:
Round 1 review addressed in aeb456a.

**Addressed**
- Blocking: `fold_into` could choose an octave outside 0..=127 at the MIDI edges, and the clamp then changed the pitch class (Ab into 120..127 became 127 = G; Bb into 0..5 became 0 = C). It now takes the other octave when the nearest one doesn't exist. I reproduced the 3,686 wrong (n, lo, hi) combinations with the old logic; the new logic has 0. New test `fold_into_midi_edges` covers both reported cases and C8 under Ab major through `transpose` (now 116, not 127).
- Blocking: the exhaustive loop in `fold_into_narrow_ranges` now sweeps lo over 0..=127 with hi = (lo + w).min(127). For every case it checks that the pitch class is kept, the result stays in 0..=127, and it is within a tritone of the range unless that octave doesn't exist.
- Corpus count corrected in the PR body. The audible count is 240 records in 82 styles, where a reachable zone differs from the mid zone. I reproduced it with a scratch scan, which was not committed. 582 and 268 also reproduce.
- Added an end-to-end check: `sff::bass_on_per_zone` now runs 6-8ChartBallad's Piano ch 13 through `transpose` over C/E. The left-hand C1 moves to E, and the chord-zone C3 plays as it does over plain C.
- Cntt Decision withdrawn from #13. The handler is a mechanical port that keeps today's replace semantics, and #14 / PR #56 owns the replace-vs-OR and Ctb2 precedence question. The PR body notes the expected textual conflicts with #56 (the SFF1 branch of `parse_ctab`, the Cntt handler, and `rule.bass_on` vs `zone.bass_on`).

**Declined**
- Switching Cntt to `|=` here. I agree it looks like a bug given the MOX_v2 data, but #56 already fixes it together with Ctb2 precedence. Doing it here too would create a second, divergent fix. Whichever PR lands second resolves the conflict by applying #56's rule to each zone.

No golden digest changed. Gates: build clean; 207 tests passed, 0 failed, 1 ignored, and the corpus tests ran; no new clippy warnings in touched code.

ryangavin:

Checked out `origin/m2/note-limits-bass-on` (aeb456a) detached, with the corpus symlinked (`6-8ChartBallad.T547.prs` is present, so `sff::bass_on_per_zone` really runs).
- `cargo test --release`: 207 passed, 0 failed, 1 ignored.
- `cargo clippy --release`: nothing new in touched code. The warnings at `sff.rs:142` (OR pattern) and `theory.rs:606` (collapsible if) are both on the base commit.

### Round 1 fixes verified
- **Blocking 1 (the pitch class changed at the MIDI edges): fixed.** I checked `fold_into` against an independent brute-force oracle in a throwaway test (not committed). The oracle enumerates every octave of the pitch class inside 0..=127. If any octave is in range, it takes the lowest one for a note below the range and the highest for a note above. Otherwise it takes the nearest, the lower on a tie. I compared **every** `lo` and `hi` in 0..=127, reversed ranges included, with `n` in -40..170 to cover over- and undershoot from `transpose`. **0 mismatches.** The reported cases are fixed: `fold_into(68,120,127)` returns 116, and `fold_into(10,0,5)` returns 10.
- **Blocking 2 (the loop wasn't exhaustive): fixed.** The loop now sweeps `lo` over 0..=127 with `hi = (lo+w).min(127)`. It asserts the pitch class, the 0..=127 bounds, and the tritone bound with the edge exception.
- The corpus count is corrected to 240 records in 82 styles. An end-to-end check through `transpose` was added to `sff::bass_on_per_zone`, and the Cntt Decision was withdrawn. The reason for declining `|=` here is sound: #56 owns that change.

### Blocking
None.

### Nonblocking
1. **`transpose_group` can move the slash bass across zones (`src/theory.rs` `transpose_group`, the permutation step).** Least-movement reassignment permutes pitch classes across all keys in a group, whatever zone each key is in. Before this PR, Bass On was ORed across zones, so every zone produced the same set of pitch classes. With per-zone Bass On, a Root Fixed group that straddles Mid Low can hand the slash bass to a chord-zone key.
   - Synthetic repro: Root Fixed / Chord in all zones, Mid Low 50, Bass On in the low zone only. Chord Eb/C, keys `[45, 50, 51]`.
     - Transposing each key alone gives `[48, 53, 54]`, so the low key plays the bass C.
     - The group gives `[42, 48, 53]`: the C moves to key 50 in the mid zone, and the low key plays F#.
   - In that synthetic setup, 29,284 of 60,192 (chord, group) cases lose the bass from the low zone.
   - **Corpus impact: 0.** I scanned all 208 styles. There are 1,445 (section, channel) rules with reachable zones whose Bass On differs, and 2,549 simultaneous note groups straddle a Bass On boundary. None of them changes, because in every such rule the zones are Root Trans / Melody-type, so `transpose_group` returns early.
   - So this is latent. A cheap guard would be to skip the permutation when the group's keys disagree on `zone_for(k).bass_on` and the chord has a bass. It could also go to #12 / #14.
2. **`yahaha dump` prints only the mid zone's `bass_on` (`src/main.rs:57`).** Now that the bit is per zone, the dump can't show the very case this PR is about (low zone on, mid zone off). Consider printing all three, e.g. `bass {}{}{}`. This is only a debug tool.
3. The narrow-limit and reversed-range Decisions are still reasonable. The manual (Reference Manual p.31, Note Limit) says only that each note is "transposed automatically to fit into the range", and the corpus has 0 narrow or reversed active limits. The oracle confirms the implementation does exactly what the Decision describes.

### Also checked
- **Real-time safety:** `fold_into` is still straight-line integer math. The added guards are comparisons only, with no loops, allocation or panic paths. `Zone` stays `Copy`.
- **Mixer principle:** untouched.
- **Public-repo hygiene:** no style content or note listings committed, and no golden digest changed. The corpus test asserts only Mid Low, three bits, and pitch classes of synthetic source keys.
- **State:** there is no new runtime state. RTR revoicing calls `transpose` with the original source key, so `zone_for` picks the same zone on chord changes.
- **Conflicts:** as the PR body notes, expect conflicts with #56 in `parse_ctab`'s SFF1 branch and the Cntt handler. GitHub reports this PR as mergeable against the integration branch today.

**Verdict: no blocking issues.** Approve from the adversarial side. The owner merges.

ryangavin:
Review round 2, addressed in 1423583.

**Addressed**
- **transpose_group and per-zone Bass On:** a Root Fixed group now skips the cross-zone pitch-class permutation when the chord has a slash bass and its keys fall in zones that disagree on Bass On (`z.bass_on || ntt == Bass`). Each note keeps what its own zone gives it. New test `theory::tests::group_keeps_bass_on_per_zone` uses the reviewer's repro: Mid Low 50, Bass On in the low zone only, Eb/C, keys [45, 50, 51]. The group now gives [48, 53, 54], the same as each key alone; without the guard it gave [42, 48, 53]. The test also checks that without a slash chord the group still voice-leads across zones. Corpus impact is 0, as the reviewer found, and no golden digests changed.
- **`yahaha dump` Bass On:** now prints all three zones, low/mid/high, for example `bass 1--` for the low zone only or `bass 111` for all three.

**Declined**
- None.

Gates: `cargo build --release` passes. `cargo test --release` passes, and the corpus tests ran. `cargo clippy --release` gives no new warnings in the code I touched. The Decisions are unchanged.

ryangavin:
## Rebased onto `integration/m2-chord-conversion` (2569dd4)

Squashed to one commit, `ebc7c3c`, and rebased on #56, #55, #59 and #58.

- **`parse_ctab`, SFF1 branch (#56 vs this PR).** #56's `decode_ntt_old` returns `(Ntt, bass_on)`, and this PR moved `bass_on` into `Zone`. Each zone now takes the decoded Bass On. It is the same value as this PR's `ntt == Ntt::Bass`, but it comes from #56's decoder, so the decision stays in one place.
- **`Cntt` handler.** This PR ported the old *replace* (`z.bass_on = bit7`). #56 decided on **OR**. Resolved as `z.bass_on |= bit7` on every zone. That keeps #56's intent: all 7 MOX_v2 Cntt Bass parts still follow slash chords, and `corpus_cntt_bass_parts_follow_slash_chords` still checks it through `transpose`.
- **Tests.** Both sides' new tests are kept (#59's SInt tests and this PR's `bass_on_per_zone`). #56's tests read `rule.bass_on`, which no longer exists. They now assert Bass On on **every** zone, which is stricter, since SFF1 zones must agree.
- **Semantic conflict with #58, no textual conflict.** The oracle's `fold_target` said narrow Note Limits fold nothing, *"what the hardware does there is open (#13)"*. This PR decides #13, so `fold_target` now follows that rule: nearest octave, the lower one on a tie, only octaves that exist in MIDI, and a reversed range read low to high. I wrote it from the rule and did not call `theory::fold_into`, so the oracle doesn't grade the transposer against itself. Its tests cover the narrow, reversed and MIDI-edge cases. docs/oracle.md is updated to match. No corpus style has a narrow limit, so no count moves.

### Gates

- `cargo test --release`: **242 passed, 0 failed, 1 ignored**. Corpus and soundfonts were symlinked, and nothing skipped: I checked with `--nocapture`.
- `cargo clippy --release --all-targets`: the same warnings, per file and lint, as the integration tip. No new ones.
- **Golden digests: unchanged.** No regeneration was needed.

## Oracle delta against the old pin

`corpus_scores` failed on the integration tip because the pin predates #55 and #56. To attribute the delta, I cherry-picked the oracle (#58) onto each merge commit, ran `yahaha oracle corpus/ --scores` there, and diffed the results:

| Step | Commit | Change to the pin |
|---|---|---|
| M1 base + oracle | f962dc1 | none (it reproduces the old pin exactly) |
| **#56** SFF1 NTT codes / Cntt | b8e6874 | **none** |
| **#55** NTT 5th-variation tables | ab9ce6c | **97 lines** (everything below) |
| #59 structured SInt | f10febf | none |
| #58 oracle | 2569dd4 | none |
| **#57** (this PR, rebased) | ebc7c3c | **none** |

The re-pinned `tests/oracle/scores.txt` is byte-identical to the scores measured at #55. The whole delta is #55's.

**Why #56 and #57 don't move it:** the oracle plays every source on a chord without a slash bass, so Bass On never applies. #56's other change, decoding Ctab codes 06H–0AH, has no corpus use. The narrow Note Limits in #57 have no corpus use either (see the PR description).

### #55: mostly up

Per NTT table, across all 5210 pairs (249,112 notes), exact hits:

| Table | Old pin | Now | Δ |
|---|---|---|---|
| Harmonic Minor | 64.1% | 67.8% | **+3.8** |
| Natural Minor | 62.4% | 66.5% | **+4.0** |
| Dorian | 64.0% | 65.7% | +1.7 |
| Melodic Minor | 66.3% | 67.0% | +0.7 |
| each 5th variant | = its base table | its base + 4 notes | the 5th variants now differ (#11) |
| Bypass / Melody / Chord | | | unchanged |

- On the two chord changes that carry almost all the notes, `M>m` and `m>M` (117k notes each), every minor table rose. Harmonic Minor went from 66.1% to 68.9% on `M>m` and from 63.8% to 68.7% on `m>M`. Natural Minor rose about 2.8 and 5.8 points.
- **The baseline headline has flipped.** Harmonic Minor, not Melodic Minor, now agrees best on both `M>m` and `m>M` (68.9% vs 68.3% on `M>m`). docs/oracle.md said Melodic Minor, so I added a "Pin history" note there, and I updated the Limits paragraph about ties, since the 5th variants and Natural Minor no longer tie.
- Smaller chord changes also rose: `7>m` HM/MM went from 77.8% to 100% (90 notes), `m11>m` HM/MM from 31% to 56%, and `m7>maj7` rose by 4 to 13 points across the tables.

### #55: the drops, investigated

**As authored** fell from 151,488 to 151,368 exact (−120) and from 172,708 to 172,196 pitch class (−512), out of 249,112 notes. That is −0.05 points. The drop is in exactly two styles:

- `T5Style/BaroqueConcerto.T150.sst`: 80.2% → 78.4%.
- `T5Style/KoolFunk.T161.sst`: 41.6% → 40.5% exact, and 78.9% → 73.0% pitch class. The pitch-class drop means these are wrong notes, not wrong octaves.

(`BandaPolka` and `CountryTwoStep` went up.)

`--pairs` on the two styles traces every drop to **Melodic Minor 5th sources that convert between a 7 source and a maj7 source**. For example, BaroqueConcerto Main D ch12 has a Cmaj7 source and a C7 source, and each scored 100% as a copy of the other before #55. Now they score 90–95%, and some Fills score 82–87%. The author's C7 copy has the dominant 7th. Before #55, the scale model produced it. Now #55 applies the manual literally ("Other notes are not changed" for major→major), so the source's major 7th stays over C7, and the reverse happens too. The per-table drops have the same cause:

- `maj7>7` fell 0.7 points on every minor table.
- `m7>7` and `m>7` fell for Dorian and Natural Minor: −8.1 and −22.2 points (1498 and 90 notes). Those tables *raise* the 7th on a minor→major change, so a minor source plays B over C7.

Smaller drops: `m7b5↔maj7` fell by 2 of 16 notes, and `M>m11 / m11>M` HM/MM by 7 of 194 notes.

**Is it a real fidelity regression?** My read: **no audible regression in these corpus styles, but it is real evidence on #55's open question.**

- **Not audible here.** Every pair that dropped is a chord-mute alternate. The source is muted on the chord it's scored on, so the hardware never plays it there. The "as authored" row is hypothetical by construction (docs/oracle.md, Scores).
- **But it's evidence.** It is the only corpus signal on #55's **playtest flag**: a minor-table channel that is *not* muted on 7th chords plays a major 7th over C7. #55 counted 103 such base-table rules. Where authors wrote the dominant version out by hand, they wrote the chord's 7th, so the oracle leans against the literal reading for 7th chords.
- **The oracle can't settle it.** These pairs exist *because* the author split the channel. One reading is that the author split it because the table leaves the 7th alone, which would support #55. That is the selection bias docs/oracle.md warns about.

I didn't change #55's behaviour here, because that is a transposer decision and out of scope for this rebase. **Recommendation:** keep #55's playtest flag open, and cite these pairs (BaroqueConcerto Main D / Fill In BB/CC ch12/14/16, KoolFunk Main B–D ch12/13/16) as the test case when a Genos is available. If the hardware lowers the 7th over dominant chords, restoring the 7th handling should recover these ~120 exact and ~512 pitch-class notes, and the `m7>7` Dorian/NM rows.

Issue #8, opened 2026-09-23
Corpus self-consistency oracle


ryangavin:
- Many styles have separate source channels for major and minor chords, routed by chord mute.
- Play the major source through our NTT on a minor chord and compare the result with the minor source the author wrote. Score it per NTT table and per style. Track the score in CI and make every NTT change report its delta.

---
Milestone: **M2 · Chord-to-note conversion 1:1** · Spec: [docs/genos-features.md](../blob/main/docs/genos-features.md) · Backlog: [docs/backlog.md](../blob/main/docs/backlog.md)

ryangavin:
Decisions made in #58 (the owner may overrule after playtesting):

- Decision: the reference is the target source's raw notes on its own Source Root/Chord. Pairs whose target is chord-muted on its own source chord are skipped. Why: the spec (RM p.28) says playing exactly the source chord reproduces the pattern unchanged, so only then are those notes the author's intended sound. Running the reference through our transposer would make the oracle grade itself.
- Decision: a pair is scored only if B is an edited copy of A. That means at least half of each source's notes start on shared ticks with the same note count, at least 4 notes are aligned, and at least half of the aligned notes are within 2 semitones after the root move. Why: in the corpus, 1680 of 6890 alternative pairs are different lines rather than re-harmonised copies. Comparing those note for note would measure the arrangement, not the conversion. The test uses only style data, so the denominator never moves when the transposer changes.
- Decision: the per-table score swaps the NTT of every zone in the source's NTR family and keeps its NTR, High Key and Note Limit. Guitar sources try only the Guitar tables, and other sources try Bypass, Melody, Chord and the eight minor tables. Why: NTR = Guitar accepts only the Guitar tables (spec §C.5, RM p.28–31). Keeping everything else fixed isolates the table.
- Decision: notes are compared per onset as multisets, both as exact MIDI note and as pitch class. Why: `transpose_group` may reassign which voice gets which chord tone, and Note Limit folding moves octaves. The pitch-class column separates chord-type conversion from octave placement.
- Decision: the committed pin covers totals, per table, per NTR, per chord change × table and per style, and uses `UPDATE_GOLDEN=1` like the golden digests. Why: every NTT change should report its delta at the granularity it affects, and one regenerate switch keeps the workflow the same.

ryangavin:
These oracle decisions come from the PR #58 review (round 1):

[...]

PR #58, opened 2026-09-23
Corpus self-consistency oracle (#8)

ryangavin:
## What

This PR adds the corpus self-consistency oracle from #8.

Many styles give one part two source channels, one written over C major and one over C minor, and let the CASM chord mute choose between them. When the second source is an edited copy of the first, it shows what the author wanted the first to become on that chord. That gives us ground truth for note conversion.

`yahaha oracle corpus/` works like this:
- It plays each such source through the public `theory::transpose_group` on the chord its sibling was written for.
- It counts how many converted notes land on the author's notes, both as exact pitch and as pitch class.
- It scores every pair with the source's own CASM settings, then again with each NTT table.
- It sums the results per NTT table, per chord change (`M>m` etc., with the table that agrees best), per NTR and per style. `--pairs` lists each pair.
- It also runs an identity check: every chord-following source is played on its own source chord, which the spec says reproduces the pattern unchanged. Notes are counted per NTR of their own zone, and a note that moves inside its own Note Limit is counted separately from a Note Limit fold, because only the latter is documented behaviour.
- Notes go to `transpose_group` in the engine's groups: consecutive note-ons of the channel on one tick, in event order, at most 8.

Two more flags:
- `--scores` prints the pinned form.
- `--diff tests/oracle/scores.txt` prints the delta against it.

`oracle::tests::corpus_scores` pins the aggregate counts in `tests/oracle/scores.txt`. Each line is a key and either one count or one score (`notes exact pitch-class`). Styles are keyed by their path below `corpus/`, and only the styles the file lists are scored, so adding styles to the corpus does not move the pin. When a transposer change moves any count, the test fails and prints each changed line, with the exact-hit rate before and after for score lines. Regenerate with `UPDATE_GOLDEN=1 cargo test --release oracle` and quote the delta in the PR. The file holds only counts: style file names, table names, chord-change labels and integers. `committed_scores_hold_only_numbers` enforces that. No note content is written anywhere.

`docs/oracle.md` covers the method, how pairs are chosen, and the limits. It also records the first baseline.

The oracle depends only on `transpose_group`, `plays`, `Style::rules_for` and `Library::scan`. It does not touch transposer internals, so #11–#14 can change the tables freely. Their PRs will then show the oracle delta.

## Baseline (208 styles)

- **Scored:** 5210 pairs, 249,112 notes, in 166 styles.
- **Not scored:**
  - 1211 pairs are not edited copies.
  - 469 pairs have a source with fewer than 4 notes in the section (325 of them none).
  - 1884 pairs share a source chord.
  - 511 pairs have a reference that is muted on its own chord.
- **As authored:** 60.8% exact, 69.3% pitch class. This is not our fidelity. 93% of the converted sources (4831 of 5210) are Root Trans + Bypass, and each source is chord-muted on the chord it is scored on, so real hardware never plays it there. The number is close to the Bypass row by construction. The findings are the table comparisons.
- **By NTT table (exact):**

  | Table | Exact |
  |---|---|
  | Melodic Minor | 66.3% |
  | Harmonic Minor | 64.1% |
  | Dorian | 64.0% |
  | Melody | 62.4% |
  | Natural Minor | 62.4% |
  | Chord | 62.4% |
  | Bypass | 60.7% |

  For both `M>m` and `m>M`, the authors' minor versions agree best with Melodic Minor, which only moves the 3rd. The ties are artifacts, not findings about Yamaha. Each 5th variant ties its base table because our transposer does not implement the 5th variants yet (#11) and no scored pair involves aug or dim. Melody and Natural Minor tie on every chord change in the pin.
- **Identity**, per NTR of the note's zone:

  | NTR | Notes | Exact | Pitch class | Moved inside Note Limit |
  |---|---|---|---|---|
  | Root Trans | 330,282 | 86.1% | 100% | 1,426 |
  | Root Fixed | 101,646 | 99.4% | 100% | 0 |
  | Guitar | 96,638 | 10.4% | 18.1% | 86,606 |

  - Root Fixed: every miss is a Note Limit fold.
  - Root Trans: 44,559 misses are folds. **The other 1,426 notes are a transposer bug.** `theory::transpose` applies High Key even when the chord root equals the Source Root. With Source Root E and High Key D#, playing Em7 (the source chord) moves the pattern down an octave, even though the notes are inside their Note Limit. RM p.28 says the source chord plays back the recorded data. This PR reports the bug and pins its count as `identity RootTrans moved-in-limit 1426`; the fix belongs in the transposer, where that count should drop to 0.
  - Guitar: `guitar()` re-voices patterns even on their own chord. This is input for #12.
- **Gaps:** almost no dominant, dim, aug or tension pairs, and no Guitar pairs. The table scores therefore say little about the 5th variants (#11) or Guitar NTR (#12). The limits are in docs/oracle.md.

## How tested

- Unit tests on a synthetic two-source style:
  - An edited copy is scored in both directions. Bypass hits 4/6 and Melodic Minor 6/6.
  - A different line is not scored.
  - Same-source-chord pairs and muted-reference pairs are counted but not scored.
  - The delta shows the rate change on score lines (including per-style lines) and none on count lines.
  - Identity buckets each note by its own zone's NTR (a Guitar high zone above a Root Trans middle zone).
  - Identity separates Note Limit folds from moves inside the limit. This includes the Source Root E / High Key D# case.
  - Onsets are grouped like the engine: 10 simultaneous notes become 8 + 2, another channel's note-on splits a group, and a note-off does not.
  - Styles are keyed by relative path and filtered by the pin.
- `corpus_scores` ran against the real corpus (208 styles, no skip). The pair scores did not change when the grouping switched to the engine's groups. `yahaha oracle corpus --diff tests/oracle/scores.txt` prints `no change`. Combining two of the three mode flags is rejected with the usage message.
- Gates:
  - `cargo test --release`: 210 passed, 1 ignored (pre-existing), 0 failed.
  - `cargo clippy --release --all-targets`: no warnings in `src/oracle.rs` or in the lines I touched in `src/main.rs`.
- Golden digests are unchanged, since the transposer was not touched.

## Decisions

- Decision: the reference is the target source's raw notes on its own Source Root/Chord. Pairs whose target is chord-muted on its own source chord are skipped. Why: the spec (RM p.28) says playing exactly the source chord reproduces the pattern unchanged, so only then are those notes the author's intended sound. Running the reference through our transposer would make the oracle grade itself.
- Decision: a pair is scored only if B is an edited copy of A. That means at least half of each source's notes start on shared ticks with the same note count, at least 4 notes are aligned, and at least half of the aligned notes are within 2 semitones after the root move. Pairs where either source starts fewer than 4 notes in the section are counted separately as too few notes. Why: in the corpus, 1211 of 6890 alternative pairs have notes on both sides but are different lines rather than re-harmonised copies (another 469 have a source with fewer than 4 notes, 325 of them none, and prove nothing either way). Comparing those note for note would measure the arrangement, not the conversion. The test uses only style data, so the denominator never moves when the transposer changes.
- Decision: the per-table score swaps the NTT of every zone in the source's NTR family and keeps its NTR, High Key and Note Limit. Guitar sources try only the Guitar tables, and other sources try Bypass, Melody, Chord and the eight minor tables. Why: NTR = Guitar accepts only the Guitar tables (spec §C.5, RM p.28–31). Keeping everything else fixed isolates the table.
- Decision: the identity check buckets each note by the NTR of its own zone, and it counts a moved note as a transposer miss unless it was written outside a Note Limit of an octave or more and lands exactly on its fold into that limit. Why: RM p.28 says the source chord plays back the recorded data, and Note Limit folding (RM p.30) is the only documented reason for a note to move, by octaves into the limit. `fold_into` folds only limits of an octave or more.
- Decision: notes go to `transpose_group` in the engine's groups (consecutive note-ons of the channel on one tick, in event order, at most 8). Why: Root Fixed voice assignment depends on the group, so the oracle must grade what the engine actually plays.
- Decision: notes are compared per onset as multisets, both as exact MIDI note and as pitch class. Why: `transpose_group` may reassign which voice gets which chord tone, and Note Limit folding moves octaves. The pitch-class column separates chord-type conversion from octave placement.
- Decision: the committed pin covers totals, per table, per NTR, per chord change × table and per style, and uses `UPDATE_GOLDEN=1` like the golden digests. Why: every NTT change should report its delta at the granularity it affects, and one regenerate switch keeps the workflow the same.
- Decision: the pin scores only the styles it lists, keyed by path below `corpus/`. Why: `corpus/` is unversioned, so adding a style must not move a pin that the transposer did not change.

Closes #8


ryangavin:

Setup: I checked out `origin/m2/oracle` (c6a0d1d) detached and symlinked the corpus. `cargo build --release` passes. `cargo test --release` gives 206 passed, 0 failed, 1 ignored. `corpus_scores` really ran against the 208 styles and reproduces the committed pin. `cargo clippy --release --all-targets` shows no new warnings in `src/oracle.rs` or in the lines added to `src/main.rs`; the two `map_or` warnings in main.rs are pre-existing. I found no note content in the committed files: `scores.txt` holds file names, labels and integers, and `--pairs` prints only section, channel, chord name and counts. The oracle runs offline, so real-time safety and the mixer are not affected.

I proved the findings below with throwaway probes in `src/oracle.rs` and removed them afterwards. Nothing was pushed.

### Blocking

**1. The identity claims in the PR body and docs/oracle.md are wrong, and they hide a transposer bug.**
- The claim: "Both are 100% on pitch class, so every miss is an octave fold of a note written outside its own Note Limit."
- My probe compared each note with its own zone's `lo`/`hi`, for every chord-following, non-drum source played on its own source chord:
  - 45,164 moved notes are octave folds of notes outside the Note Limit. Those match the claim.
  - **1,426 moved notes are inside their own Note Limit and still move an octave.** By table: RootTrans/Bypass 1151, Melody 219, HarmonicMinor 52, Chord 4.
- Cause: `src/theory.rs:614-617` applies High Key even when the root does not move. Example: Source Root E (4), High Key D# (3), played chord Em7 (the source chord). `shift` is 0, then `chord.root(4) > high_key(3)` makes it -12. The whole pattern drops an octave on its own source chord (for example key 82 becomes 70; lo=40, hi=127).
  - RM p.28 (Source Root/Chord) says that playing the source chord "will play back the originally recorded data". So this is a fidelity bug in the transposer, and the oracle found it.
  - The PR explains it away as Note Limit folding instead of reporting it. Please correct the docs and PR text, and raise the bug for the transposer owners (#13/#14 or a new issue). It needs no fix in this PR.
- "Root Trans … 100% on pitch class" is also false. It is 360502/360629, and the 127 misses are rounded away in the display.
  - The misses are notes in **Guitar** high zones (keys 89/91/93/95) of rules whose *middle* zone is Root Trans.
  - `analyse` buckets identity by `to.zones[1].ntr` (`src/oracle.rs:238`), so Guitar-zone notes are counted as Root Trans.
  - Fix: bucket per note by `rule.zone_for(k).ntr`. Otherwise the per-NTR identity lines #12 will depend on are contaminated.

**2. The `delta()` rate is computed from the wrong numbers for `style` and `pairs` lines** (`src/oracle.rs:503-506`).
- `rate` always takes the last three integers.
  - On a `style` line those are the **identity** score, not the authored score.
  - On the `pairs` line they are unrelated/same_chord/muted_own.
- Proof, calling `delta()` directly:
  ```
  want: style X.sty 5 100 50 60 10 10 10   got: style X.sty 5 100 90 95 10 10 10
  →  style X.sty: [...] -> [...]  exact 100.0% -> 100.0% (+0.0)      (authored went 50% → 90%)
  want: pairs 10 1 2 3   got: pairs 10 1 2 4
  →  pairs: [10, 1, 2, 3] -> [10, 1, 2, 4]  exact 200.0% -> 200.0% (+0.0)
  ```
- The delta is this PR's main deliverable: the sibling PRs (#11–#14) are told to quote it. It currently reports "+0.0" on exactly the per-style lines where a regression would show.
- Fix options: compute the rate per score triple, or put the authored triple last. In either case, don't print a rate for the `pairs` and `styles` lines.
- `delta_shows_rate_change` only tests a `table` line, which is why this went unnoticed. Add a `style` line case.

### Nonblocking

1. **The pin depends on an unversioned input.** `corpus/` is git-ignored and local. Adding any style changes `styles`, `pairs`, `authored`, `table …` and `change …`, so `corpus_scores` fails for a reason unrelated to the transposer. The golden harness only checks styles that have a digest. Consider pinning only the styles listed in `scores.txt` and reporting (not failing on) new ones.
2. **"The same call the engine makes" is only approximately true.**
   - The engine (`src/engine.rs:1020-1047`) groups at most **8** *consecutive* note-ons of one source, in **event order**, and stops at any other event.
   - The oracle groups **all** note-ons at a tick, sorted, up to 16.
   - For Root Fixed (38 pairs and about 70k identity notes), `transpose_group`'s voice reassignment can then see different groups than live playback. Either match the engine's grouping or qualify the claim in the docs.
3. **`score()` truncates `a` to 16 notes but counts all of `b`** (`src/oracle.rs:190-193`). A chord with more than 16 notes would silently lower the score. It doesn't happen in this corpus; a `debug_assert` or a skip would make it explicit.
4. **Some baseline "findings" are artifacts.**
   - Every one of the 5210 pairs scores `X5 == X` for all four minor tables, and `Melody == NaturalMinor` in aggregate.
   - Both follow from the lack of aug/dim pairs and from the current table implementation. They say nothing about Yamaha.
   - docs/oracle.md states "Melody = Natural Minor" as a result. Phrase it as "indistinguishable on this data".
5. **"As authored" is essentially Bypass.** 4831/5210 pairs are RootTrans/Bypass (I confirmed this count), and on hardware those sources are muted on that chord and never converted. The headline "as authored 60.8%" is easy to misread as "our fidelity is 60.8%". Consider dropping it from the headline or relabelling it.
6. **Sibling churn.** `scores.txt` has per-style lines that include identity. #12 (Guitar re-voicing), a High Key fix and any NTT change will all rewrite it, so the parallel PRs will conflict on this file. Merge this PR first and have each sibling regenerate on rebase, or split the per-style section into its own file.
7. **Duplicate file names.** `delta()` keys `style` lines by file name only. Two styles with the same name in different folders would overwrite each other (none today). Consider using a path relative to the corpus.
8. **CLI flags.** `yahaha oracle … --scores --diff f` silently prefers `--diff`, and `--pairs` combined with `--scores` is silently ignored. This is minor.

### Decisions check
- Using the target's raw notes on its own source chord as the reference is right per RM p.28, and that page is what exposes blocking finding 1.
- The table families in `TABLES`/`GUITAR_TABLES` match RM p.29-30.
- My rerun matches the corpus counts: 5210 / 1680 / 1884 / 511, and 4831 RootTrans/Bypass.
- The edited-copy heuristic uses only style data, so the denominator stays independent of the transposer. Good.

Verdict: **changes requested.** The two blocking items are small: correct the identity analysis and bucketing, and fix the delta rate.

ryangavin:
Round 1 review addressed in 9c0d0a3.

## Blocking

- **False identity claims, and the High Key bug.** The reviewer is right. The identity check now counts a moved note two ways. A **fold** is a note written outside a Note Limit of an octave or more. A note that **moved inside its Note Limit** is a transposer miss under RM p.28. The corpus has exactly the 1,426 Root Trans notes the reviewer found, and none for Root Fixed. The cause is that `theory::transpose` applies High Key when the chord root equals the Source Root (Source Root above High Key, for example E with High Key D#).
  - docs/oracle.md and the PR body now call this a bug instead of saying every miss is a fold.
  - The pin tracks it as `identity RootTrans moved-in-limit 1426`.
  - Test: `identity_tells_note_limit_folds_from_transposer_moves` covers a real fold (not counted) and the E / D# case, which is counted exactly when `transpose` moves the note.
  - I did not fix the transposer. That is outside #8, and #11–#14 are editing the same function. It needs a ticket; the pinned count should drop to 0 when the fix lands.
- **Root Trans "100% on pitch class".** Identity is now bucketed per note by `zone_for(k).ntr`. Root Trans is now 330,282 notes at 100% pitch class. The Guitar high-zone notes moved to Guitar. Test: `identity_buckets_each_note_by_its_own_zone`.
- **delta() rate.** Every pinned line is now a key plus either one count or one score (`notes exact pitch-class`). delta() prints a rate only for score lines. Style lines are split into `pairs`, `authored`, `identity` and `moved-in-limit`, and the old `pairs` line is split into `pairs scored/unrelated/same-chord/muted-own`. Test: `delta_shows_rate_change` now checks that a per-style authored change from 50% to 90% prints `+40.0` and that a count line prints no rate.

## Nonblocking, addressed

- **The pin depends on the unversioned corpus.** The test and `--diff` score only the styles that scores.txt lists. Every style is listed, including styles without pairs, because they count toward identity. Adding styles to corpus/ no longer moves the pin, and the test prints how many it left out. Test: `styles_are_keyed_by_path_and_filtered_by_the_pin`.
- **Grouping differs from the engine.** The oracle now groups the way `emit_at_index` does: consecutive note-ons of the channel on one tick, in event order, at most 8. Note-offs are moved to the front of their tick, and aftertouch, sysex and meta events are ignored. Pair scores did not change on the corpus. Test: `onsets_group_like_the_engine`.
- **score() truncating at 16.** Gone. Every group is at most 8 notes and all of them are converted.
- **5th-variant and Melody/Natural Minor ties.** docs/oracle.md (Limits) and the PR body now call these artifacts. Our 5th variants are not implemented yet (#11), and Melody and Natural Minor tie on every chord change in the pin.
- **"As authored" is essentially Bypass.** The docs and PR body now say it is a hypothetical: the source is chord-muted on that chord on real hardware. It is not our fidelity; the table comparisons are the finding.
- **Style names keyed by file name.** Styles are now keyed by their path below corpus/, with whitespace replaced by `_`.
- **CLI.** `--pairs`, `--scores` and `--diff` are mutually exclusive. Combining any two prints the usage message.

## Nonblocking, declined

- **Split the per-style section into its own file.** Declined. Any transposer change that moves a per-style line also moves the totals, table, NTR and chord-change lines in the same file, so a second file would not prevent conflicts between sibling PRs. The practical fix is to merge #58 first. Sibling PRs then rebase and regenerate with `UPDATE_GOLDEN=1`, which rewrites the file anyway.

## Decisions added (also posted on #8)

- Decision: the identity check buckets each note by the NTR of its own zone. A moved note counts as a transposer miss unless it was written outside a Note Limit of an octave or more. Because RM p.28 says the source chord plays back the recorded data, and Note Limit folding is the only documented reason for a note to move.
- Decision: notes go to `transpose_group` in the engine's groups, because Root Fixed voice assignment depends on the group.
- Decision: the pin scores only the styles it lists, keyed by path below corpus/, because corpus/ is unversioned.

Gates: `cargo build --release` passes. `cargo test --release`: 210 passed, 1 ignored (pre-existing). The corpus tests ran on 208 styles. `cargo clippy --release --all-targets` shows no warnings in the changed code.

ryangavin:

**Setup.** I checked out `origin/m2/oracle` detached at 9c0d0a3 and symlinked the corpus.
- `cargo test --release`: 210 passed, 0 failed, 1 ignored (pre-existing).
- `corpus_scores` ran against the real corpus. It did not skip: the probes below read the same 208 styles.
- `cargo clippy --release --all-targets`: no warnings in `src/oracle.rs` or in the added lines of `src/main.rs`. The two `map_or` warnings at main.rs:33 and :203 were there before this PR.
- `yahaha oracle corpus/ --diff tests/oracle/scores.txt` prints `no change`.
- I used throwaway probes in `src/oracle.rs` and `src/theory.rs` and reverted them all. Nothing was pushed.

### Round 1 fixes, verified

1. **Identity split into folds and moves inside Note Limit.** I checked every note the oracle counts as a "fold": does the transposer output equal `fold_into(key, lo, hi)` exactly, and is the pitch class kept?
   - Root Trans: 44,559 folds, 0 that differ from the true fold.
   - Root Fixed: 605 folds, 0 that differ.
   - Guitar: 0 folds.
   - All 1,426 Root Trans moves inside the limit have `src_root > high_key`, so every one comes from the High Key-on-own-chord case.
   - Then I patched `theory.rs:615` to skip High Key when the chord root equals the Source Root. `corpus_scores` failed with `identity RootTrans moved-in-limit: [1426] -> [0]` and `exact 86.1% -> 86.5%`, and per-table/per-change deltas printed with correct rates. The pin works as a regression detector, which is the point of #8.
2. **Per-zone bucketing.** Root Trans now reads 330282/330282 on pitch class, and the Guitar high-zone notes are in the Guitar bucket. `identity_buckets_each_note_by_its_own_zone` covers this.
3. **delta() rate.** Every line now holds either one count or one score triple, so the round 1 proof (a per-style authored line going 50% → 90% and printing +0.0) can no longer happen. The test now covers a style line.
4. **Grouping matches the engine.** I compared `onsets()` against `engine.rs:1019-1037`:
   - Engine side: the Prepared filter keeps On/Off/Cc/Pc/Bend, and a stable sort puts note-offs first within a tick.
   - `onsets()` side: it splits on another channel's On, or on any Cc/Pc/Bend. It ignores Off and dropped events, starts a new group after 8, and treats a tick change as a break. Velocity-0 note-ons are already parsed as Off (`sff.rs:442`).
   - The two match.
5. **Pin filtered by listed styles, keyed by path.** Verified. CLI flags are now exclusive (`--pairs --scores` exits 1 with the usage message).

### Blocking

None.

### Nonblocking

1. **The High Key bug still has no ticket.** The round 1 reply says "It needs a ticket", but the only record is a comment on #8, which this PR closes. Once #8 is closed the 1,426 finding has nowhere to live. #13 and #14 don't mention it. Open an issue, or comment on #13, citing `theory.rs:614-617` and the pinned `identity RootTrans moved-in-limit 1426`.
2. **The `folds` test is looser than its doc comment** (`src/oracle.rs:268`). It excuses any move of a note written outside the limit, even a move to a wrong pitch or octave. The probe above shows this happens 0 times today. But a sibling that breaks folding for out-of-range notes, or adds a High Key shift on top of a fold, would not show up in `moved-in-limit`, only in the aggregate exact rate. A stricter test costs about one line: `let folds = ... && o == Some(fold(k))`, with a local copy of the octave-fold rule. It gives the same count today.
3. **The "not edited copies" count is inflated by empty sources, and the Decision cites it as evidence.** I probed the 1,680 unrelated pairs:
   - **325** have a source with no notes at all in that section.
   - **469** have a source with fewer than 4 notes.
   - So "1680 of 6890 alternative pairs are different lines rather than re-harmonised copies" overstates it: about a fifth are empty channels, not different lines.
   - Scores are unaffected, since this is only a denominator label. Consider an `empty` count, or reword the Decision and the docs.
4. **`--diff` on a subfolder or a single style gives a misleading delta.** `yahaha oracle corpus/MOX_v2 --diff tests/oracle/scores.txt` prints 1,151 lines. Keys are relative to the root you pass (`AustinCityBlues…` instead of `MOX_v2/AustinCityBlues…`), so every style shows as `gone` and `authored` shows `-60.8`. Even with matching keys, the aggregate lines would compare a subset against full-corpus totals. Either document that `--diff` needs the same root as the pin, or when the set of styles differs, print just the per-style lines plus a note.
5. **`identity_tells_note_limit_folds_from_transposer_moves` never checks that its scenario is exercised** (`src/oracle.rs:777-782`). It asserts whatever `transpose` does, so it keeps passing after the High Key fix (intended), but it would also pass if `moved` were false today. Consider asserting `moved` today, with a comment to flip it when the transposer is fixed, or adding a synthetic case with a known move inside the limit.

### Content and safety check
- `scores.txt` contains only paths, labels and integers. `--pairs` prints section, channel, chord name and counts.
- No note content is committed. The golden digests are untouched.
- This is an offline tool, so there is no real-time, mixer or engine-state impact.

### Verdict
**Approve from the adversarial side.** Nothing blocking remains. I'd still file the High Key issue (item 1) before or at merge, so the finding survives #8 closing.

ryangavin:
Review round 2 addressed in 5ecd9d3. There were no blocking items.

**Addressed**
- **High Key bug has no issue of its own:** I commented on #13 with the details (https://github.com/ryangavin/yahaha/issues/13#issuecomment-5801989975). #13 is the octave-placement ticket.
- **`folds` excused any move of a note outside its limit:** a note now counts as folded only when it lands exactly on `fold_target(k, lo, hi)`, the octave RM p.30 puts it in. Anything else counts as moved. `fold_target_is_the_octave_inside_the_limit` tests it, and so does an injected transposer that moves a note two octaves instead of one: it counts 1 moved. Corpus pin unchanged, which matches the probe (every fold today is exact).
- **The "not edited copies" count overstated the evidence:** pairs where either source starts fewer than 4 notes in the section are now a separate count, `pairs too-few 469`. 325 of those have no notes at all. `pairs unrelated` drops from 1680 to 1211. I reworded the Decision in the PR body and updated docs/oracle.md. Test: `a_source_without_notes_is_too_few_not_unrelated`. This is the only pin change:
  ```
  pairs unrelated: [1680] -> [1211]
  pairs too-few: new [469]
  ```
- **`--diff` on a subfolder printed bogus lines:** the new `oracle::diff_against` handles this.
  - Each style is re-keyed to the single pinned key that ends in its path.
  - When the run covers only some of the pinned styles, it says so and compares only those styles' own lines, not the totals.
  - `yahaha oracle corpus/MOX_v2 --diff tests/oracle/scores.txt` now prints `35 of 208 pinned styles in this run ...` / `no change`.
  - Test: `diff_against_a_subset_run_compares_only_its_styles`.
- **The identity test asserted whatever `transpose` did:** it now uses `identity_with` with injected transposers that have known moves: a move inside the limit, a note outside the limit moved by the wrong amount, and a narrow limit. I chose this over asserting `moved` against the real transposer, so that the High Key fix does not have to edit this test.

**Declined:** none.

Gates: `cargo build --release`, and `cargo test --release` (213 pass, and the corpus oracle and golden tests ran). `cargo clippy --release` shows no warnings in the touched code.

Issue #15, opened 2026-09-23
Parse SInt properly and re-apply it on section change


ryangavin:
- Parse SInt into structured per-channel init (bank, program, volume, pan, sends, XG effects).
- Re-apply it on every section change, the way newer instruments do.

---
Milestone: **M2 · Chord-to-note conversion 1:1** · Spec: [docs/genos-features.md](../blob/main/docs/genos-features.md) · Backlog: [docs/backlog.md](../blob/main/docs/backlog.md)

ryangavin:
Decisions made in #59 (the owner may overrule after playtesting):

- Decision: re-apply the SInt only when the section actually changes, not when a section repeats itself, because the ticket says "on every section change". The corpus has 1244 program changes after tick 0 inside sections, and re-sending on every loop would undo them each time round.
- Decision: on a section change, a part fader the player has not moved goes back to the SInt's CC7, and a moved one keeps its level and is not re-sent. The SInt CC7 is style data, like a pattern CC7, which already moves untouched faders only. This also matches what Start already did.
- Decision: never send the SInt's system resets (GM System On/Off, GM2 On, XG System On, XG All Parameter Reset, GS Reset). Those messages reset every channel of the receiver, including the player's own parts on ch 1–2 and the mixer levels, and not just the style's parts. They are in 207 of 208 corpus styles.
- Decision: route XG Multi Part parameters through the same Main A rules as the CCs (part number = channel), rewritten to device 0x10. Drop XG part Volume on parts 9–16, because the part fader owns the volume. No corpus style has one.
- Decision: send each part as bank, program, controllers, then its XG part parameters, and the global effect and drum SysEx last, instead of file order. A program change on an XG receiver resets the part parameters, and drum setup needs the part mode in place first.
- Decision: SysEx goes to the `yahaha` port only. The built-in synth does not implement XG effects, and its input ring holds 3-byte messages.

ryangavin:
Decisions added in PR #59 after review round 1. The owner may overrule.

- Decision: on a section change, re-send only the SInt's part setup (bank, program, controllers, XG part parameters), not the XG effect and drum setup SysEx. Patterns never send SysEx, so there is nothing to restore, and an XG receiver resets an effect when it gets its type again, which would cut reverb and delay tails at every fill. Start and style load still send all of it.
[...]

Issue #22, opened 2026-09-23
Section Change Timing settings (Main: Immediate/Next Bar; Intro/Ending: Next Bar/End of Section)


ryangavin:
_Details in the spec._

---
Milestone: **M5 · Sections & transport parity** · Spec: [docs/genos-features.md](../blob/main/docs/genos-features.md) · Backlog: [docs/backlog.md](../blob/main/docs/backlog.md)

ryangavin:
Implemented in #94 (Engine::change_point, docs/section-timing.md).
- Decision: Next Bar includes the RM p.12 rule 'at once when pressed within the first beat', for Mains, style changes and Intro/Ending changes, because the manual defines Next Bar that way.
- Decision: into Ending I, from a Main/Fill into an Intro/Ending, and the missing-Ending stop keep the old next-bar-line rule, because the manual's 'conventional rules' are not described and that was yahaha's behaviour.
- Decision: both settings default to Next Bar, because no factory defaults are listed and it is closest to the old behaviour.
- Decision: a style change stays Immediate even with Auto Fill on, because the RM fallback names section changes only.

ryangavin:
Done in #94, merged to main via #113.

PR #59, opened 2026-09-23
Structured SInt, re-applied on section change (#15, #50)

ryangavin:
## What / why

**#15: structured SInt, re-applied on section change**
- `sff::SInt::parse` (and `Style::sint()`) turns the style's init events into per-channel `ChannelInit`: the bank in effect at the program change, program, a bank select after the last program change (pending: it selects no voice, and goes out after the program change), volume (CC7), pan (CC10), reverb (CC91), chorus (CC93), XG Multi Part parameters (`F0 43 1n 4C 08 pp aa vv F7`, filed under channel pp), and every other controller and pitch bend in file order. RPN/NRPN sequences need that order. SysEx that isn't a part's (XG effect, insertion effect and drum setup, Yamaha model-specific) stays in a list in file order. System resets are left out.
- `Prepared::init` is now `Msgs`: variable-length messages stored back to back and built at load time, so SysEx can go out from the engine thread without allocating. Each part is routed through the Main A rules as before, in this order: bank, program, controllers, XG part parameters (after the program change, which resets them on an XG receiver). The drum setup SysEx (`4C 30/31`, Drum Setup Reset `4C 00 00 7D`) follows, then the effect SysEx. In the effect SysEx, an Insertion Effect Part (`4C 03 nn 0C pp`) or Variation Part (`4C 02 01 5B pp`) assignment is moved to part pp's destination channel, or set off (`7F`) when the style never routes part pp.
- `Engine::transition` re-applies the SInt's part setup (voices, controllers, XG part parameters) and its drum setup SysEx, but not the effect SysEx, when the section actually changes (`next != cur`), after `all_off` and before the new section's first event. The new section's own tick-0 changes therefore still win. The full SInt, effects included, goes out on start and style load.
- A section entered mid-bar (a Fill or Break at the next beat) now also sends its skipped controller, bank, program and bend events, in order, right after the SInt (`Engine::chase`). Its skipped notes stay skipped. Without this, the part played the SInt voice for the rest of the bar.
- **Mixer:**
  - The SInt's part CC7 never goes out as is. It is `Prepared::mix`, the style's level.
  - On a section change, a part the player has **not** moved goes back to that level (fader and CC7), like a pattern CC7 moving an untouched fader. A part the player **has** moved keeps its level and gets no CC7.
  - Start and style load behave as before.
  - An XG Multi Part *Volume* SysEx on parts 9–16 is dropped, because the fader owns the volume. No style in the corpus has one.
- `live::Out` keeps SysEx off the built-in synth's 3-byte ring. It goes to the `yahaha` port only.
- `yahaha dump` prints the structured SInt, one line per channel.

**#50 (first bullet): raw data bytes**
- `key_msg` already masked the key (`& 0x7F`) on integration. The root cause is in `midi::for_each_message`, which copied any byte into a data position. So `[0x90, 60, 0x90, …]` produced a message with a data byte of `0x90`, and that reached every consumer (pads, synth program, etc.).
- It now follows the MIDI spec:
  - A status byte in a data position ends the cut message: it is dropped, and reading continues from the status byte, with running status updated.
  - Realtime bytes inside a message are skipped without breaking it.
- `key_msg` keeps its masks.

## Corpus evidence (208 styles)
Across the corpus's SInt parts:
- GM System On appears in 207 styles and XG System On in 184.
- There are 1126 XG effect (`4C 02 01`) messages and 998 insertion effect (`4C 03`) messages (241 of them part assignments `4C 03 nn 0C`; 5 Variation Part `4C 02 01 5B`), plus drum setup (`4C 30/31`).
- XG Multi Part parameters used: 07 (part mode), 11 (dry level), 20, 22, and 72/73/76/77 (EQ). Volume (0B) and pan (0E) are never set by SysEx.
- 14 channels in 8 styles send a bank select after their last program change (e.g. ChartPop1 ch 15: CC0 8, CC32 2, PC 3, CC0 104, which selects 8/2/3).
- Insertion effects assigned to a part CASM moves: CountryTwoStep, SoulSwing and KoolFunk; to a part Main A never routes: JazzWaltzFast and Bebop.
- The Yamaha model-specific messages (`43 73 …`, `43 76 …`) are passed through unchanged.

## How tested
- `sff::sint_parses_per_channel`: the fields, last value wins, order of the other controllers and bend, XG part parameters filed by part, resets dropped.
- `sff::sint_structures_every_corpus_style`: in every corpus style, each channel's last value of every controller and program is where the structure puts it, a receiver fed the structure ends with the same voice and bank registers as one fed the file, and every non-reset SysEx is kept (ran on 208 styles).
- `sff::sint_bank_after_program_is_pending` and `sim::mixer::sint_voice_keeps_the_bank_of_its_program_change` (corpus: ChartPop1 ch 15 is 8/2/3, Let'sFunk ch 15 is 104/5/0).
- `sim::mixer::effect_part_assignments_follow_the_routing`: synthetic CASM plus CountryTwoStep (0E, 0F become 0D, 0C) and JazzWaltzFast (unrouted 0B becomes 7F).
- `sim::mixer::mid_bar_fill_plays_its_own_voice` and `mid_bar_fill_voice_in_corpus` (ChartPop1 Fill A from beat 3: ch 15 is on Fill A's 104/8/4). Both fail without the chase.
- `sim::mixer::init_is_structured_and_mixer_owns_volume`: the exact message order, and no reset, part CC7 or XG part volume.
- `sim::mixer::section_change_reapplies_init`:
  - A Main A repeat sends no SInt.
  - Main B gets the voice, XG part parameters and the untouched part's level back (fader and CC7), and no effect SysEx.
- `sim::mixer::section_change_resends_drum_setup_after_program_changes`: the drum setup goes after the part's program change and part mode at Start and again at a section change; the effect SysEx does not. `sim::mixer::corpus_drum_setup_survives_section_change`: AustinCityBlues, Main A to B to C to A: every change sends the whole drum setup after all the SInt program changes.
  - The fader the player moved keeps its level and gets no CC7.
  - The SInt goes out before the new section's first note.
- `sim::mixer::section_change_restores_untouched_style_levels` (corpus): TickingAway's Intro B sets part 2 to 76; Main A now starts from the SInt's 90.
- `sim::mixer::restart_restores_untouched_style_levels`: this test used to carry Intro B's 76 through Main A to the stop. Main A now resets it, so the test now leaves the level set by Ending C (the last section before the stop) and checks that the restart restores it.
- `live::sysex_skips_the_synth`, `midi::data_bytes_are_always_below_0x80` (every byte value in the data positions; the #50 regression test). `live::malformed_key_bytes_never_panic` is a guard for the whole input path; it also passes on integration, because `key_msg` already masked.
- Gates:
  - `cargo test --release`: 213 passed, 0 failed, with the corpus present (golden and corpus tests executed).
  - `cargo clippy --release` has no new warnings.
- The golden digests are unchanged, because they hash notes only.

Not verified on hardware: CoreMIDI packet-list behaviour with SysEx mixed into the same timestamp as channel messages. `MIDIPacketListAdd` is documented to keep SysEx in its own packet.

## Decisions
- Decision: re-apply the SInt only when the section actually changes, not when a section repeats itself, because the ticket says "on every section change". The corpus has 1244 program changes after tick 0 inside sections, and re-sending on every loop would undo them each time round.
- Decision: on a section change, a part fader the player has not moved goes back to the SInt's CC7, and a moved one keeps its level and is not re-sent. The SInt CC7 is style data, like a pattern CC7, which already moves untouched faders only. This also matches what Start already did.
- Decision: never send the SInt's system resets (GM System On/Off, GM2 On, XG System On, XG All Parameter Reset, GS Reset). Those messages reset every channel of the receiver, including the player's own parts on ch 1–2 and the mixer levels, and not just the style's parts. They are in 207 of 208 corpus styles.
- Decision: route XG Multi Part parameters through the same Main A rules as the CCs (part number = channel), rewritten to device 0x10. Drop XG part Volume on parts 9–16, because the part fader owns the volume.
- Decision: send each part as bank, program, controllers, then its XG part parameters, and the global effect and drum SysEx last, instead of file order. A program change on an XG receiver resets the part parameters, and drum setup needs the part mode in place first.
- Decision: SysEx goes to the `yahaha` port only. The built-in synth does not implement XG effects, and its input ring holds 3-byte messages.
- Decision: on a section change, re-send the part setup (bank, program, controllers, XG part parameters) and then the drum setup SysEx, but not the XG effect SysEx. The drum setup must go again because the Genos Data List (Drum Setup, Note) says a program change on a part with a drum setup initializes that drum setup, and the section change re-sends the parts' program changes (170 of 208 corpus styles have drum setup SysEx, 2249 messages). The effect SysEx stays out because patterns never send it, and an XG receiver resets the effect when it gets its type again, cutting reverb and delay tails at every fill. Start and style load send all of it, in the same order.
- Decision: an Insertion or Variation effect assigned to a part the style never routes (Main A rules) is set off (`7F`) rather than dropped, because the source part never sounds, and dropping the message would leave that insertion on whatever part the previous style assigned.
- Decision: a section entered mid-bar sends its skipped non-note events in order (chase), not only at bar-aligned entries, because a Fill's first-beat voice is part of the Fill, and the review counted 21 of 208 corpus styles with a Fill or Break whose first-beat program change differs from the SInt.
- Decision (#50): validate in the stream splitter instead of only masking at each use. A status byte in a data position ends the message, as the MIDI spec says, and realtime bytes inside a message are allowed.

Closes #15
Closes #50


ryangavin:

I checked out `origin/m2/sint` (c79a913) with the corpus linked.
- `cargo test --release`: 208 passed, 0 failed. The corpus tests ran on 208 styles.
- `cargo clippy --release`: no new warnings in the lines this PR touches.
- I wrote throwaway probes against the corpus to check the findings below. None of them are committed.

What holds up:
- **Mixer principle:** the SInt's part CC7 becomes `mix`, a moved fader is not re-sent, and an XG part volume on parts 9–16 is dropped.
- **Real-time safety:** `Msgs` is built at load time and iterated without allocating, and `PacketSink` flushes when its buffer is full.
- **#50 splitter:** correct.
- **Corpus counts I could reproduce:**
  - GM On in 207 styles and XG On in 184.
  - 1126 `4C 02 01` messages.
  - 1244 PCs after tick 0.
  - XG part addresses {07, 11, 20, 22, 72, 73, 76, 77}. 0B and 0E are never used.
- **Leaked content:** none. The tests use synthetic bytes, or level numbers only.

### Blocking

**1. XG insertion effect part assignments are not remapped, but XG part parameters are** (`src/engine.rs` `Prepared::new`, `for v in &sint.sysex { init.push(v) }`)
- The Insertion Part message (`F0 43 10 4C 03 nn 0C pp F7`) carries the *source* part number, and it goes out raw.
- The XG Multi Part parameters of the same source part are rewritten to its destination (`d`). So the effect and the voice end up on different parts.
- Corpus example, CountryTwoStep.T151: Main A routes source ch 15 to dest ch 14 and source ch 16 to dest ch 13.
  - The SInt sends `4C 03 00 0C 0E` and `4C 03 01 0C 0F` unchanged.
  - Result: insertion 1 lands on dest ch 15, which yahaha uses for the "Tele" parts, not PedalStl. Insertion 2 lands on ch 16, where nothing is routed.
  - Expected: the pedal steel on ch 14 and the piano on ch 13 get their own insertion effects.
- Probe result: `insertion part assigns sent: [..0C 0E], [..0C 0F]` while source part 15 routes to dest 12 (0-based).
- Affected styles:
  - Remapped to a different part: CountryTwoStep, SoulSwing, KoolFunk.
  - Pointing at a source channel Main A doesn't route (the effect lands on whatever plays there): JazzWaltzFast, Bebop.
- Fix: rewrite `4C 03 nn 0C pp` (and Variation Part, `4C 02 01 5B pp`) through `route[pp]`. Drop the message when `route[pp]` is `None`.
- Add a test that fails today.

**2. Re-applying the SInt on a mid-bar entry resets the voice, and the new section's own setup is then skipped** (`src/engine.rs` `transition` → `reapply_init`, then `seek(at - start)`)
- Fills and breaks come in at the next beat, and `queue_fill` sets `sec_start` to the bar start. `seek` then skips every event before the entry point, including the fill's tick-0 bank and PC.
- Before this PR, the part kept the previous Main's voice, which usually matches the fill's. Now `reapply_init` first puts it back to the SInt voice, and nothing restores the fill's voice until the next section.
- Trace, ChartPop1.T160, dest ch 15, Fill A pressed in bar 2 and entering at beat 3:
  - `t=0`: SInt sets 104/2/3. Main A tick 0–8 then sets 104/8/4.
  - `t=3564ms` (fill entry): `reapply_init` sends 104/2/3. Fill A's own `CC0 104, CC32 8, PC 4` at ticks 0–8 are skipped by `seek`.
  - Until the next bar, the part is on the SInt voice instead of 104/8/4, the voice both Main A and Fill A use.
- Corpus count: 21 of 208 styles have a Fill or Break with a PC in its first beat that differs from the SInt for that channel, and notes after the first beat. Any mid-bar entry into one of those sections plays that part with the wrong voice.
- The upcoming Section Change Timing = Immediate (#22) will hit the same path for Mains.
- Fix, one of:
  - chase the skipped non-note events (bank, PC, CC, bend) from tick 0 to the entry point in `seek`, after `reapply_init`; or
  - re-apply only at bar-aligned entries.
- The first matches what hardware sequencers do and fixes a bug that was already there. Add a mid-bar fill test. `section_change_reapplies_init` only covers bar-aligned changes.

**3. Collapsing bank and program to "last value wins" changes which voice is selected** (`src/sff.rs` `SInt::parse`; `Prepared::new` sends bank, then PC)
- A bank select only takes effect at the next program change. Some SInts end a channel with a bank select *after* its last PC. Reordering to bank then PC selects a voice the file never selects.
- ChartPop1.T160, ch 15, SInt in file order: `CC0 8, CC32 2, PC 3, CC0 104`.
  - The file selects 8/2/3.
  - This PR sends `CC0 104, CC32 2, PC 3`, which selects 104/2/3. `voices[14]` changes from (8,2,3) on integration to (104,2,3).
- Let'sFunk.T161, ch 15: integration selects (104,5,0), this PR selects (8,5,0).
- Corpus: 14 channels in 9 styles have a trailing bank select after their last PC. Two of them change a routed voice.
- `sint_structures_every_corpus_style` only compares last values, so it cannot catch this.
- Fix: record `bank_msb`/`bank_lsb` as the values in effect when the PC arrives. Keep a trailing bank select as a pending value, sent after the PC. Test it with the corpus.

### Nonblocking

- **Effect and drum SysEx on every section change.** Patterns can't send SysEx: `Prepared::new` drops `Ev::Sysex` from sections. So re-sending the effect and drum setup on each change restores nothing.
  - On XG receivers, re-sending an effect type resets that effect, so reverb and delay tails get cut at every fill.
  - It is also up to 153 messages / 659 bytes per change across the corpus. That is about 210 ms of the downbeat on a 5-pin DIN link.
  - Suggest re-sending only the per-part bank, PC and CCs on a section change, and the full SInt at start and load.
- The PR body says the corpus has 1116 insertion effect (`4C 03`) messages. I count 998. It doesn't matter for the change, but please re-check the figure.
- `restore_untouched_levels` now calls `takeover.software_moved` on every section change for untouched parts. So an untouched hardware fader is re-armed for pickup at every change, including one where the level ends up unchanged. That is harmless, but skip it when `v == mixer[p]`.
- `live::malformed_key_bytes_never_panic` also passes on integration, because `key_msg` already masked the key. The real regression test for #50 is `midi::data_bytes_are_always_below_0x80`, which is good. Say so in the PR body, or have the live test drive a status-in-data packet and assert the result.
- Decisions on the resets, on repeat vs change, and on sending SysEx to the port only are well argued. I agree with them.

**Verdict: not approved.** Items 1–3 are blocking.

ryangavin:
Review round 1 addressed in 0b25514.

**Blocking, all fixed and tested**
- **Effect part assignments not remapped.** In the SInt effect SysEx, Insertion Effect Part (`4C 03 nn 0C pp`) and Variation Part (`4C 02 01 5B pp`) now move to part pp's destination channel through the Main A rules. A part Main A never routes gets `7F` (off). I set it off rather than dropping the message, so no insertion stays on whatever part the previous style assigned. CountryTwoStep: 0E, 0F become 0D, 0C. JazzWaltzFast and Bebop: 0B becomes 7F. Test: `sim::mixer::effect_part_assignments_follow_the_routing` (synthetic CASM plus the corpus styles).
- **Mid-bar entry.** After `seek`, `Engine::chase` now sends the new section's skipped controller, bank, program and bend events, in order, right after the SInt. The skipped notes stay skipped. The `PKind` Cc/Pc/Bend handling is factored into `emit_control`, so a chased CC7 still goes through the mixer rules. Tests: `mid_bar_fill_plays_its_own_voice` (synthetic) and `mid_bar_fill_voice_in_corpus` (ChartPop1 Fill A from beat 3: ch 15 is on 104/8/4). Both fail with the chase commented out.
- **Bank and program collapsed to "last value wins".** `ChannelInit` now stores the bank in effect at the last program change in `bank_msb/lsb`. A bank select after it goes in `pending_msb/lsb` and is sent after the program change. ChartPop1 ch 15 is 8/2/3 again (104 pending), and Let'sFunk ch 15 is 104/5/0. `sint_structures_every_corpus_style` now feeds a model receiver both the file and the structure and compares the resulting voice and bank registers for every channel, so a reordering like this one fails it. New tests: `sint_bank_after_program_is_pending` and `sint_voice_keeps_the_bank_of_its_program_change`. My count is 14 channels in 8 styles (you had 9 styles). My synthetic `sint_style` had the PC before its bank select, so its init now lists the bank first.

**Nonblocking**
- **Effect/drum SysEx on every section change: done.** `Prepared::init_parts` marks where the part setup ends. `reapply_init` sends only that part. Start and style load still send everything. `section_change_reapplies_init` now checks that Main B gets the XG part parameter and not the effect SysEx.
- **1116 vs 998: you were right.** I count 998 `4C 03` messages. The PR body is corrected.
- **`restore_untouched_levels` re-arming pickup: done as a guard, no behaviour change.** It now skips a part already at the style's level. `Takeover::software_moved` depends only on the last hardware position and the value. Any hardware report while picked up goes through `set_volume`, which marks the part as moved. So for an untouched part at an unchanged value, the call was idempotent and did not actually re-arm anything. No test was added, because none could fail before the change.
- **`live::malformed_key_bytes_never_panic`: agreed.** Its doc comment now says it guards the whole input path, and names `midi::data_bytes_are_always_below_0x80` as the #50 regression test. The PR body says the same.

**Decisions added** (also posted on #15):
- Decision: on a section change, re-send only the part setup, not the XG effect and drum setup SysEx. Patterns never send SysEx, so there is nothing to restore, and an XG receiver would cut reverb and delay tails at every fill.
- Decision: an effect assigned to an unrouted part is set off (`7F`) rather than dropped, because the source part never sounds, and dropping it would leave the previous style's assignment.
- Decision: a mid-bar entry chases the skipped non-note events, rather than re-applying only at bar-aligned entries, because a Fill's first-beat voice is part of the Fill.

Gates:
- `cargo test --release`: 213 passed, 0 failed, 1 ignored (the existing loudness probe), with the corpus linked and executed.
- `cargo clippy --release` shows no new warnings in the touched code.
- Golden digests are unchanged.

ryangavin:

I checked out `origin/m2/sint` (0b25514) with the corpus linked. I checked that the files the corpus tests name are there (MOX_v2, T5Style), so those tests really ran.
- `cargo test --release`: 213 passed, 0 failed, 1 ignored.
- `cargo clippy --release`: no new warnings. The two engine.rs warnings (lines 907 and 1335) were already there before this PR.

I wrote throwaway probes and did not commit them. The tree is clean.

### Round 1 fixes: checked
- **Effect part remap.** `4C 03 nn 0C pp` and `4C 02 01 5B pp` now go through `route[pp]`. An unrouted part gets `7F`. A `pp` of 16 or more (AD part, off) is left as it is. Correct.
- **Mid-bar chase.** `chase` sends only the Cc, Pc and Bend events before `ev_idx`. It uses the new section's rules, sends CC7 through `pattern_volume`, and doesn't allocate.
  - I checked whether chasing a bend could leave a part detuned. In 122 beat entries across 28 styles, a Fill or Break's last bend before the entry is off-centre and followed by a note. The pattern plays that note bent anyway, so the chase is faithful there. Not a bug.
- **Pending bank.** Fixed. The corpus test now models a receiver, which covers the reordering.
- **Effect SysEx out of the section change, pickup guard, PR body counts:** all done.

### Blocking

**1. A section change wipes every custom drum setup: the drum part's program change is sent again, but its Drum Setup SysEx is not** (`src/engine.rs` `reapply_init`, which takes only `init.iter().take(init_parts)`; the drum setup (`4C 3n`) sits in `sint.sysex`, after `init_parts`)
- Genos Data List, MIDI Data Format, Drum Setup table, Note: "When a part to which a Drum Setup is assigned receives a program change, the assigned Drum Setup will be initialized."
- `reapply_init` sends the PC of each drum part (Part Mode `07` = 2 or 3, i.e. Drum Setup 1 or 2) and its XG part parameters, but never the `4C 30/31` messages. On any receiver that follows that rule, the style's per-note drum pitch, level, pan and sends are set by Start and lost at the first section change, for the rest of the run.
  - So the Intro (or the first Main bar) plays one drum sound and everything afterwards plays another.
- Corpus: 170 of 208 styles have drum setup SysEx (2249 messages). In all 170, a routed part is in drum setup mode and has an SInt PC.
- Probe (AustinCityBlues.S930, Main A then Main B):
  - Start sends 2 drum setup messages.
  - At the Main B change, PCs go out on ch 9 and ch 10 (`[0xC8, ..]`, `[0xC9, ..]`), with 0 drum setup messages after them.
  - Expected: the drum setup follows the drum PCs again.
- This also contradicts the Decision "patterns never send SysEx (nothing to restore)". The PR's own PC re-send is what needs the restore.
  - The drum setup is not an effect type, so re-sending it cuts no reverb or delay tails. It is at most 39 messages (HoldOnAMinute).
- Fix, either:
  - split `sint.sysex` into effect SysEx (sent at Start and load only) and drum setup SysEx (`4C 3n`, re-sent after the part setup on every change); or
  - don't send the drum setup at all. The Drum Setup table's MIDI Reception column says "(Available only for Song parts)", which may mean Genos ignores it for style parts. That needs checking.
- Whichever you choose, Start and section changes must agree. Record it as a Decision.
- Add a test: after a section change, the last message that touches a drum-setup part's kit is its drum setup, not the PC.

### Nonblocking

- **The SInt is still routed by Main A only, and a section change now re-applies that routing on purpose** (`Prepared::new`, `route` = Main A rules).
  - In 48 cases across 5 styles, a section routes a *different* source channel (with its own SInt voice) to a destination and has no PC of its own before its first note. That destination then gets the voice of Main A's source, or no voice at all.
  - Examples:
    - ModernJazzBld.T157 Intro C and Ending C: src 13 and 14 go to dest 15. The SInt gives src 13 its own voice. Main A's src 15 has none.
    - Swingin'Boogie.T161 Main C and Main D: src 7, 8 and 9 go to dest 15. Main A doesn't route dest 15 at all.
  - This was already there before the PR and is not a regression. But `reapply_init` runs exactly where a per-section routing (send the SInt of `rules[src]` for the section being entered) would fix it.
  - Suggest a follow-up issue rather than more scope here.
- **Size of a section change.** It now sends every part's full setup, twice per fill (in and out). That is fine for the port. Keep an eye on it for DIN output once Section Change Timing = Immediate (#22) lands.
- **Tests that return early without the corpus.** `prep()`/`t5()` return early when the corpus is missing. That follows the existing convention, and with the corpus linked they ran.
- **Committed content.** The test comments quote a few bank and program numbers from corpus SInts (ChartPop1, Let'sFunk). No notes, so no leak, but keep it to that.
- **#50 splitter.** Correct.
  - A stray `F7` in a data position ends the message and is then delivered as a 1-byte `[F7]`, as before.
  - `F1`/`F2`/`F3` data bytes are skipped because running status is 0.
  - No consumer indexes on them.

**Verdict: not approved.** Item 1 is blocking.

ryangavin:
Review round 2, addressed in 5b453c7.

**Addressed**
- **Blocking: drum setup lost at the first section change.** A section change now re-sends the SInt's drum setup SysEx right after the part setup. That covers Drum Setup 1/2 parameters (`4C 30/31`) and Drum Setup Reset (`4C 00 00 7D`), classified by the new `sff::is_drum_setup`. `Prepared::init` is now ordered as part setup, then drum setup, then effect SysEx. `init_parts` is renamed to `init_resend` and covers the first two groups, so Start and section changes send the drum setup in the same place: after every program change and part mode. The effect SysEx is still sent only at Start and style load, so reverb and delay tails aren't cut.
  - Tests:
    - `section_change_resends_drum_setup_after_program_changes` (synthetic). It checks the order PC, part mode, Drum Setup Reset, drum setup parameter, and that no effect SysEx goes out at a change.
    - `corpus_drum_setup_survives_section_change` (AustinCityBlues, Main A→B→C→A). Every change sends the whole drum setup block after all the SInt program changes. Before this fix it failed at "drum setup re-sent".
  - The PR body's Decision is updated.
  - README is updated.
- **Nonblocking: per-section SInt routing.** This predates the PR and isn't a regression. I filed #64 with the review's counts and a proposal.

**Declined**
- **Full part setup sent twice per fill, and DIN output.** Nothing to change now. It fits the port. I'll revisit it with #22 (Section Change Timing = Immediate) if DIN output lands.
- **prep()/t5() tests return early without the corpus.** This is the existing convention. The symlinks were in place and the corpus tests ran (215 passed, 1 ignored as before).
- **Bank and program numbers in test comments.** No change. They are voice numbers only, with no notes, and I added none beyond the existing ones.

Gates:
- `cargo test --release`: 215 passed, 0 failed.
- `cargo clippy --release`: no new warnings in the code I touched.
- Golden digests are unchanged.

ryangavin:

I checked out `origin/m2/sint` (5b453c7) with the corpus linked.
- `cargo test --release`: 215 passed, 0 failed, 1 ignored. The corpus tests ran: the sff structuring test walked 208 styles.
- `cargo clippy --release`: no new warnings in the touched lines. The warnings at engine.rs:913/1341, sff.rs:140/464 and live.rs:288/393/589 are in code this PR doesn't change.
- My throwaway corpus probes are deleted, and the tree is clean.

### Round 2 blocking item (drum setup lost at a section change): fixed
- The Data List Drum Setup note is quoted correctly. It also says a program change applies to *every* part sharing that drum setup. Re-sending the whole block after all the PCs covers that.
- `is_drum_setup` matches `4C 30/31` and Drum Setup Reset `4C 00 00 7D nn`, device `1n` only.
- `init` is now ordered as part setup, then drum setup (`init_resend` ends here), then effects. Start and a section change therefore send the drum setup in the same place. Section repeats and stop still send nothing.
- `section_change_resends_drum_setup_after_program_changes` would fail without the fix: `o[2]` would be `usize::MAX` and `o[2] < o[3]` would fail.
- `corpus_drum_setup_survives_section_change` checks that the whole block comes after every SInt PC on three changes.
- No allocation or locking in `reapply_init`: it only iterates the precomputed `Msgs`.
- The mixer rules are unchanged by this commit.

### Corpus re-counts (probe over all 208 styles)
- 170 styles have drum setup SysEx. That matches the PR.
- **File order vs PR order.** In 169 of those 170, the file already puts the drum setup after the last PC of a drum-setup-mode part. The PR's reordering is therefore what authors intend.
  - The one exception is CoolRevibed. In file order, a strict XG receiver would wipe its drum setup with the later PC. The PR applies it.
  - That follows the recorded Decision ("drum setup needs the part mode in place first"). Fine.
- No SInt uses XG drum NRPNs (CC99 14–1F) on a drum-setup part. So moving the Drum Setup Reset after the part controllers can't wipe an NRPN drum edit.

### Blocking
None.

### Nonblocking
- **Pattern PCs on drum-setup parts.** Two styles (6-8ChartBallad, Gunslinger; 12 events) have a program change *inside a pattern* on a drum-setup-mode part.
  - That PC comes after `reapply_init` (at tick 0, or through `chase`). By the Data List rule it resets the drum setup again for the rest of that section.
  - It is faithful to the XG rule, and it is better than before: the next section change now restores the setup.
  - Whether Genos really re-initialises a style part's drum setup on a pattern PC can't be confirmed from the manuals, which mark drum setup reception "Available only for Song parts". Worth one playtest line in #15. No change asked.
- **Test detail.** `assert_eq!(p.init_resend, 8)` in the synthetic test restates the implementation. The order and count checks next to it are the behavioural part. Harmless.

**Verdict: approve-worthy.** Nothing blocking. (Posted as a comment only; I am not approving or merging.)

ryangavin:
Round 3 review: there were no blocking items. Pushed c115e85.

**Addressed**
- The synthetic drum-setup test no longer asserts `p.init_resend == 8`, which only restated the implementation. The behavioural checks stay: the order is PC, then part mode, then drum setup; the drum setup is sent exactly once; and no effect SysEx is sent at a section change.
- Added a playtest note on #15 about in-pattern program changes on drum-setup-mode parts.

**Declined**
- CoolRevibed puts its drum setup before the drum part's last PC in file order. I kept the reordering. The recorded Decision sends each part as bank, program, then its parameters, with drum SysEx last. The other 169 of 170 drum-setup styles already use that order. In file order, a strict XG receiver would wipe the setup the author clearly meant to apply.

**Gates**
- `cargo build --release` succeeds.
- `cargo test --release`: 215 passed, 0 failed. The corpus tests ran.
- `cargo clippy --release` shows no new warnings.

PR #55, opened 2026-09-23
NTT 5th-variation tables (#11)

ryangavin:
## What

The Reference Manual (Style Creator > Channel parameters > NTT) defines each minor table by the degrees it moves. Melodic Minor moves the 3rd. Harmonic Minor moves the 3rd and 6th. Natural Minor moves the 3rd, 6th and 7th. Dorian moves the 3rd and 7th. Each one lowers those degrees by a semitone on a major-to-minor change, raises them on a minor-to-major change, and then says "Other notes are not changed". Each "5th" table is the same thing, plus it "transposes the perfect fifth note above the Source Root with augmented and diminished chord types".

Before this PR, all eight tables used the general scale model with the table's minor scale. That had two effects. Every table moved the 5th over aug and dim. And over chords that are not plain minor (dim, dim7, 7th and the other dominant chords), all four tables played the same generic scale, so they could not be told apart.

This PR adds `minor_table_map()` in `src/theory.rs`, which works as that switch:

- A chord counts as minor when its 3rd is minor: m, m6, m7, m7b5, dim, dim7, mM7 and the minor tension chords. Every other chord that has a 3rd counts as major.
- Going from major to minor, the table lowers the major 3rd and the major 6th and 7th that it names. Going from minor to major, it raises the ones it names. Every other note keeps its interval above the root, and that includes chromatic notes.
- The "5th" tables also move the source's perfect 5th to the chord's #5 or b5 over aug and dim chords. The base tables keep the perfect 5th.
- A target chord with no 3rd (sus4, 7sus4, 1+8, 1+5, 1+2+5) still goes through the scale model, as it did before.

Example: a C D E F G A B source played over C chords.

| Table | Cdim | Cdim, "5th" table | C7 |
|---|---|---|---|
| Melodic Minor | C D Eb F G A B | C D Eb F F# A B | unchanged |
| Harmonic Minor | C D Eb F G Ab B | C D Eb F F# Ab B | unchanged |
| Natural Minor | C D Eb F G Ab Bb | C D Eb F F# Ab Bb | unchanged |
| Dorian | C D Eb F G A Bb | C D Eb F F# A Bb | unchanged |

Before this PR, every "5th" table gave C D Eb F F# Ab A over Cdim. Every base table sent B to A over dim and to Bb over C7.

Source side:
- A C7 source over Cm gives C Eb G Bb in every table. Before, Melodic Minor 5th raised the Bb to B.
[...]

PR #60, opened 2026-09-23
Guitar NTR: All-Purpose, Stroke, Arpeggio (#12)

ryangavin:
## What

This PR implements the three Guitar NTR tables: All Purpose, Stroke and Arpeggio. It rests on one finding from the full corpus: **a Guitar source is string-coded, not a line of pitches.**

- **Decoding the source.**
  - A key's pitch class picks a string of a guitar voicing of the played chord: B = string 1, A = 2, G = 3, F = 4, E = 5, D = 6, with flats and sharps going to the nearest of these. C is the bass and C♯ is the fifth above it.
  - The key's octave picks the neck position.
  - Keys from 96 up are MegaVoice noise keys. They pass through untouched.
- **Where it lives.** The model is `theory::guitar` / `theory::voicing`, written up in `docs/genos-features.md` under "Guitar NTR voicing model".
- **Oracle.** The Guitar identity check is replaced with a Guitar check (`oracle::GuitarCheck`, documented in `docs/oracle.md`).
- **Engine, on top of #63's revoice** (in `src/engine.rs`). #63 brought two-phase revoicing, muted twin voices and Pitch Shift bends, which replace this PR's earlier `revoice` changes. On top of that:
  - A Guitar string on a key that another string of the part struck within a 32nd note joins it as a muted twin, so each pitch is struck once per strum.
  - A late chord brings in the strings the previous chord muted.
  - Guitar noise keys are never moved by a chord change, never vote on the part's bend, and go out uncompensated on a bent part.
- **Chord table.** 7♯9's 3rd is its major 3rd, in the shared table.

Rebased onto `integration/m2-chord-conversion` at 31cc61a, after #66, #62 and #63. It is one commit.

## Why

The earlier draft of this PR modelled Guitar sources as CM7 fingerings played as pitches. That was based on a `.sty`-only scan: 43 rules in 35 styles, all of them MOX_v2. The full corpus, walked with `library::corpus_styles` over every file type, contradicts it:

| | Earlier claim | Full corpus |
|---|---|---|
| CASM records with a Guitar zone | 43 in 35 styles | **382 in 140 styles**: 26 All Purpose, 236 Stroke and 117 Arpeggio in the middle zone, plus 3 with a Guitar outer zone |
| Arpeggio rules | 0 | **117** |
| Source chord | all C / M7 | **42 are not CM7**: 32 Cm11, 3 Cm7, 3 AM7, 2 F♯M7, 1 C6/9, 1 C |
| Bass On | 0 of 43 | **158 of 382** set it in at least one zone |
| Guitar notes | 2,039 | **97,156**, of which **20,887** are noise keys (96 and up) |
| Strums spanning more than one tick | 254 of 345 | **14,919 of 19,742** groups of 2+ notes within 30 ticks |

The strums are string codes:

- **Stacked seconds.** Of the strums of three or more strings, 13,372 of 14,153 are stacked seconds (adjacent scale steps), one note per string. No guitar plays those as pitches.
  - T5Style: 10,633 of 11,092.
  - SX900: 2,739 of 2,787.
  - MOX_v2: 0 of 274.
- **String counts, not chord roles.** In Stroke's playable range, each of the four upper-string codes appears 10–12k times, about equally. The bass and lower-string codes appear 0.4–3.6k times. A fingering would favour chord tones. These counts favour strings.
- **Source chord.** The 42 non-CM7 rules use the same code alphabet as the CM7 ones. That confirms RM p.29: Source Root/Chord are ignored for Guitar.

Read as pitches, most corpus strums would sound as tone clusters. Read as codes, they are guitar voicings.

## How the model works

- **Voicing.** The bass is the root, or the slash bass with Bass On. It goes on the lowest string that reaches it within the position. Each string above takes the lowest fret of a chord tone not yet sounding, important tones first (3rd, 5th and 7th; for five-note chords the 3rd, 7th and tension). The open position gives:
  - x32000 for CM7, x32010 for C, 320003 for G, xx0232 for D, x02210 for Am, 320001 for G7;
  - C D B♭ C E for C9;
  - a C7♯9 that keeps its major 3rd.
- **Position.** Keys up to 59 play at frets 0–4, keys 60–71 at frets 5–9, and keys from 72 at frets 10–14.
- **Range.** Nothing sounds below the open low E (MIDI 40). Note Limit folds what sounds, and never decides which strings sound.
- **Tables:**
  - **All Purpose:** every string sounds. A string below the bass plays the 5th or the root within reach (the alternate bass of 332010), else doubles the string above it.
  - **Stroke:** strings below the bass, and strings with no chord tone within reach (1+8), are muted, as RM p.30 says ("some notes may sound as if they are muted").
  - **Arpeggio:** the bass stays on string 6 or 5. Strings 1–4 take the fingering within reach that rises from the bass and, with the bass, sounds the most chord tones. At most 5⁴ fingerings are tried, on the stack.
- **Bass On** is read per zone (#57).
- **Pitch-written sources (MOX_v2)** are read as codes too:
  - An x32000 strum becomes a subset of the target's guitar voicing: chord tones only, and always the root.
  - Over CM7 it plays C E G. Over G it plays G B G G D.
  - It keeps the 3rd over 286 of the 348 chords that have one. When the bass sits on string 5, the E written on that string doubles the bass.
- **Retrigger Rule.** Guitar zones keep the RTR their author chose and go through #63's revoice.
  - The Yamaha-authored parts are written for Pitch Shift: 374 of the 375 Guitar zones in T5Style and SX900 (the other is Retrigger to Root). All 43 MOX_v2 Guitar zones are Retrigger.
  - Under Pitch Shift, the part bends by the shift most of its ringing strings need, and strings that need another shift are retriggered.
- **Real time.** Each note is a pure function of (key, zone, chord). The voicing is a fixed `[Option<u8>; 6]`, with no allocation and no locks. Out-of-range roots and types are reduced mod 12 or clamped.

## How tested

- **Unit tests in `theory::tests`:**
  - `guitar_string_codes_play_guitar_voicings`: the standard open shapes and the octave-as-position rule.
  - `guitar_source_chord_is_ignored`: six declared source chords, every key, every target.
  - `guitar_noise_keys_pass_through`.
  - `guitar_plays_chord_tones_on_the_neck`: every key below 96, every chord and slash, Bass On and Off; only Stroke may mute.
  - `guitar_arpeggio_strings_carry_four_voices`: 332 of 336 four-note chords get all four tones, against 248 for All Purpose.
  - `guitar_7sharp9_keeps_the_third`, `guitar_note_limit_folds_and_never_mutes`, `guitar_bass_on_moves_the_bass_string` (including the per-zone case), `guitar_pitch_written_strums_stay_consonant`.
- **Corpus test `corpus_guitar_parts_play_chord_tones`.** This replaces `corpus_guitar_parts_play_as_written_over_cm7` and covers every file type through `library::corpus_styles`. For every Guitar channel and every chord it plays: noise keys come back untouched, every other output is a chord tone on the neck, and only Stroke mutes. It prints the counts in the tables above and asserts that notes exist whenever the corpus does.
- **Sim tests, rebuilt on #63's engine:**
  - `corpus_guitar_strikes_each_pitch_once`: no sounding guitar unison, and no strike on a key that is already sounding.
  - `corpus_late_chord_restores_guitar_strings`.
  - #63's `corpus_chord_changes_leave_no_blips_or_doubled_keys` and `corpus_full_performance_no_stuck_notes` replace this PR's earlier `corpus_stream_matches_sounding_notes`.
- **Noise keys are exempt** in `corpus_one_plus_eight_and_one_plus_five` and in #63's `corpus_pitch_shift_sounds_what_retrigger_would`. They are not pitches, and on a bent part they sound with the bend.
- **Gates:**
  - `cargo test --release`: **303 passed, 0 failed, 1 ignored** (pre-existing). The corpus ran: 208 styles and 7,537 channel rules; 140 Guitar styles and 97,156 Guitar notes.
  - `cargo clippy --release --all-targets`: 18 warnings, against 19 on the base (31cc61a). None is in the new code.

## Oracle (re-pinned)

- **Unchanged:** every pair and table score, and the Root Trans and Root Fixed identity lines.
- **Removed:**
  - `identity Guitar 96638 78539 78539` and `identity Guitar moved-in-limit 18099`. Pitch identity has no meaning for string codes, and no chord plays a Guitar source back as written.
  - The corpus identity total falls from 528,566 to 431,928 notes.
  - Per style: 140 styles' `identity` lines and 74 styles' `moved-in-limit` lines drop their Guitar notes.
- **Added:** the Guitar check. Every Guitar note is played on every chord its channel plays (12 roots × every CASM type, as the mutes allow).

| Table | notes | noise | noise moved | plays | muted | not a chord tone | out of range |
|---|---|---|---|---|---|---|---|
| All Purpose | 1,463 | 33 | 0 | 583,440 | 0 | 0 | 0 |
| Stroke | 70,770 | 15,426 | 0 | 22,506,408 | 719,196 | 0 | 0 |
| Arpeggio | 24,923 | 5,428 | 0 | 7,848,768 | 0 | 0 | 0 |

- **Note count.** The check counts 97,156 notes, where identity counted 96,638. Identity skipped the 518 notes of channels muted on their own declared source chord.
- **Same check on earlier models (not pinned):**
  - Model before #60: noise moved everywhere; 84, 3,780 and 27,996 non-chord tones for All Purpose, Stroke and Arpeggio.
  - This PR's earlier pitch model: noise moved everywhere; 121,428 and 7,749,312 non-chord tones for All Purpose and Stroke, 0 for Arpeggio.

## Golden changes (regenerated with `UPDATE_GOLDEN=1`)

- **ch12 Chord1 in the two golden styles with a Guitar part:** BluesOrganTrio bars 3–40 and JackDoesItAgain bars 2–40, the Main A–D span.
  - Both are MOX_v2 styles written as pitches.
  - Their strums now play the string-code reading described above: a subset of each chord's guitar voicing, chord tones only.
  - Intro and Ending are unchanged.
- **The 7♯9 Chord-table fix:**
  - G7♯9 in bar 39 of JackDoesItAgain (ch13) and TickingAway (ch12, ch13).
  - In TickingAway ch13, notes held into bar 39 from bar 37 now ring on, so their bar 37 lengths change, and they continue into bar 40 (C/G).
- **Not in this PR any more:** the earlier JackDoesItAgain ch13 `revoice` changes. #63 fixed the same behaviour, and its digests already include it.
- No other part or style moved. The oracle pin didn't move with #63.

## Decisions

- **Decision: a Guitar source is string-coded, because** of the corpus counts above.
  - 13,372 of 14,153 strums of 3+ strings are stacked seconds, one per string: 13,372 of 13,879 in T5Style and SX900, 0 of 274 in MOX_v2.
  - In Stroke's playable range the four upper-string codes each appear 10–12k times, far more than the bass and lower-string codes.
  - The code table (B..D = strings 1..6, C the bass, C♯ the fifth) is the one this repo used before this PR. Every playable pitch class the corpus uses maps to it.
- **Decision: the key's octave picks the neck position, because** the corpus writes each strum's codes inside one octave, and in at least two octave blocks. Frets 0–4, 5–9 and 10–14 keep the output near where the codes were written. This comes from the corpus, not a manual, and is flagged for a hardware capture (#9).
- **Decision: keys from 96 up pass through untouched, because** they are MegaVoice noise keys, not pitches: 20,887 of the 97,156 Guitar notes, spread over every pitch class at high velocity. Transposing them would pick a different noise.
- **Decision: Source Root/Chord are ignored for Guitar, because** RM p.29 says so, and the 42 non-CM7 rules use the same codes as the CM7 ones. The oracle agrees: no Guitar identity against a declared chord.
- **Decision: each note is converted on its own, because** 14,919 of 19,742 corpus strums span more than one tick. A per-note mapping gives the same voicing whatever the timing.
- **Decision: All Purpose sounds every string, Stroke mutes below the bass and where no chord tone is in reach, and Arpeggio keeps the bass on string 6 or 5 with four chord voices on strings 1–4, because** those are the three behaviours RM p.30 names: "works both for strumming and arpeggio", "some notes may sound as if they are muted", and "four-note arpeggio sounds".
- **Decision: pitch-written sources (MOX_v2, 43 rules) are read as codes too, because** the hardware has one Guitar rule, not two. They stay consonant: chord tones only, always the root, the 3rd over 286 of 348 chords. They are thinner than the original fingering (x32000 over CM7 loses its B).
- **Decision: the oracle's Guitar check tests validity, not identity, because** no manual page or corpus data says which voicing the Genos picks. It checks that noise keys are untouched and that every output is a chord tone on the neck and inside Note Limit, over every chord each channel plays. It does not grade string assignment, which would test the model against itself.
- **Decision: with Bass On, the voicing is built over the slash bass on the lowest string (032010 for C/E, 332010 for C/G), because** RM p.30 says only "the bottom note as Bass inside the Guitar voicings" follows slash chords. Bass On is read per zone (#57). 158 of 382 Guitar records set it in some zone.
- **Decision: the neck (low E), not Note Limit Low, decides what Stroke mutes, because** RM p.31 says of Note Limit that "every note is transposed automatically to fit into the range".
- **Decision: strings on one key are struck once per strum, with the later string kept as a muted twin (#63's mechanism, widened from one instant to a 32nd note for Guitar), because** one MIDI channel cannot sound a unison, and a strum spreads its strings over several ticks. With the twin, the key sounds until both strings end, so the earlier round's shortened-unison caveat is gone.
- **Decision: Guitar parts keep their authored RTR, Pitch Shift bends included, because** the authors of string-coded parts chose Pitch Shift for 374 of 375 Guitar zones. The MOX_v2 pitch-written parts chose Retrigger (43 of 43). Retriggering every guitar string instead would ignore the style data.
- **Decision: noise keys never move, don't vote on the bend, and go out uncompensated on a bent part, because** a noise key is a sample, not a pitch: a compensated key would pick a different noise. The cost is that a noise struck while the part is bent sounds bent.
- **Decision: 7♯9's 3rd is its major 3rd, because** the Data List spells it 1+♯2+3+(5)+♭7. This is in the shared Chord table.

Closes #12


ryangavin:

I checked out `origin/m2/guitar-ntr` (9adcabd) with the corpus symlinked.

- `cargo test --release`: 207 passed. The corpus tests ran.
- `cargo clippy --release`: no new warnings in the guitar path.

I used throwaway tests (not committed) to probe the code. **Verdict: changes requested.** Three findings are blocking.

### Blocking

**1. The 7♯9 voicing drops the major 3rd and doubles E♭ (`src/theory.rs:733`)**

Guitar All Purpose, Note Limit 40–127, target C7♯9 (type 27):

| Source shape | Output | Missing |
|---|---|---|
| x32000: C2 E2 G2 B2 E3 | C2 E♭2 B♭2 **E♭3 E♭3** | E, G |
| x35453: C2 G2 B2 E3 G3 | C2 B♭2 **E♭3 E♭3** B♭3 | E, G |
| Upper triad: G2 B2 E3 | B♭2 **E♭3 E♭3** | C, E, G |

- Stroke gives the same result.
- The E in the source is a 3rd, but the `chord_map` role mapping sends it to the ♯9 (D♯), not to the target's 3rd. Every C7♯9 strum comes out as a C–E♭–B♭ minor-7 sound.
- The PR body says "Five-note targets keep the 3rd, 7th and tension". For 7♯9 they don't.
- The PR body lists 7♯9 among the Main C/D targets whose golden changed, so the golden now locks this voicing in.
- `guitar_open_cm7_follows_as_guitar_shapes` tests C9 but not 7♯9. Please add a test that every target's strum contains the chord's 3rd and 7th.

**2. One strum sends the same pitch twice, which retriggers and cuts notes**

The PR says the old code "gave zero-length retriggered unisons" and that this is fixed. The new code still produces them, and two tests lock them in:

- `theory.rs:1484` expects `["G1", "G1", "D2", "G2", "G2"]` over G1+5.
- `theory.rs:1521` expects `["x", "G1", "G1", "E2"]` for Stroke with Bass On over C/G.

Other cases over C:

- x32000 over 1+8: C2 C2 C3 C3 C3.
- x32000 over 1+5: C2 C2 G2 C3 C3.
- The upper triad over 1+8: C3 C3 C3.
- Every 7♯9 case in finding 1.

Why this matters in the engine: `Engine::note_on` (`engine.rs:1085`) steals an identical sounding note. Each duplicate therefore sends note-off then note-on on the same key a few ticks apart. You hear a flam or retrigger, and the earlier string is cut to near zero length.

Corpus count: I grouped each channel's note-ons that fall within 30 ticks of each other, and left out groups whose source already has a repeated key. 309 of the 345 groups produce a duplicate output pitch under at least one chord that the channel's CASM mute lets it play.

| Chord type | Group × chord cases |
|---|---|
| 1+8 | 3708 |
| 1+5 | 3252 |
| 7♯9 | 2652 |

Two acceptable fixes:

- Make the mapping collision-free, for example move a string that lands on a pitch already taken to the next chord tone up.
- Or state explicitly that power-chord unisons are intended, and drop the "fixed" claim.

As it stands, the tests enshrine the artefact the PR says it removed.

**3. Stroke treats Note Limit Low as the bottom of the neck and mutes strings instead of folding them (`theory.rs:745`, `theory.rs:755`)**

`lo = z.lo.max(40)` feeds both the Stroke mute threshold and the fold. Note Limit is documented as a range that notes are octave-shifted into. It does not mute. Results for the x32000 strum under Stroke with a raised Note Limit Low:

| Note Limit Low | Target | Output |
|---|---|---|
| 52 | **CM7, the source chord** | x x x x E3: one string of five |
| 52 | C | x x x C3 E3 |
| 48 | B♭ | x x x B♭2 D3 |
| 48 | G | x x x G2 B2 |

- This breaks the PR's own invariant that the source plays as written over CM7 on every table.
- The invariant only holds while Note Limit Low is at or below C2 (48).
- The corpus only has Low = 0 or 40, so the goldens and the corpus test can't see this. Any other style with a guitar Note Limit above C2 will go mostly silent.
- Suggested fix: compute the mute threshold from `LOW_E`, or from the unfolded neck position, and let `fold_into` apply the Note Limit afterwards, as every other rule does.

### Nonblocking

1. **Two Decisions have no corpus evidence.** I reran the counts. 43 Guitar rules, 19 All Purpose, 24 Stroke, and Note Limit Low = 40 on 15 rules (all Stroke) all match the PR. But there are **0 Arpeggio rules and 0 Bass On rules** in the corpus.
   - The Arpeggio snap and the Bass On slash placement are therefore pure invention. The PR body should say so, so the owner knows to playtest them.
   - Arpeggio also breaks the "source plays as written over CM7" rule the other tables follow: D2 over CM7 comes out as C2. Please add that to the Decision too.
2. **Stroke mutes low source strings even over the source chord.** For a six-string 332000 C shape over CM7, the G1 is dropped. There are no corpus instances: all 10 source notes below C2 are G1 on All Purpose channels, and no Stroke channel has one. So the `key < 48 → None` branch at `theory.rs:1552` never runs against the corpus.
3. **`corpus_guitar_parts_play_as_written_over_cm7` checks very little.**
   - It passes with `notes == 0` when the corpus is missing. The repo convention is to skip, but please assert `notes > 0` when `corpus/` exists.
   - It only checks that notes come back unchanged over CM7, so no transposition bug could fail it. A corpus sweep that checks chord-tone coverage and the absence of duplicates over every chord a channel is allowed to play would catch findings 1 and 2.
4. **The strum-spread evidence is understated.** With the same 30-tick window I get 254 of 345 groups (74%) spanning more than one tick, not "about half". This supports the per-note Decision more strongly than the PR says.
5. **Merge-conflict risk.** The PR edits the shared `out_of_range_types_do_not_panic` test and the 1+8 bullet in `docs/genos-features.md`. Both are likely to conflict with sibling M2 PRs. This is minor.

### Verified OK

- **Manual quotes:**
  - RM p.29, "The settings here are not applied when NTR is set to Guitar", refers to the Source Root/Chord page. That supports ignoring Source Root/Chord.
  - The RM p.30 NTT Bass wording about "only the bottom note" is quoted accurately.
  - The Stroke and Arpeggio descriptions match the manual.
- **Real-time safety:** `guitar()` makes no allocation. `scale()` returns `[u8; 7]`. The indexes are taken `% 12` and the outputs are clamped, and out-of-range roots and types don't panic.
- **Mixer principle:** nothing on the mixer or gain path was touched.
- **Golden changes:** exactly 77 lines changed, all on ch12 Chord1 (38 in BluesOrganTrio, 39 in JackDoesItAgain), which matches the PR body. The files are digests only, and no readable note content is committed.
- **Scope:** the code change is confined to the guitar path. `transpose_group` still skips the Root Fixed permutation for Guitar, which is correct.

ryangavin:
Review round 1 is addressed in 49a68de. The PR body has the updated model, tests, golden notes and Decisions.

**Blocking: all three fixed**

1. **7♯9 dropped the major 3rd.** The root cause was in the shared Chord table. `five_note_roles` took the first 3rd it found, and for 7♯9 (1+♯2+3+(5)+♭7, Data List) that is the ♯2. The major 3rd is now the 3rd and the ♯2 is the ♯9 tension.
   - x32000 over C7♯9 plays C2 E2 B♭2 E♭3 E3. x35453 plays C2 B♭2 E♭3 E3 B♭3.
   - Tests: `guitar_7sharp9_keeps_the_third` and `chord_table_7sharp9_keeps_the_major_third`.
   - The corpus test now sweeps all 345 corpus strums that hold the source's 3rd, over every chord the channel plays, and asserts a 3rd remains. With the old roles it fails on C7♯9.
   - Because the fix is in the shared table, Chord-NTT parts over 7♯9 changed in the goldens too. The PR body lists them.
2. **Duplicate pitches within one strum.** A per-note mapping has no strum context, so a single note can't know that another string already covers its pitch. 1+8 has only one pitch class for up to six strings. The fix is therefore in the engine:
   - `note_on` skips a Guitar note whose pitch another string of the same part, struck within a 32nd note, is ringing.
   - `revoice` never moves a guitar string onto the pitch another string ends on (one that stays, or a lower-indexed one that moves there). That string stops instead.
   - A later strum still re-strikes the pitch, as a guitarist would.
   - No allocation: the check scans the fixed `sounding` array, plus a 256-entry stack array in `revoice`.
   - The two tests you named now show both layers. The per-note mapping still reads `[G1,G1,D2,G2,G2]`, and a `struck()` helper that mirrors the engine gives `[G1,D2,G2]`.
   - New corpus sim test `sim::corpus_guitar_strikes_each_pitch_once`. All 6 styles with a Guitar part play through 1+8, 1+5, 7♯9, slash and plain chords, with changes at half bars and just after the downbeat. It asserts zero same-strum re-strikes and zero doubled strings. Disabling the `note_on` check fails it with 13 re-strikes in BluesOrganTrio, and disabling the `revoice` check fails it with 6 doubled strings in GirlsFolk.
3. **Stroke used Note Limit Low as the bottom of the neck.** The mute threshold now uses the neck (low E) and the chord's bass only, and the Note Limit folds whatever is strummed. That matches RM p.31: "every note is transposed automatically to fit into the range".
   - Test: `guitar_stroke_ignores_note_limit_for_muting`. At Low 48 and 52, Stroke equals All Purpose over CM7, C, B♭, G and Am7, with no string muted.

**Nonblocking**

- **Addressed: no corpus evidence for Arpeggio and Bass On.** The PR body, the Decisions and `docs/genos-features.md` now say that 0 of 43 rules use Arpeggio and 0 set Bass On. They also say that Arpeggio changes the source over CM7 (D2 plays C2).
- **Addressed: corpus test passing with zero notes.** `corpus_guitar_parts_play_as_written_over_cm7` now asserts notes > 0 when `corpus/` exists, and includes the 3rd-coverage sweep over every allowed chord. The duplicate sweep is the engine-level sim test above. The per-note mapping still folds strings by design, so duplicates are checked where they are struck.
- **Addressed: strum-spread evidence.** I reproduced your 254 of 345 (74%). The PR body, the doc and the code comment say "most", with the numbers.
- **Declined: Stroke dropping a low source string (G1 of 332000) over CM7.** In this model, leaving out strings below the bass is what Stroke means. I checked the alternative, exempting source strings below C2: after the shape slides, the chord-tone strings of a CM7 shape at or above C2 land at or above the target's bass. Exempting would therefore make Stroke effectively identical to All Purpose. No Stroke note in the corpus is below C2. The limit is now stated as a Decision, and `guitar_source_chord_plays_as_written` keeps asserting it.
- **Declined: the edits to `out_of_range_types_do_not_panic` and the 1+8 bullet.**
  - The test edit is required, because the test called `guitar_voicing`, which this PR removes.
  - The 1+8 bullet was factually wrong after this PR (there is no "fifth string" any more).
  - I checked the open sibling PRs (#55, #56, #57, #58, #59, #61). None of them touches that test or that bullet, so no conflict is pending.

**Gates:** `cargo build --release` passes. `cargo test --release` passes: 211 passed and 1 ignored (existing), and the corpus ran with 2039 guitar notes. `cargo clippy --release` shows no new warnings. The goldens were regenerated, and the changed bar groups are explained in the PR body.

ryangavin:

I checked out `origin/m2/guitar-ntr` (49a68de) with the corpus symlinked.

- `cargo test --release`: 211 passed, 1 ignored. The corpus tests ran: `corpus_guitar_parts_play_as_written_over_cm7` reports 2039 guitar notes, 345 strums and 254 spread over more than one tick.
- `cargo clippy --release`: no new warnings. The two warnings near touched code (`engine.rs:1306`, `theory.rs:465`) are on lines that were already there.

I probed the code with throwaway tests, none of them committed. **Verdict: nothing blocking.** All three round-1 blockers are fixed.

### Round-1 fixes verified

1. **7♯9 keeps its major 3rd.**
   - The Data List line is `1+#2+3+(5)+b7`, as quoted.
   - The `five_note_roles` change only matters for a type that holds both 3 and 4, and 7♯9 is the only one. No other 5-note type moves.
   - `guitar_7sharp9_keeps_the_third` and `chord_table_7sharp9_keeps_the_major_third` assert the right voicings.
   - The corpus 3rd sweep is a real behavioural test. It covers every chord a channel plays that has a 3rd.
2. **One strike per pitch.**
   - I disabled the `revoice` filter (`engine.rs:1291`). `corpus_guitar_strikes_each_pitch_once` then fails with "GirlsFolk: two guitar strings on one pitch", as the PR says. The test does guard the code.
   - `doubles` resolves collisions correctly even when groups are processed out of index order: the lowest-index string always survives.
3. **Stroke mutes by the neck, not the Note Limit.** The threshold is now `LOW_E` plus the bass pitch class, and `fold_into` runs afterwards. `guitar_stroke_ignores_note_limit_for_muting` covers Low 48 and 52.

**Other checks:**

- **Golden changes:** the 8 changed digest lines are exactly the ones the body lists: bar 17 and bar 39 in BluesOrganTrio and JackDoesItAgain, and bars 39–40 in TickingAway. The files are digests only.
- **Committed content:** no readable note content in the diff. That covers the docs, PLAN.md and the digests.
- **Corpus counts:** I reran 345 strums, 254 of them spread, and 2039 notes. They match.
- **Real-time safety:**
  - `note_on` and `revoice` use only fixed arrays: `fin` is `[Option<u8>; 256]` on the stack. There are no locks.
  - `is_guitar` uses `.get()`. The indexing in `revoice` repeats indexing the base loop already does.
  - The `ppq / 8` window is a float-to-u64 cast and cannot panic.
- **Mixer:** untouched. No gain path changed.

### Blocking

None.

### Nonblocking

**1. `revoice` can still cut a guitar string. This bug predates the PR; the PR makes it less frequent.**

- **Cause:** when string A moves onto the pitch that string B is leaving, the loop sends A's note-on before B's note-off. B's note-off then silences A on the synth, while the engine still thinks A is sounding.
- **Trace:** BluesOrganTrio ch12, A major held into bar 2, then C arrives 25 ms after the downbeat:
  ```
  +25.000ms  8B 49 0   9B 52 76   <- E string moves C#2 -> E2
             8B 52 0              <- G string leaves E2: kills the E string's E2
             9B 55 96 ...
  ```
  Result: the strum over C sounds without its E.
- **Scale:** I ran a plain-chord script at half-bar changes (C Am C F G Em7 Am7 Dm7 G7 C) and counted note-ons followed by a same-key note-off at the same instant on the 6 all-guitar channels.
  - 49a68de: 9 cuts.
  - Base f962dc1: 17 cuts.
  - Non-guitar parts show the same pattern (1770 hits on this branch), so the cause is general to `revoice`, not the guitar code.
- **Why the test misses it:** `corpus_guitar_strikes_each_pitch_once` checks the engine's `sounding` bookkeeping (`guitar_unisons`), not the MIDI stream. The PR body's "no two strings ever sound the same pitch" is true of the bookkeeping but not of what the synth hears.
- **Suggestion:** a follow-up issue. `revoice` should send every note-off of a pass before any note-on, and the test should check the recorded stream. It is out of scope for #12.

**2. A late chord doesn't bring back strings left out under the previous chord.**

- **Scenario:** I held 1+8, 1+5 or A just before the downbeat, then played C 25 ms late. The sounding set at +30 ms is missing strings compared with C held throughout. Example: BluesOrganTrio `[48,52,55,60,64]` drops to `[48,55]` after C1+8.
- **Cause:** the strike-once skip, and the Stroke mute, leave no `sounding` entry for `revoice` to re-pitch. `catch_up` only restarts parts that were fully silent.
- **Compared with base:** base loses the same strings through the old steal (1558 vs 1739 diff cases across my 4×corpus sweep; most of the extra are the A→C cut from item 1). This is not a regression, but a late-chord catch-up for skipped strings is worth a follow-up.

**3. A skipped string can sound shorter than written.**

- The skipped string's pitch lasts only as long as the string that is ringing it.
- **Corpus count:** over C1+8, C1+5, G7♯9, C and Am7, the ringing string ends before the skipped one in 41,278 of 105,400 folded pairs, on average 119 ticks early.
- This is short (under a 16th at 1920 ppq). It is fine for now, but worth a line in the Decision.

**4. `unison_steals`:** it is counted for every part but asserted only for guitar channels. That is fine. It could be restricted to guitar channels to match its doc comment.

No other scope creep. The shared-table 7♯9 fix is justified and its golden effects are listed.

ryangavin:
Review round 2 is addressed in 8d3639d.

**Addressed**

- **`revoice` cuts (nonblocking 1):** `revoice` now defers its note-ons and sends them after every note-off of the chord change, so a string moving onto a pitch another string is leaving is no longer cut. While doing this I found a second bug that also predates the PR. `revoice` restored start times inside its loop, so the rest of a group was regrouped and moved again when the loop reached its next member: one JackDoesItAgain Chord2 note went 69 → 62 → 65 → 71 → 62 in a single pass. With note-ons deferred, that bug would have left notes on. Each note now moves at most once per pass (a fixed `done` array on the stack; nothing allocates). Tests:
  - `corpus_guitar_strikes_each_pitch_once` now checks the MIDI stream the synth receives: no Guitar pitch is struck while the channel already sounds it. It fails with the note-ons sent inline (GirlsFolk ch12).
  - New `corpus_stream_matches_sounding_notes`: every corpus style plays the golden script. After every wake-up, the stream's open notes must equal the engine's sounding notes, and no note-off may go to a pitch that is off. It fails if a note can move twice (JackDoesItAgain ch13).
  - Golden: JackDoesItAgain ch13 Chord2, bars 13, 18 and 23, lose the zero-length notes from the repeated moves. The PR body explains each one.
- **Late chord doesn't restore strings (nonblocking 2):** `catch_up` now brings in Guitar notes from the late window that aren't sounding, meaning unisons that were skipped and strings Stroke muted, unless another string of the part already rings their pitch. New test `corpus_late_chord_restores_guitar_strings`: in every Guitar corpus style, a chord 20 ms late must leave that strum's strings exactly as the same chord on the beat does. The previous chords are C1+8, G1+5, D and A7. Without the fix it fails (GirlsFolk, C after C1+8, missing the E). Strings from earlier strums are not part of this. They follow their retrigger rule in both cases, and the catch-up can't bring back a note whose note-on is older than the window.
- **Skipped unison length (nonblocking 3):** the one-strike Decision and `docs/genos-features.md` now say that a left-out string lasts only as long as the string ringing its pitch, and they quote your corpus measurement.
- **`unison_steals` (nonblocking 4):** removed. The problem was worse than the counter covering too many parts: `note_on` returns before counting a Guitar note, so it could never count a guitar steal, and the guitar assertion always passed. The MIDI-stream check above replaces it.

**Declined**

- Nothing from this round. One related issue is still open and is out of scope: two notes from different sources of a non-guitar part can still land on the same pitch during `revoice`. `note_on` steals the older note in that case, but `revoice` has no such rule. The notes stay balanced, but the synth hears one voice. This predates the PR and affects every Chord part, so it belongs in its own ticket.

Gates: `cargo build --release` passes. `cargo test --release`: 213 passed, 1 ignored (existing), and the corpus ran. `cargo clippy --release`: no new warnings.

ryangavin:
## Rebased onto `integration/m2-chord-conversion` (a1daaf5, after #57)

This is squashed to one commit, `2238b98`. **Do not merge yet: the Arpeggio design is going back for a fix (see below).**

- **`src/theory.rs` tests.** This was a textual conflict only. I kept #55's `minor_tables_*` tests and this PR's `chord_table_7sharp9_keeps_the_major_third`.
- **Semantic conflict with #57.** `guitar()` read `rule.bass_on`, which #57 removed; Bass On is now stored per zone. It now reads the Bass On of the note's own zone (`z.bass_on`), which is what this PR's Bass On decision describes. The unused `rule` argument is gone. `guitar_bass_on_moves_only_the_bottom_root` sets Bass On on every zone and adds a per-zone case: with Mid Low 50, a C2 follows the slash only when the **low** zone has Bass On.
- **Golden digests.** This PR's regenerated digests still pass on the new base, so nothing was regenerated.

### Gates

- `cargo test --release`: **255 passed, 0 failed, 1 ignored**. The corpus was symlinked and ran.
- `cargo clippy --release --all-targets`: no new warnings. There are two fewer than on the #57 tip.

## Oracle delta (this PR only, against the a1daaf5 pin)

**Only the Guitar identity check moved.** No scored pair is a Guitar channel, so the table, change and NTR rows are unchanged.

| | Before | After |
|---|---|---|
| identity Guitar, exact | 10.4% | **81.3%** |
| identity Guitar, pitch class | 18.1% | **81.3%** |
| Guitar notes moved inside Note Limit | 86,606 | **18,099** |

- **Improvements:** 140 styles have Guitar zones. Every MOX_v2 guitar style now reaches or nearly reaches 100%. Most T5/SX900 styles rose by 5 to 60 points.
- **One drop:** `T5Style/WorshipFast.T154.sst` lost 211 pitch-class notes, with exact unchanged. It is in group 2 below.
- **Pin and docs:** `tests/oracle/scores.txt` is re-pinned. docs/oracle.md now has a Pin history entry.

### What the 18,099 remaining misses are

Every one changes pitch class. None is an octave or a mute.

1. **10,690 notes: Arpeggio sources recorded over CM7.** `ARP_SNAP` moves every note to its nearest CM7 tone, even on CM7 itself. These sources are about 15% F, 21% A, and some D, Db and Eb. So **49.5% of all CM7-sourced Arpeggio notes (10,690 of 21,579) do not play back as recorded on their own chord.** If CM7 is the source chord, that contradicts RM p.28. I think this is a real fidelity problem.
2. **7,409 notes (Arpeggio 2,584, Stroke 4,825): rules whose CASM Source Root/Chord is not C M7.** These are C, Cm7, Cm11, C6/9, F♯M7 and AM7. The oracle plays them on that chord, and this PR reads every Guitar source as CM7 (RM p.29). The pitch content of these sources looks like the CM7-declared ones, not like their declared chords. For example, the "Cm11" Stroke sources are heavy on F, G, A and B, with almost no Eb or Bb. That supports RM p.29. So this group is most likely the oracle's choice of chord, not a transposer miss. The oracle is unchanged; this is noted for a follow-up.

## Concerns (why this goes back)

**1. This PR's corpus evidence covers only about 35 of the 208 styles.** `corpus_guitar_parts_play_as_written_over_cm7`, like the older `corpus_every_rule_every_chord`, only walks `.sty` files. It skips the `.prs` and `.sst` files that make up the T5Style and SX900 libraries. On the full corpus (all extensions, as `Library::scan` and `sim::corpus()` read it):

| Claim in this PR | Full corpus |
|---|---|
| 43 Guitar rules in 35 styles | **379** Guitar CASM rule records in **140** styles |
| 0 Arpeggio rules | **117** Arpeggio rule records (plus 236 Stroke, 26 All Purpose) |
| All Guitar rules are C / M7 | 337 C/M7; **42 are not** (32 Cm11, 3 Cm7, 3 AM7, 2 F♯M7, 1 C6/9, 1 C) |

**2. The Arpeggio decision was taken on "no corpus evidence", and the evidence exists.** The snap-to-CM7 rule changes half of the real Arpeggio notes on their own source chord. If the corpus test walked every extension, it would fail on exactly these notes: it asserts that every non-Stroke note plays unchanged over CM7. I did not widen the test here, because that reopens the Arpeggio design, which is going back to the author.

**3. Worth re-checking against the full corpus:**
- the Stroke claim that no note lies below C2;
- the strum statistics (345 strums, 74% spread over more than one tick);
- the "0 of 43 set Bass On" count.

**Suggested fix direction:** Arpeggio should leave notes as recorded on the source chord, the same as All Purpose does over CM7. After that, widen the corpus test to all style extensions. The oracle's `identity Guitar` line should then drop by about 10,690 moved notes.

ryangavin:
## Reworked around the string encoding: 7daadb2

This is rebased onto `integration/m2-chord-conversion` at cc20044 (after #66 and #62) and squashed to one commit. **Do not merge yet.** The owner should playtest it first.

### What the full corpus showed

The earlier model read Guitar sources as pitches fingered over CM7. It was based on a `.sty`-only scan: 43 rules, all MOX_v2. `library::corpus_styles` walks every file type and finds 382 Guitar records in 140 styles, with 97,156 notes.

- **Stacked seconds.** 13,372 of the 14,153 strums of three or more strings are stacked seconds, one note per string. That is 10,633 of 11,092 in T5Style, 2,739 of 2,787 in SX900, and 0 of 274 in MOX_v2.
- **Per-string counts.** In Stroke's playable range the four upper-string codes each appear 10–12k times, far more than the bass and lower-string codes. That is a count per string, not a chord fingering.
- **Noise keys.** 20,887 notes are MegaVoice noise keys, at 96 and above.
- **Non-CM7 source chords.** The 42 rules declared over another chord use the same codes, which confirms RM p.29.

So these are string codes. Playing them "as written" over CM7, the fix first proposed, would have sounded the clusters.

### What changed

- **`theory::guitar` decodes the strings.** B..D are strings 1..6, C is the bass and C♯ the fifth above it. The octave picks the neck position. Keys from 96 up pass through untouched.
  - A voicing of the played chord is built per note. It is pure, allocation-free, and at most 5⁴ fingerings for Arpeggio, on the stack.
  - The source chord is ignored. Bass On is read per zone.
- **The three tables:**
  - All Purpose sounds every string.
  - Stroke mutes strings below the bass and strings with no chord tone within reach.
  - Arpeggio keeps the bass on string 6 or 5 and puts four chord voices on strings 1–4. That works for 332 of 336 four-note chords, against 248 for All Purpose.
- **MOX_v2 sources written as pitches** become subsets of the target's voicing: chord tones only, always the root, and the 3rd over 286 of 348 chords. Over CM7 an x32000 strum plays C E G.
- **Oracle.** Guitar notes leave the identity check. The new Guitar check plays every note over every chord its channel plays. For all three tables it finds 0 noise keys moved, 0 non-chord tones and 0 notes out of range. Stroke mutes 719,196 of 22,506,408 plays.
  - The Root Trans and Root Fixed lines are unchanged, and so is every pair.
  - `docs/oracle.md` explains the check and has before/after figures for the pre-#60 model and the earlier pitch model.
- **Tests:**
  - The corpus Guitar test covers every file type.
  - The 1+8 sim test exempts Guitar noise keys.
  - The new unit tests are listed in the PR body.
- **Goldens:** only ch12 Chord1 of BluesOrganTrio (bars 3–40) and JackDoesItAgain (bars 2–40) changed. Both are MOX_v2 styles written as pitches.
- **Docs:** `docs/genos-features.md` has the new model and the corpus counts. PLAN.md is updated. The PR body's Decisions section is rewritten with the corrected counts.

### Gates

- `cargo test --release`: 289 passed, 0 failed, 1 ignored (pre-existing). The corpus ran: 208 styles, and 140 of them have Guitar parts.
- `cargo clippy --release --all-targets`: 18 warnings, against 20 on the base. None is in the new code.

### Open for playtest or capture (#9)

- The octave-as-position reading and the exact voicings are our choices; the corpus doesn't settle them.
- Whether All Purpose doubles, or Stroke mutes, exactly the strings the Genos does is also unverified.

ryangavin:
## Rebased onto #63: 5146906

This is rebased onto `integration/m2-chord-conversion` at 31cc61a (#63) and is still one commit. **Do not merge yet.**

### Conflicts

- **Engine.** #63's two-phase `revoice_part` already covers this PR's earlier `revoice` work: note-offs first, each note moved once, and two voices on one key keep one. So I took #63's engine and re-applied only the Guitar pieces:
  - **Strike once per strum.** `note_on`'s "same instant" test for muted twins now counts, for a Guitar note, any string of the part that struck the key within a 32nd note. The later string joins as a muted twin instead of being dropped, so the key rings until both strings end. The earlier round's caveat about shortened unisons is gone.
  - **Late-chord catch-up.** A late chord now brings in only Guitar strings that have no voice, meaning strings the previous chord muted. Unisons are twins and `revoice` re-parts them.
  - **Test helpers.** `guitar_unisons` counts only unmuted strings. `guitar_strings` reports the sounding pitch, bend included.
- **Sim.** `corpus_guitar_strikes_each_pitch_once` and `corpus_late_chord_restores_guitar_strings` are back. #63's stream and no-blip tests replace this PR's earlier `corpus_stream_matches_sounding_notes`.
- **Goldens.** Regenerated on the new base:
  - Guitar ch12 in BluesOrganTrio (bars 3–40) and JackDoesItAgain (bars 2–40): the string-code voicings.
  - The 7♯9 fix at bar 39, in JackDoesItAgain ch13 and TickingAway ch12/13. In TickingAway ch13, notes held from bar 37 now ring through bar 39, which changes their lengths in bar 37 and carries them into bar 40.
  - The earlier JackDoesItAgain ch13 `revoice` changes are no longer in this PR, because #63 already made them.
- **Oracle.** The pin is unchanged by #63.

### Should guitars use Pitch Shift bends?

Yes, as the style data says. I checked the RTR of every Guitar zone in the corpus:

| Folder | Guitar zones | RTR |
|---|---|---|
| T5Style and SX900 (string-coded, Yamaha-authored) | 375 | 374 Pitch Shift, 1 Retrigger to Root |
| MOX_v2 (written as pitches) | 43 | 43 Retrigger |

So Guitar zones keep their authored RTR through #63's bend path. The part bends by the shift most of its ringing strings need, and strings that need a different shift are retriggered.

One fix was needed: **noise keys (96 and up) are samples, not pitches.**
- A chord change never moves them, and they have no say in the part's bend.
- A noise key struck while the part is bent goes out as written. Compensating it would pick a different noise, so it sounds bent instead.
- `corpus_pitch_shift_sounds_what_retrigger_would` and the 1+8 test exempt these keys on Guitar parts.

### Gates on the combined tree

- `cargo test --release`: 303 passed, 0 failed, 1 ignored (pre-existing). The corpus ran: 208 styles, 7,537 channel rules, and 140 styles with 97,156 Guitar notes.
- `cargo clippy --release --all-targets`: 18 warnings, against 19 on 31cc61a. None is in the new code.

ryangavin:
## Adversarial review of 5146906, plus fixes in 50da293

I reviewed `origin/m2/guitar-ntr` at 5146906 with the corpus symlinked, then pushed one commit on top: **50da293**. It is a normal push, not a force-push.

**Verdict: approve.** The one blocking issue was a voicing bug that dropped the 3rd. It is fixed in 50da293, with tests. Nothing else blocks. The owner should know about the decisions listed at the end.

### Gates on 50da293

- `cargo test --release`: **305 passed, 0 failed, 1 ignored** (pre-existing). That is 303 plus the 2 new tests.
- The corpus ran: 208 styles, 140 of them with Guitar parts, 382 Guitar rules, and 97,156 Guitar notes, 20,887 of them noise keys.
- The RTR tests all pass, including `corpus_chaos_leaves_no_note_detuned` and `corpus_pitch_shift_sounds_what_retrigger_would`.
- `cargo clippy --release --all-targets`: 18 warnings, the same as 5146906. None is new.

### Blocking, now fixed: strums dropped the 3rd

For each string, `voicing` took the lowest fret of *any* important tone. So a tension on a lower fret pushed the 3rd out. I probed a full open-position strum (C D E F G A B codes) over all 12 roots and every type. **17 chords per table lost the 3rd** on All Purpose and Stroke:
- D9 played xx0210: D A C E, with no F♯.
- Also A9, Am9, Dm9, D6/9, Dadd9, Dm(add9), and 10 more.

Arpeggio lost it over Gm9, B♭m9 and 4 more. The goldens showed it too: D9 in bar 26 and Cm(add9) in bar 37 of BluesOrganTrio and JackDoesItAgain had no 3rd. Cm(add9) sounded as C G D.

**Fix:**
- A string that reaches the 3rd (the sus tone of a sus chord) takes it before the other important tones.
- Arpeggio scores the 3rd above the tone count.

**What changed and what didn't:**
- The standard shapes are unchanged: CM7 x32000, C x32010, G 320003, D xx0232, Am x02210 and G7 320001.
- C9 is now x32333, D9 xx0212 and C7♯9 x32344.
- Arpeggio still sounds all four tones for 332 of 336 four-note chords. All Purpose goes from 248 to 254.
- x32000 keeps the 3rd over 308 of 348 chords, up from 286.
- The docs are updated.

**Test:** the new `guitar_strums_keep_the_third` covers every table, all three positions, every chord, and Bass On slash chords. It fails on the old rule, first at Cadd9 in the C4 position.

**Goldens:** only Guitar ch12 changes, in the two MOX styles:
- bar 26: D9 gains F♯;
- bar 37: Cm(add9) gains E♭;
- bar 39: G7♯9's 3rd and ♯9 swap octaves.

The oracle pin is unchanged.

### Guitar with Pitch Shift bends: I accept the decision, with one cost

The corpus RTR counts on Guitar zones match yours:
- 374 Pitch Shift and 1 Retrigger to Root, all Yamaha-authored;
- 43 Retrigger, all MOX_v2.

The bent strings are not detuned. I mutated the engine to send guitar strings without bend compensation. `corpus_pitch_shift_sounds_what_retrigger_would` failed, which shows the strings disagreeing with the bend really are retriggered at the right pitch.

I measured with a throwaway script over all 140 Guitar styles: a half-bar progression through Main A–D with 12 chord types, run as authored and again with Pitch Shift converted to Retrigger.

| | Pitch Shift (as authored) | All Retrigger |
|---|---|---|
| Bend changes on Guitar parts | 2,617 | 0 |
| Strings carried through a bend with no new attack | 3,015 | 0 |
| Re-attacks at chord-change instants | 7,589 | 9,885 |
| Noise hits sounding on a bent part | **2,460 of 19,086 (12.9%)**, about 2 semitones on average | 0 |
| Held noise voices bent mid-note | 311 | 0 |

On the Genos, Pitch Shift moves each note with no attack. Bending the majority shift and retriggering the rest is closer to that than retriggering everything, which would re-attack about 2,300 more strings as an unstrummed block.

**The cost:** the bend persists until the part falls silent. That rarely happens with continuous strumming, so about 1 in 8 noise hits sounds transposed, about 2 semitones on average. A noise key can't be both at its own key and correctly pitched on a bent channel, so the only alternative is never bending Guitar parts. That is the owner's call below.

### Other checks

**String model and the manual:**
- RM p.29–31 only says "approximate voicings played with natural guitar fingering". Source Root/Chord is ignored, Bass On moves "only the bottom note", and Note Limit folds.
- The code model is consistent with all of that. It is a corpus inference, not something the manual states, and the PR says so.

**Stroke thinness:**
- On the upper four string codes, which T5/SX900 use most, all 4 strings sound for 384 of 396 chords (1+8 excluded). The other 12 are dim triads, where 3 strings sound.
- The thinnest full strums are 6/9 chords, with 3 distinct pitches, and 1+8 with 2 (C2 and C3), which is by design.
- Not too thin.

**Edge cases:**
- Slash chords with Bass On give 032010 (C/E), 332010 (C/G) and D/F♯ with F♯ on string 6. With Bass Off the slash is ignored.
- 1+5, 1+8, dim, dim7 and aug all give chord tones only.
- Cancel never reaches `guitar()`, because `effective_chord` filters it. Chord types and roots out of range don't panic.
- The 7♯9 decision (major 3rd, ♯9 as the tension) is fine.

**Real time:**
- No allocation or locks. `arpeggio_voices` tries at most 625 fingerings per note, on the stack.
- `is_guitar` does `.get()` lookups.

**Committed content:** the goldens are digests only, and `scores.txt` is numbers only. The docs name model voicings, not style patterns.

**Does the oracle Guitar check catch regressions?** Partly. These are mutations to `theory::guitar`:
- A string giving a non-chord tone: caught (`not-chord-tone` 0 → 4.6M on Stroke).
- Noise keys moved: caught (`noise-moved` 0 → 15,426).
- **Every string playing the bass: not caught.** The oracle pin still passed. The unit, sim and golden tests caught it, 12 failures.
- **The dropped-3rd bug above: not caught.** The oracle pin is identical before and after the fix.
- This is the "validity, not identity" decision working as designed. The unit tests are what protect voicing quality.

**Engine guards, by mutation:**
- **Noise-key guards.** Removing `revoice`'s noise-key `Leave` failed only the golden digest. Sending noise keys compensated on a bent part failed nothing.
  - 50da293 adds `corpus_guitar_noise_keys_are_never_moved`. It fails on both mutations and confirms it runs the case: 48,936 noise voices observed on bent parts.
- **Strike-once rule.** Removing the per-strum twin rule fails `corpus_late_chord_restores_guitar_strings` and the golden, but not `corpus_guitar_strikes_each_pitch_once`.
  - With the rule off, `note_on` steals the key (note-off, then note-on), so neither the bookkeeping nor the "struck while sounding" stream check sees it.
  - I tried a stream check for a re-strike within one strum. It can't tell a steal from the pattern's own note-off and re-strike on the same tick (HeySixteen ch12 does this), so I didn't commit it. The rule is still guarded, by the other two tests.
- **Chaos test.** `corpus_chaos_leaves_no_note_detuned` does not catch guitar strings sent without compensation. Its model checks the receiver's bend against the engine's shift, not the key. The Pitch-Shift-vs-Retrigger test covers that case.

**Cleanup in 50da293:** `off_where`'s doc comment had slid onto `is_guitar`. It is back on `off_where`.

### Decisions for the owner

1. **Guitar parts keep their Pitch Shift bends.** About 13% of noise hits sound transposed by the part's bend, about 2 semitones on average. The alternative is to retrigger Guitar parts and never bend them. That keeps noise exact, at the cost of about 2,300 more re-attacks in the probe script. I would keep the current choice, but it is worth a listen on a noisy T5 strum style.
2. **Master Transpose shifts guitar noise keys.** `Engine::master` runs after `theory::guitar`, so under Master Transpose a noise key goes out as another noise. This is untested and not in the PR's scope. It is worth a follow-up, and a check on hardware of how the Genos transposes MegaVoice noise.
3. **The reading is still a model.** The octave-as-position rule, the exact voicings, and which strings All Purpose doubles or Stroke mutes are still unverified against a Genos (#9).

Not merged.

Issue #34, opened 2026-09-23
Controllers: pedals, joystick, sustain, assignable functions


ryangavin:
_Details in the spec._

---
Milestone: **M6 · Parts, voices & live play** · Spec: [docs/genos-features.md](../blob/main/docs/genos-features.md) · Backlog: [docs/backlog.md](../blob/main/docs/backlog.md)

ryangavin:
Implemented in #93 (docs/controllers.md). These are the decisions where the manuals don't settle the behaviour:

- **Decision: the sustain pedal reaches Left by default**, and each part can be taken out. The RM says Sustain affects "all notes played on the keyboard" (p.139), and the parts are a per-function setting. The panel SUSTAIN button (Right 1-3) and LEFT HOLD are separate features.
- **Decision: a part must be on to be sustained or bent.** A part switched off is released and centred at once; one switched on joins the held pedal and the bend. A switched-off part must never ring on, or come back with a stuck pedal.
- **Decision: the default pedals are the GM pedals** (CC 64 Sustain, 66 Sostenuto, 67 Soft), not the Genos jack defaults (ART.1, Volume). yahaha has no Super Articulation, and a Volume pedal would be a hidden gain, which breaks the mixer principle.
- **Decision: modulation defaults to Right 1-3 and pitch bend to all four parts**, as the Genos joystick does (OM p.70). The Pitch Bend Range defaults to 2 semitones (0-12, RM p.140).
- **Decision: Fill Up at Main D / Fill Down at Main A** play that Main's own fill and stay there. While stopped, Fill Up/Down select the next Main. OTS +/− wrap round.
- **Decision: Registration +/−** are in the function table but unavailable until Registration Memory (#36) lands.
- **Decision: after Panic, or a keyboard unplugged with its pedal down, the pedal counts as up** until it is pressed again.

ryangavin:
Review round 1 on #93 added these decisions:

- **Decision: re-picking a pedal's function (or its CC, by Learn) releases what it drove**, unless another pedal still holds the same switch. A latched Toggle sustain, a held pedal or a Pitch Bend pedal mid-sweep would otherwise stay stuck with nothing left to turn it off. Changing only Control Type, polarity or Range keeps the switch.
- **Decision: the registration functions are Registration Bank +/−** (REGIST BANK [+]/[−], pedal-assignable in the RM list). Registration Sequence +/− is not pedal-assignable on the Genos, so it is not in the table. Both stay unavailable until #36.
[...]

PR #61, opened 2026-09-23
Keyboard parts model: Right 1–3 + Left, Panel/Style fader pages (#31)

ryangavin:
## What and why

Owner feedback: "the sounds my right hand plays aren't on the faders, and that's not how arrangers work." That's right. On the Genos, the Mixer's **Panel** tab (and the sliders' Balance assign type) balances the keyboard parts **Right 1, Right 2, Right 3 and Left**. The **Style** tab balances the 8 style channels. One Touch Settings load voice, on/off, volume and octave into those four parts (OM p.48 and p.90). yahaha had eight generic right-hand "voice slots" instead, with layering on the fader buttons. This PR replaces them with the Genos keyboard-parts model.

**Parts** (new `src/parts.rs`):
- There are four parts: Right 1, Right 2, Right 3 and Left.
- Each part has a voice, a volume (its CC7), an octave shift and on/off.
- The Right parts that are on sound together on the right of the split. That is layering. Left sounds left of the split, which is still the existing split point.
- With Left off, the Right parts play over the entire keyboard (OM p.48). The exception is Lower chord detection outside the Full Keyboard types: there the keys left of the split only give the chord (OM p.56, ACMP on and LEFT off).
- The state is plain atomics in `Shared`, so it exists with or without the built-in synth.

**Channels:**
- Right 1 = ch 1 and Left = ch 2, the same as before, so the owner's Ableton set keeps working. Right 2 = ch 3 and Right 3 = ch 4.
- The input thread sends each note to every part that is on, on that part's channel, shifted by that part's octave. It remembers what each key sounded, so the note-off goes to the same place even if the parts changed while the key was held.
- The synth's player instance plays exactly what the port gets, one synth channel per part. It no longer has its own layering or octave logic.
- Each part's volume goes out as CC7 on its channel from the engine thread, to the port and the synth alike. Under Manual Bass, Left plays the Style's Bass voice at Left's own volume (Panel fader 4).

**Launchkey faders** now have two pages, like the Genos Mixer tabs:
- **Panel:** faders 1–4 = Right 1, Right 2, Right 3, Left. Their buttons turn the part on/off; Shift + button selects that part for voice editing. Faders and buttons 5–8 are dark and unused.
- **Style:** faders 1–8 = the Style parts, as today. Their buttons mute the parts.
- The button under the master fader toggles the page (`F9` in the terminal). It lights blue on Panel and green on Style. The master fader is always master.
- Soft takeover applies whenever a page switch or an OTS recall moves a value under a fader:
  - Physical fader positions are tracked per fader across both pages.
  - Switching to Panel rebinds the part pickups.
  - Switching to Style records the positions in `Parts` and raises a flag. The engine thread takes it on its next wake, and before applying any Style fader move, and calls `Engine::faders_at`. It rebuilds each `Takeover` with the new `Takeover::at`. That is the only engine change, and it is limited to mixer state. A flag, rather than a ring command, means a full ring can't lose the rebind.

**Voice selection:**
- Pick the part to edit with `F1`–`F4`, the page-3 EDIT pads, or Shift + fader button.
- `9`/`0` and the page-3 VOICE −/+ pads step that part's voice.
- `5` `6` `7` `8` turn the parts on/off; `l` still toggles Left.

**OTS recall** fills Right 1–3 and Left: voice, on/off, CC7 and octave. The Panel faders pick up the new volumes with soft takeover.

**Launchkey pad page 3:**
- Top row: OTS 1–4, OTS Link, (unused), Voice −, Voice +.
- Bottom row: Right 1/2/3/Left on/off, then Edit R1/R2/R3/L.

**TUI:** the mixer area shows a Panel block and a Style block side by side. The block on the active fader page is outlined in yellow, and the selected part is marked with ▶. The old "voices / LAYER" status line is gone.

**README** is updated: parts, channels, both fader pages, page 3, buttons, keys and the Ableton channel table.

## How tested

- `cargo build --release`: clean. No new warnings.
- `cargo test --release`: 209 passed, 1 ignored (the existing loudness probe). The corpus and soundfont tests ran: the `corpus` and `soundfonts` symlinks were in place, and no test printed a "skipping" line.
- Golden digests are unchanged. This PR doesn't touch the style engine's output.
- `cargo clippy --release`: no new warnings in the code I touched. The warnings left in `ui.rs`, `live.rs` and `engine.rs` were there before.
- New tests:
  - `parts::tests`: channel map, OTS filling all four parts (SlowWalker OTS 1 → R1+R2+Left), Manual Bass voice on Left and the Left LEDs, voice keys editing the selected part, and Panel fader soft takeover across an OTS recall and page switches.
  - `live::tests::right_parts_layer_and_left_follows_its_switch`: layering with a different octave per part, Right 1 off leaves ch 1 silent, Left gated by on/off or Manual Bass, and Right parts left of the split when Left is off.
  - `live::tests::left_off_right_parts_cover_the_keyboard_unless_chord_section`: through `Input::key_msg` with Left off, key 40 plays Right 1 + Right 2 in Upper, Full Keyboard and AI Full Keyboard, and is silent in Lower Fingered / Fingered On Bass. The chord is still read from the left hand.
  - `live::tests::keyboard_volume_and_voice_messages_are_dropped`: the keyboard's CC7, CC0, CC32 and Program Change reach no part; the mod wheel reaches all four.
  - `live::tests::style_rebind_survives_a_full_ring`: with the command ring full at the page switch, the next Style fader move doesn't jump its part.
  - `live::tests::layered_notes_release_where_they_started`: note-offs follow the note-ons after parts change; pedals reach all four parts.
  - `live::tests::fader_pages_route_faders_and_buttons`: Panel faders and buttons, Shift+button select, the page toggle handing the engine the physical positions, Style faders and buttons reaching the engine, no jump when coming back to Panel.
  - `sim::mixer::style_faders_rebind_after_a_page_switch`: the engine side of the page-switch takeover.
  - `synth::parts_tests::keyboard_parts_play_on_their_own_channels`: renders each part through the SoundFont on its own channel; the band synth never hears it; CC7 mutes only its own part.
  - `launchkey::tests::fader_buttons_follow_the_page`, and updated page-3 LED/action tests.
- I rendered the front panel with `yahaha screen` to check the new mixer layout at 150 columns.

## Decisions

- **Decision: no "whole Style" volume fader on the Panel page, because** the Genos Panel tab has one but it scales all 8 style channels. Under the CC7 principle, that would need a hidden per-part multiplier or would overwrite the eight parts' CC7s. The Style page covers the style parts one by one, and master stays the only non-CC gain. Panel faders 5–8 are left dark for now.
- **Decision: the port carries exactly what sounds, so a part that is off sends nothing, Left included, because** that is the Genos model (a part that is off is silent) and it keeps port and synth identical.
  - With Left off, the keys left of the split play the Right parts (on their channels) in Upper detection and in the Full Keyboard types (OM p.48). In Lower detection with the other fingerings they only give the chord, and nothing goes out for them (OM p.56).
  - Consequence: ch 2 carries the left hand only while Left is on (off by default, as on the Genos).
  - To hear the left hand on ch 2 in Ableton, turn Left on (`l`, `8`, the page-3 pad, Panel fader button 4, or an OTS). Under Manual Bass it always sounds.
  - Owner: overrule this if the Ableton set needs ch 2 unconditionally.
- **Decision: each part's octave shift is applied to the notes it sends on the port, because** what the port carries should be what the part plays. With layered parts in different octaves, the Ableton tracks then match the built-in synth.
- **Decision: voices stay synth-only (no Program Change on the port), because** before this PR the port never carried the right/left voice either. An Ableton set picks its own instruments, and a Program Change could hit whatever is mapped on those tracks.
- **Decision: pedals, wheels and pressure from the keyboard go to all four parts, on or off, because** a sustain release must always reach a note that is still ringing, and the Reference Manual's Assignable functions table says Sustain affects "all notes played on the keyboard". The Genos has a per-part setting for this; per-part controller routing belongs to #34.
- **Decision: the keyboard's own CC7, bank select (CC0/CC32) and Program Change are dropped, because** a part's volume is its CC7 and its voice is its program. Passing them through would change what sounds behind the parts' backs, and the screen and the last-sent CC7 cache would show the wrong value.
- **Decision: under Manual Bass, Left plays the Style's Bass voice at Left's own volume (Panel fader 4), not the Style's Bass fader, because** OM p.51 says the Bass voice "is assigned to the left part". The Style's Bass part is muted meanwhile, and Panel fader 4 stays live. The Left on/off LEDs show Left lit while Manual Bass sounds it, the same as the screen.
- **Decision: under Manual Bass, Left's octave shift is not applied to the bass, because** OM p.51 assigns only the Bass *voice* to the left part, and Left's octave belongs to Left's own voice. Genos/Tyros OTS in the corpus set Left to octave +1 in 654 of 696 settings and +2 in 40 (review count), for pads and strings. Applied to the bass, any OTS recall would put the bass line one or two octaves up. The octave comes back when Manual Bass is off.
- **Decision: under Manual Bass, Left's on/off switch is refused (Panel button 4, the Left pad, `8`, `l`), because** the left hand sounds the bass whatever the switch says. A flip would change nothing you can hear or see and would only surprise you later when Manual Bass goes off. The screen says to turn Manual Bass off (`D`) to switch Left.
- **Decision: the fader page starts on Panel, because** the complaint is that your own parts weren't on the faders, and Panel is the Genos Mixer's first tab. One press of the master-fader button gives the old Style layout.
- **Decision: at start only Right 1 is on (Grand Piano); Right 2 = Strings, Right 3 = Brass, Left = Strings, all at volume 100, because** Right 1 alone is the old default (slot 1, piano). Strings and Brass were slots 4 and 5, the usual layers.
- **Decision: pad page 3's style-part mute pads become keyboard-part on/off and Edit pads, because** the style mutes now live on the Style fader page buttons (and `z`…`,`). The Genos front panel has PART ON/OFF and PART SELECT buttons for exactly these four parts.
- **Decision: OTS and Left pads work without the synth, because** the parts now drive the port too (routing, octave, CC7), not just the built-in synth.

## Not in this PR

Issue #31 also covers three split points and Left Hold. Per the scope for this ticket, Left still uses the single existing split, and Left Hold is not implemented. The PR only references #31, so merging it leaves the issue open for those two; `docs/backlog.md` lists them as still open.

Part of #31 (split points and Left Hold are still open, see above).


ryangavin:

**Setup:** I checked out `origin/m2/keyboard-parts` (8f6e165) detached, with the `corpus` and `soundfonts` symlinks in place.

**Gates:**
- `cargo test --release`: 206 passed, 1 ignored. No test printed "skipping", so the corpus and soundfont tests ran.
- Golden digests: unchanged.
- `cargo clippy --release`: 20 warnings, down from 26 on the integration base. None of them is new.

I wrote throwaway tests in `live.rs` to prove the findings below, then removed them. Nothing was pushed.

**What holds up:**
- The part/channel map is right, and each part's CC7 goes out unchanged on its own channel.
- The synth has no per-part gain left. The player synth now plays exactly what the port gets.
- The `Sounded` bookkeeping releases every layered note where it started, even after on/off, octave or transpose changed.
- `Takeover::at` and `faders_at` are the right minimal engine change.
- The paths I checked are real-time safe. `sounds()`, `fader_button`, `set_fader_page`, `sync_part_volumes` and `sync_player` use only fixed arrays and atomics, with no allocations and no locks.
- I found no readable style content. The OTS test asserts voice numbers only, as the old test did.
- The "no whole-Style fader" Decision is sound under the CC7 rule.

### Blocking

**1. Fidelity: with Left off, left-of-split keys are silent in every mode. The Genos plays the Right parts there, except in the ACMP chord section.**
- Where: `src/live.rs:348` (`let left = k <= split`) and `sounds()` at `src/live.rs:212`.
- Behaviour: every key at or below the split goes to the Left side only, so with Left off it sounds nothing.
- The manual:
  - OM p.48: *"To play one single Voice over the entire keyboard: Turn on the RIGHT 1, 2 or 3 part. Make sure the LEFT part is turned off."*
  - OM p.56 draws the left section as chord-only just for ACMP on + Lower detection, so silence is correct there.
  - Upper detection (OM p.51): the chord section is right of the split. The left side is the LEFT part section, so with LEFT off the Right parts cover it.
  - Full Keyboard / AI Full Keyboard (RM "Fingering Type" table; OM p.46): "anywhere on the keyboard using both hands—like conventional playing of a piano".
- Proven through `Input::key_msg`:
  - Upper mode, Left off, key 40 on: output `[]`.
  - `Fingering::FullKeyboard`, Lower, Left off, key 40 on: output `[]`.
  - Expected in both cases: `[0x90, 40, v]` on the Right parts that are on (ch 1 by default).
- Regression: before this PR the port carried these notes on ch 2, so an Ableton set playing Full Keyboard or Upper now loses the left hand altogether.
- Your Decision says "a part that is off is silent". That's true, but it treats "Left off" as "left hand silent", and the OM contradicts that outside the Lower chord section.
- Suggested fix: in `key_msg`, send a left-of-split key to the Right parts when Left is not sounding, unless it is in the Lower (non-Full-Keyboard) chord section. Keep the `route`/chord logic as is. Add a test for each of the three cases.

### Nonblocking

**2. Keyboard CC7, bank select and Program Change get past the parts' state, and the mixer display lies afterwards.**
- Where: `src/live.rs:388`.
- Behaviour: every CC and PC from the keyboard is copied to ch 1, 3 and 4, and on to the port and the player synth.
- Proven:
  - A keyboard `B0 07 14` gives `[B0 07 14] [B2 07 14] [B3 07 14]`, while `parts.volume[R1]` still reads 100.
  - `sync_part_volumes`' `last` cache never sends the 100 again. The TUI and the Panel fader show 100 while the parts sound at 20.
  - A keyboard `C0 28` re-voices Right 1–3 in the synth, while `parts.program` (and the screen) still say Grand Piano.
- Why it matters: that breaks "a part's volume is its CC7". Ch 1 already had this before the PR; this PR spreads it to three parts.
- Suggestion: drop CC7, CC0, CC32 and PC from the keyboard, or fold them into `Parts`.

**3. Manual Bass: Panel fader 4 edits a volume nobody hears, and the Left LEDs disagree with the screen.**
- Where: `parts.rs:175` and `parts.rs:146`.
- Fader: under Manual Bass, `hw_fader(LEFT, …)` returns true and sets `volume[LEFT]=20`, but `channel_volume(LEFT)` stays 90 (the Style's Bass fader). The fader looks dead, and no `↕` is shown.
- LEDs: pad page 3's Left pad (`launchkey.rs:563`) and the Panel fader button use `on_mask()` (`ui.rs:72`), so they show Left **off**. The TUI uses `left_sounds()`, so it shows Left **on**.
- Suggestion: either route fader 4 to the Bass fader under Manual Bass, or show it as inactive. Use `left_sounds()` for the LEDs.

**4. The sustain Decision needs a source.**
- The RM's "Assignable Functions" table says Sustain gives "all notes played on the keyboard" a longer sustain, and lets you choose which parts it applies to.
- Leaving Left out is a guess. Either cite a default that excludes Left, or send CC64 to Left too.
- "Left never received them before" describes yahaha's old behaviour, not the Genos's.

**5. A dropped `Cmd::FadersAt` leaves the engine's takeover stale, and the next Style fader move jumps its part.**
- Where: `live.rs:492` (and `ui.rs` `toggle_fader_page`).
- If the command ring is full, the engine keeps `picked=true` with a `hw` from before the Panel page. The first Style-page move then applies straight away.
- This is unlikely with a ring of 16/64, but it's cheap to fix. For example, keep a "Style page entered" flag, or have `hw_fader` re-check against `parts.fader_hw` in the engine.

**6. Scope and bookkeeping.**
- The body says `Closes #31`, but the PR itself says three split points and Left Hold aren't done. Merging would close #31 with that scope still open. Use "Part of #31", or split the issue first.
- `docs/backlog.md:165` still describes "the current voice slots".

**7. Minor points.**
- Upper mode + Left off: the left hand now sends nothing on ch 2. It did before this PR. See item 1.
- `ots_fills_all_four_parts` compares `parts.octave` with the raw `o.octave`. That only holds because `parse_ots` already clamps to −2..2. Fine, but `apply_ots` clamps twice.

**Verdict: changes requested (1 blocking).**

ryangavin:
Round 1 review addressed in daa21c2.

**Addressed**
- **Blocking: with Left off, keys left of the split were silent in every mode.** Fixed. `sounds()` now takes a `chord_only` flag. With Left off, keys left of the split play the Right parts that are on (OM p.48) in Upper detection and in Full Keyboard / AI Full Keyboard. They stay chord-only in Lower detection with the other fingerings (OM p.56, ACMP on and LEFT off). The test `left_off_right_parts_cover_the_keyboard_unless_chord_section` drives `Input::key_msg`: in Upper, Full Keyboard and AI Full Keyboard, key 40 gives note-ons on ch 1 and ch 3 (Right 1 + Right 2), and the note-offs match. In Lower Fingered and Fingered On Bass it gives nothing, and the chord is still read.
- **Keyboard CC7/CC0/CC32/PC bypassed Parts.** These messages from the keyboard are now dropped (`keyboard_volume_and_voice_messages_are_dropped`), so a part's CC7 and program are the only ones that exist.
- **Manual Bass fader and LEDs.** Left now always plays at its own volume, so Panel fader 4 is live under Manual Bass; `bass_vol` and `channel_volume` are gone. The LEDs use the new `Parts::sounding_mask()`, the same as the TUI.
- **Sustain excluded Left.** Pedals, wheels and pressure now reach all four parts, per the Reference Manual's Assignable functions table (Sustain affects all notes played on the keyboard).
- **FadersAt lost to a full ring.** `Cmd::FadersAt` is replaced by a flag and a position snapshot in `Parts`. The engine takes them on each wake and before applying any Style fader move, so nothing can drop the rebind. `style_rebind_survives_a_full_ring` fills the ring at the switch and checks there is no jump.
- **"Closes #31".** The body now says "Part of #31". `docs/backlog.md` marks the parts model done and lists split points and Left Hold as still open.
- **apply_ots double clamp.** Removed. `parse_ots` already clamps.
- Also fixed one clippy `needless_range_loop` in a `parts.rs` test.

**Declined**
- None.

**Decisions changed** (PR body updated; also posted on #31)
- Decision: with Left off, the Right parts play left of the split except in the Lower chord section, because OM p.48 and p.56 say so. This replaces "the left hand is silent while Left is off".
- Decision: pedals, wheels and pressure go to all four parts, Left included, because the RM says Sustain affects all notes played on the keyboard. Per-part routing is left to #34.
- Decision: the keyboard's own CC7, CC0/CC32 and Program Change are dropped, because a part's volume and voice are its own CC7 and program.
- Decision: under Manual Bass, Left uses its own volume (Panel fader 4), not the Style's Bass fader, because OM p.51 assigns the Bass voice to the Left part.

Gates: `cargo build --release` is clean. `cargo test --release` gives 209 passed and 1 ignored; the corpus ran and nothing was skipped. `cargo clippy --release` shows no new warnings in touched code. The golden digests are unchanged.

ryangavin:

**Setup:** I checked out `origin/m2/keyboard-parts` (daa21c2) detached, with the `corpus` and `soundfonts` symlinks in place.

**Gates:**
- `cargo test --release`: 209 passed, 1 ignored. No test printed "skipping", so the corpus and soundfont tests ran.
- Golden digests: unchanged. No files under `tests/golden` changed.
- `cargo clippy --release`: nothing new in touched code. The two `live.rs` hits are the older collapsible-`if` patterns.

I wrote throwaway tests and corpus scripts, then removed them. Nothing was pushed.

### Round 1 fixes checked
- **Blocking #1 (Left off leaves the left of the split silent): fixed.** `sounds()` with `chord_only` is correct for Upper, Full Keyboard, AI Full Keyboard, Lower Fingered and Fingered On Bass. `left_off_right_parts_cover_the_keyboard_unless_chord_section` goes through `Input::key_msg` and checks note-ons, note-offs and the chord. I reread OM p.48, p.51 and p.56 to confirm.
- **#2 (keyboard CC7, CC0, CC32 and PC):** dropped, and tested.
- **#3 (Manual Bass fader and LEDs):** the pad page 3 LED, the fader-button LED and the TUI all use `sounding_mask()` now (`ui.rs:72`, `ui.rs:499`). Panel fader 4 is live.
- **#4 (sustain):** I checked the RM Assignable Functions table (Sustain: "all notes played on the keyboard"). Sending it to all four parts is a sound choice.
- **#5 (full ring):** the flag and snapshot replace the command, and `rebind_faders` runs both before the rings drain and inside `Cmd::PartVolume`. `style_rebind_survives_a_full_ring` really fills the ring.
- **#6 and #7:** done.

### Independent checks
- **OTS channel and part mapping, checked against the corpus.** I reran the counts over all OTS tracks in the corpus.
  - PSR-S930 styles (no Right 3): ch 3 of the OTS track always holds the GM default (bank 0/0, PC 0, CC7 127 in 87 of 92 OTS), and SysEx part 2 is always off (96 of 96).
  - ch 4 holds the lower-level voices (CC7 mostly 50 to 75), and SysEx part 3 is the only one with param 0x02.
  - So OTS ch 1–4 = Right 1, Right 2, Right 3, Left, as `parse_ots` and `apply_ots` assume.
  - This is not the Genos song and port channel order (RM p.75: ch 1 Right 1, ch 2 Left, ch 3 Right 2, ch 4 Right 3). `parts::CHANNEL` uses that song order for output, so both sides are right.
- **Mixer principle:** the parts' volume is only their CC7, sent from the engine thread via `sync_part_volumes`. The player synth has no gain left. rustysynth's CC121 does not reset CC7, so a keyboard Reset All Controllers can't desync the cache.
- **RT safety:** `sounds`, `Keys`, `fader_button`, `set_fader_page`, `take_rebind`, `sync_part_volumes` and `sync_player` use only fixed arrays and atomics. They don't allocate, lock or panic.
- **Leaked content:** none. The tests assert only voice numbers and volumes, and no note listings were added.

### Blocking
None.

### Nonblocking
1. **Octave folding makes two held keys share one note, and releasing one cuts the other** (`live.rs` `sounds()` → `engine::shift_key`).
   - Repro: Right 1 octave +2. Key 96 on, key 108 on, key 108 off.
   - Output: `[90 78 5A] [90 78 5A] [80 78 00]`. Key 96 is still held but its note 120 was just released.
   - Both keys are on an 88-key board. Transpose makes it more likely.
   - The old synth folded the same way. The port now carries it too, because it gets the octave-shifted notes.
   - A possible fix: drop notes that go out of range instead of folding them, or count note-ons per (ch, note).
2. **Polyphonic aftertouch (0xA0) goes out with the raw key** (`live.rs` controller branch).
   - With Right 1 at octave +1, key 60 sounds note 72, but `A0 3C` reaches ch 1, 2, 3 and 4 as key 60.
   - So it targets a note that isn't sounding. A cheap fix is to map it through `keys.sounding[k]`, or to drop it.
3. **Manual Bass after an OTS recall: the bass plays an octave or two up, at the Left pad level.**
   - In the corpus, Genos and Tyros OTS set Left's octave to +1 in 654 of 696 OTS and to +2 in 40. Left's CC7 is mostly 50–70.
   - Upper mode turns Manual Bass on by default. After any OTS recall, the Style's Bass voice then plays 1–2 octaves above the keys, at pad volume.
   - The octave part is not new (the base synth applied `left_oct` under Manual Bass too). But it is now on the port as well, and the volume moved from the Style Bass fader to Left.
   - The Decision is defensible (OM p.51, "assigned to the left part"). The owner should know this is what Upper plus OTS will sound like in practice. Consider ignoring Left's octave under Manual Bass, and recording that as a Decision.
4. **Memory ordering on `Parts::changed`.**
   - `apply_ots` and `step_program` store `program` and then `changed`, both `Relaxed`. The synth does `changed.swap(false, Relaxed)`.
   - On Apple Silicon the synth can see `changed` before the new `program` and clear the flag. The stale voice then stays until the next change.
   - Fix: `store(true, Release)` and `swap(false, Acquire)`. The same pattern existed as `slots_changed` before, so it isn't a regression.
5. **Small race between OTS recall (UI thread) and a Panel fader move (input thread).**
   - `set_volume` and `hw_fader` both read, modify and write `volume` and `picked`.
   - If a fader moves while the OTS recall is running, the fader can overwrite the OTS level and keep `picked=true`.
   - The window is microseconds, so this is noted only.
6. **Under Manual Bass, Panel button 4 and the Left pads flip `on[LEFT]` with no audible or visible effect.** The LED stays lit through `sounding_mask`. Consider ignoring the toggle, or flashing the LED, while Manual Bass holds Left.

**Verdict: approve-able. No blocking issues remain. The nonblocking items can follow up.**

ryangavin:
Review round 2, addressed in c545304. There were no blocking items. All six nonblocking items are addressed.

**Addressed**
- **Octave folding and shared notes:** `Keys` now counts the held keys per (channel, note). A note's off goes out only when the last key holding it lets go. This covers an octave fold (the key 96/108 at +2 case) and a transpose change between two presses that lands them on the same note. Test: `shared_note_stops_with_its_last_key`.
- **Poly aftertouch:** `A0` now goes to the notes its key actually sounds, on those parts' channels, after the transpose and octave. A key that isn't sounding sends nothing. Test: `poly_aftertouch_follows_the_sounding_notes`.
- **Manual Bass after an OTS:** Left's octave is no longer applied to the bass under Manual Bass (`Parts::octave_of`), and the screen shows the octave actually in effect. Tests: `manual_bass_ignores_left_octave_and_switch`, updated `right_parts_layer_and_left_follows_its_switch`.
- **`Parts::changed` ordering:** Release stores, and an Acquire swap in the synth callback.
- **`set_volume` / `hw_fader` race:** a part's volume and its fader pickup bit now share one `AtomicU8`. `hw_fader` does a `fetch_update`, so a move is always judged and stored against a single reading, and an OTS level can't be overwritten by a move judged against the old one. Test: `fader_move_never_overwrites_a_concurrent_recall`, a threaded stress test.
- **Left switch under Manual Bass:** `Parts::toggle(LEFT)` is refused while Manual Bass is on, and the UI tells you to turn Manual Bass off (`D`). On/off stays consistent with the lit LED.

**Declined**
- None.

**Decisions** (added to the PR body and posted on #31):
- Decision: under Manual Bass, Left's octave shift is not applied to the bass. OM p.51 assigns only the Bass voice to the left part, and Left's octave is for its own voice. In the corpus OTS, Left is at +1 in 654 of 696 settings and +2 in 40.
- Decision: under Manual Bass, Left's on/off is refused, because the switch would change nothing audible or visible.

**Gates:** release build OK. `cargo test --release`: 213 passed, 0 failed, and the corpus tests ran (no skips). `cargo clippy --release`: no new warnings in the touched code.

ryangavin:
## Rebased onto `integration/m2-chord-conversion` (a1daaf5)

This PR is squashed to one commit, `8175195`, and rebased onto a1daaf5. That base has #56, #55, #59, #58 and #57. It does **not** have #60.

- **`src/main.rs`**: the module list now has both `mod oracle;` (#58) and `mod parts;`.
- **`README.md`**: this PR's keyboard-parts, channel and two-page mixer text replaces the old voice-slot/layer bullets. Its Style-page bullet predated #59: it said "Section changes keep the levels you set. Start/Stop puts only the parts you have not touched back". I replaced that sentence with #59's description:
  - every section change re-plays the SInt;
  - untouched parts go back to the style's levels, and user-set levels are kept;
  - a mid-bar Fill or Break plays the voice changes of its skipped beats;
  - Start/Stop also sends the XG effects, and no GM/XG resets are sent.

  Both behaviours are now documented.
- There were no code conflicts. `engine.rs` merged cleanly: this PR adds only `Takeover::at`, `Takeover::resume` and `Engine::faders_at`.

### Mixer principle check: #59's SInt re-apply against this PR's part CC7

The principle is that CC7 is the only per-part volume and that the user's faders win. **#59 and this PR agree with it**:

- **They use separate channels.** #59 routes the SInt through the CASM rules to destination channels. It re-sends it on every start and section change. All 7,529 corpus rules have destinations in ch 9–16, so the SInt never reaches this PR's keyboard parts on ch 1–4. The same holds for pattern CC7 (`pattern_volume`) and XG Multi Part SysEx, which is routed to the destination part.
- **New guard.** `sim::corpus_full_performance_no_stuck_notes` now also asserts that no engine message is sent on a keyboard part's channel. That covers channel messages and XG Multi Part SysEx. The test plays every corpus style through Intro, the Mains, Break, Fill, Ending, transposes and a start, so #59's re-apply runs many times.
- **Style parts.** #59 puts the SInt CC7 of each part into `mix` rather than into the re-sent init. It drops the XG part-volume SysEx for the parts. On a section change, it restores only the parts the player has not moved (`user_set`). This PR's page switch (`faders_at`) rebuilds only the soft-takeover state and never touches `user_set` or the levels. So a fader the player moved still wins across page switches and section changes.
- **Keyboard parts.** Their only volume is the part's CC7, sent by the engine thread on the part's channel whenever it changes. The keyboard's own CC7, CC0/32 and Program Change are dropped. Nothing in the style path re-sends those channels, and the GM/XG resets that could clear them are never sent (#59). An OTS recall moves a part's level as a software move, and the Panel fader has to pick it up (soft takeover).
- **Style mutes** gate notes and don't use CC7, so a mute can't drift a part's level.
- **Manual Bass**: Left plays the Style Bass voice at Left's CC7 on ch 2. The Style Bass part's own CC7 stays on ch 11, and #59's re-apply sends it only there.

### Gates

- `cargo test --release`: **255 passed, 0 failed, 1 ignored**. The corpus and soundfonts were symlinked, and nothing skipped.
- `cargo clippy --release --all-targets`: no new warnings. There are 6 fewer than on the a1daaf5 tip.
- **Golden digests: unchanged.** Nothing was regenerated.

## Oracle delta

**None.** `yahaha oracle corpus/ --diff tests/oracle/scores.txt` prints `no change`, and `corpus_scores` passes on the a1daaf5 pin. This PR doesn't touch the transposer. The pin already includes #55's delta, analysed on #57.

## Concerns

- **Minor, and older than this PR:** `Prepared::new` sends a SInt CC7 as a raw init message when a rule's destination is outside ch 9–16 (`Some(v) => init.push(CC7)`), and `pattern_volume` does the same for such channels. No corpus rule has such a destination. But a malformed style routing to ch 1–4 would then write a keyboard part's CC7, and it would also play notes on that channel. The new corpus guard would catch that for any corpus style. It might be worth clamping destinations at parse time.
- `synth::style_bass_program` still follows only the SInt Bass voice under Manual Bass. That TODO is already in docs/backlog.md.
- **#60 is not in this base.** `git merge-tree` of this head with #60's rebased head (`2238b98`) gives no textual conflicts, so whichever of the two lands second should rebase cleanly. Its tests still need to be re-run on the combined tree.

Issue #9, opened 2026-09-23
Genos-owner reference capture kit


ryangavin:
- Write a test .mid chord script with instructions for a Genos/PSR owner: load style X, play the script into ACMP, and record the MIDI out.
- Write a diff tool that turns their recording into golden data. This produces the forum post draft; posting it is up to Ryan.

---
Milestone: **M2 · Chord-to-note conversion 1:1** · Spec: [docs/genos-features.md](../blob/main/docs/genos-features.md) · Backlog: [docs/backlog.md](../blob/main/docs/backlog.md)

ryangavin:
Decisions made in #62 (capture kit). Playtest and overrule as needed:

- **Decision:** the owner plays the kit from a DAW into USB port 1, with ch1 set to Receive = Keyboard, ACMP on and SYNC START armed on the panel. The sections are sent as Section Control SysEx (`F0 43 7E 00 ss 7F F7`, then `ss 00` as the release), because the Data List documents Section Control reception for Style, and it is the only way to press section buttons from a file.
- **Decision:** no Tempo Control SysEx. The owner leaves the style tempo alone, and the importer measures the tempo and refuses to verify beyond 0.2%. This is because the `t4..t1` tempo encoding is undocumented.
- **Decision:** the kit uses Fingered On Bass with the F#2 split, and close root-position voicings with the bass lowest, between E0 and F#2. Every script chord must read back through our recognizer as itself. Ambiguous shapes are written as we read them: C6/G (Am7 over G), F6(9)/G (Dm11 over G), Fadd9/G, C7/Bb, Bbadd9/C. The recording's Chord SysEx shows how the hardware reads them. This is because the style diff only means something when both sides follow the same chord.
- **Decision:** the kit styles are the six golden styles, all from the free MOX performance styles V2 pack so owners load byte-identical files, then the factory 60s8Beat, SoulShuffle and RockShuffle, compared with our Tyros5 copies. This is because only identical style data makes a diff meaningful. The drum-match check flags factory data that differs.
- **Decision:** bar 1 is found from the drums (the parts that play as written), followed by a least-squares tempo and offset fit, with a 25 ms tolerance. Matched notes take our timing and length. Unmatched notes take the timing of our nearest note on the same channel. This is because MIDI jitter is not information and must not change a digest.
[...]

PR #62, opened 2026-09-23
Genos-owner reference capture kit (#9)

ryangavin:
## What and why

This is the kit that lets Genos and Genos2 owners give us ground truth for M2. An owner plays our chord script into ACMP and records the MIDI the style sends back. We then diff that recording against our engine, bar by bar and part by part.

**`yahaha capture-kit <out-dir> [style]...`**
- Renders `docs/capture-kit/capture.script` as one type 0 MIDI file per kit style, at that style's tempo, time signature and ppq. With no styles given, it uses the kit styles found in `./corpus`.
- Each file has a two-bar lead-in (Main A is selected during it), then the chord keys on ch1, a Section Control SysEx for every button, and a marker on every bar.
- Nothing is sent on a beat line (`capture::plan`): buttons go half a beat ahead of their slot, and chords a per-style lead ahead (see Decisions). Our engine plays the kit at exactly the same ticks (`sim::perform_steps`).
- Every chord is voiced close, in root position, with the on-bass note lowest. All keys sit between E0 and F#2.
- Before writing, the generator checks that every voicing reads back through our Fingered On Bass recognizer as the chord our engine is given.
- It also writes the owner instructions (`docs/capture-kit/README.md`), plus a table of files, tempos and durations.

**The script** (about 100 bars, 140 chord changes) covers:
- all 38 chord types, including Cancel, 1+5, 1+8 and the three display-only types, with roots moving round the circle of fourths;
- major and minor on all 12 roots;
- 16 slash chords and inversions;
- 2, 4 and 8 changes a bar, changes off the beat, and a release;
- Main A→A (its own fill), whole-bar fills into B and D, a whole-bar Break leading into Main C, a one-beat fill into A with a chord change on its first beat, and Ending A;
- no chord struck again while it is held (the test checks this).

**`yahaha capture-import <recording.mid> <style>`**
- Reads any type 0 or 1 SMF (tempo map, any ppq).
- Finds bar 1 by lining up the parts that play as written (drums). It tries each part's first note, at the style tempo and at the tempo the drums' span suggests, then fits tempo and offset by least squares over every matched note.
- Reports:
  - the alignment, the instrument's tempo, and how closely matched notes line up (median and 99%, the setup's jitter);
  - how far the chords landed from where yahaha plays them, against the style's margin;
  - every chord the instrument read differently from the script, taken from its Chord Control SysEx or Song chord meta events;
  - per bar and part, the notes only the instrument played, the notes only yahaha played, and differing as-written counts;
  - a per-part summary.
- `--golden <dir>` writes `<style>.digest` (hashes only, the golden digest format, rendered by `Take::render_reference`: bar lines with the script steps only, every note of every part hashed) and a `<style>.known` list of the current differences. It does this only for a verified recording, unless `--force` is given. `--listing` writes the readable listing for local use only.

**Reference test.** `capture::tests::reference_captures` plays the capture script on every style that has a digest in `tests/reference/`. It fails on any bar/part difference that is not listed in that style's `.known` file, and says when a listed difference has been fixed. `tests/reference/` holds only a README for now; digests arrive with owners' recordings.

**Refactor, no behaviour change.** `sim::snapshot` is split into `sim::perform` (returns a `Take`), `sim::perform_steps` (steps at given ticks; `sim::golden_act` is the old one-tick-early rule) and `Take::render`, so a recording can be rendered in exactly the golden listing format. `fnv1a`, `part_key` and `digest` moved from the test-only `golden.rs` to `sim.rs`. Golden listings are byte-identical, and no digest changed. `sff::parse_track` is now `pub(crate)`. Chords can be spelled `C6(9)` / `C69`, because `C6/9` reads as a bass note.

**Forum post draft:** `docs/capture-kit/forum-post.md`. Posting it, and attaching the zipped kit, is up to the owner. The kit's MIDI files contain only our script, no style data.

Nothing readable from any style is committed. The script, instructions and post are our own text, and reference digests are hashes. `committed_references_hold_no_notes` enforces this for `tests/reference`.

## How tested

- `cargo test --release`: 218 passed, 1 ignored (pre-existing, in synth.rs). The corpus is symlinked and the corpus tests ran (golden, and every kit style in the round trip and jitter tests). `docs/manuals` was symlinked too, so the Data List test ran. Golden digests are unchanged.
- `cargo clippy --release` (and `--all-targets`): no new warnings in the code I touched.
- Tests:
  - `script_reads_back_and_covers_the_kit`: every chord is inside E0..F#2 and reads back as itself; no chord is struck again while held; every button has a Section Control code and presses outside a bar's first beat; all 38 types, at least 12 slash chords and an 8-change bar are present.
  - `voicings`, `chord_bytes` (Chord Control decoding).
  - `kit_midi_plays_the_script`: exact key and SysEx timing after the lead-in, and no event within 5% of a beat of a beat line.
  - `kit_timing_survives_jitter`: for every kit style, moving every step 3 ms earlier or later leaves the sections and notes the same (a cut or retriggered note moves with its chord), and the style has more than 3 ms of room. With the golden timing (steps on the line), arriving 3 ms late does change the take.
  - `import_round_trip`, for every kit style: a fake recording of our own take at 480 ppq on a different clock, bar 1 at 3.2 s, ±4 ms order-preserving jitter, chord SysEx on the computer's clock and keyboard echo on ch4. It imports with 0 differing bars, verifies, and yields exactly our reference digest.
  - `import_type1_with_notes_before_bar_1`: a type 1 file with the tempo on its own track, at another tempo, with three notes played before bar 1.
  - `import_reports_differences`: one bass note moved shows up as bar 20 ch11 only; a 1% tempo error fails verification.
  - `import_clock_drift_over_every_style`: for every kit style, the instrument's clock off by half the style's budget verifies. Off by twice the budget (still inside the tempo check), it fails with and without Chord SysEx, and the report says recording again will not help and gives the measured `--clock-ppm`. A kit made with that value verifies again and yields our reference digest. `kit_clock_correction_sets_the_tempo` checks the file's tempo.
  - `first_beat_presses_are_refused`: a button on beat 2 presses in beat 1, and `kit_midi` refuses it. A button on beat 3 or on a downbeat is fine.
  - `import_rejects_a_missed_section_change`: a recording where one section button was not taken fails the drum check.
  - `write_reference_files`: `--golden` refuses an unverified recording, `--force` writes it, and `.known` lists the differing bar and part and names the recording by file name only.
  - `reference_digest_ignores_our_sections`: moving one of our section changes changes the golden listing but not the reference digest.
  - `factory_kit_styles_are_genos_presets`: every factory kit style is a Genos preset name in the Data List (skipped without the manuals).
  - `reference_captures`, `committed_references_hold_no_notes`, `digest_differences_name_bars_and_parts`.
- Ran `yahaha capture-kit` on the corpus: 9 files.

## Decisions

- Decision: the owner plays the kit from a DAW into the instrument's USB port 1, with ch1 set to Receive = Keyboard, ACMP on and SYNC START armed on the panel. The sections are sent as Section Control SysEx (`F0 43 7E 00 ss 7F F7`, then `ss 00` one tick later as the button release), because the Data List documents Section Control reception for Style. This is the only way to press section buttons from a file.
- Decision: no Tempo Control SysEx in the kit. The owner leaves the style at its own tempo, and the importer measures the tempo and refuses to verify beyond 0.2%. This is because the Data List does not document how `t4..t1` encode the tempo.
- Decision: the kit uses Fingered On Bass, with the default F#2 split. Voicings are close root position with the on-bass note lowest, kept between E0 and F#2. Every script chord must read back through our recognizer as the chord the engine is given. Ambiguous shapes are written the way we read them: G A C E is C6/G, not Am7/G; Dm11 over G is F6(9)/G; F over G is Fadd9/G; C over Bb is C7/Bb; Bb over C is Bbadd9/C. The instrument's Chord SysEx in the recording shows how the hardware reads them, and the report lists every disagreement. This is because the style comparison only makes sense when both sides follow the same chord.
- Decision: kit styles:
  - First, the six golden styles (Organ Cruise, Jack Does It Again, Smooth It Over, Blues Organ Trio, Austin City Blues, Ticking Away). They are all in Paul Drongowski's free MOX performance styles V2 pack, so owners load byte-identical files.
  - Then three Genos factory presets that have Tyros5 copies in our corpus under the same name: 60s8Beat, SoulShuffle and 90sDisco (several source channels per part). RockShuffle was dropped: the Genos has only US RockShuffle, PopRockShuffle and Rock&RollShuffle.
  - This is because only identical style data makes a diff meaningful. For the factory styles, the drum-match check reports whether the Genos data differs from our copy.
- Decision: bar 1 is found from the parts that play as written (drums), which are chord-independent, followed by a least-squares tempo and offset fit over all matched notes. The tolerance is 8 ms by default, and the report shows the measured jitter so it can be adjusted. A matched note takes our start, and our length when the ends also agree within twice the tolerance (an end can come a burst of MIDI traffic late). An unmatched note takes the start of our nearest note on the same channel within tolerance. Notes that start together keep our order. This is because USB MIDI jitter is not information, and it must not change a digest or push a note into the neighbouring bar.
- Decision: a recording is "verified", and so allowed to become a reference, when:
  - at least 98% of our as-written notes are matched;
  - the tempo is within 0.2%;
  - no chord landed further from where we play it than its room (next decision);
  - the recording reaches our last note, within one bar.

  Chord-reading differences do not block verification, but they are listed first. This is because those checks catch the wrong style, sections not received, a changed tempo or a truncated recording, while a different chord reading is itself hardware data we want.
- Decision: nothing in the kit goes out on a beat line. Section buttons press half a beat ahead of their slot, because the Genos changes sections on a beat or a bar line (RM p.12, Section Change Timing) and half a beat is as far from both as a press can be. Chords (and releases) go out a little ahead of their slot. For each position in the beat the script puts chords on, the lead is the shortest between a 24th and a quarter of a beat that keeps furthest from every note start and end, and every point 40 ms after a note start (our engine's late-chord window), in the sections the script calls up, rhythm channels left out. It depends only on the style's patterns and the script, not on the engine, so a kit file stays valid while M2 changes the engine. This is because a fixed lead left several styles with notes a 3 ms shift would cut, retrigger or revoice differently (measured over the kit; the new test proves the plan is stable). The on-beat leads chosen for the kit run from 0.09 of a beat (Jack Does It Again) to a 16th (Blues Organ Trio): early, but no earlier than a player who changes ahead of the beat. The first chord starts the style and is bar 1 itself.
- Decision: a recording verifies only if every chord landed within its own room of where we play it. The room is the distance from where the chord goes out to its slot, or to the nearest point where a chord would decide something in the sections playing around it: a note start or end, or the point 40 ms after a start (rhythm channels left out). The slip is measured from the Chord SysEx relative to the first chord, so the instrument's own delay in reading chords cancels out. Without Chord SysEx it is estimated from the fitted drift. This is because the computer plays the chords on its clock and the instrument the style on its own, and a drift inside the 0.2% tempo check is most of a beat by the end. The per-chord room replaces the kit-wide minimum, which rejected recordings where no chord came near a note. Each style then takes 32 to 113 ppm of clock difference: 90sDisco 32, TickingAway 38, SoulShuffle 43, OrganCruise 56, and the rest 80 to 113. The kit's table lists it, and the report prints it next to the measured drift.
- Decision: when a pair of clocks drifts past a style's budget, the fix is a kit made for that instrument. `yahaha capture-kit <dir> --clock-ppm N` runs the file N ppm faster, and `capture-import --clock-ppm N` imports its recording; the report prints N as it measured it. This is because the drift between a given computer and instrument stays the same, so recording again cannot help, and the least-squares tempo fit measures it to about a ppm. I did not slave the instrument to the DAW's MIDI clock (RM p.151, Clock = USB1). With an external clock, RM p.151 says the Style cannot be controlled by the instrument's buttons. The style would then start on the DAW's Start (FA) at the top of the file instead of by Sync Start on the first chord, so the lead-in, the Intro press and the first-chord start would all have to change, untested on hardware.
- Decision: a fill called up by a press half a beat before a beat line starts on that beat (`engine::queue_fill`). This is because RM p.12 documents when the Main section changes but not where the fill starts. The report puts the drums of every fill bar on a line of their own, so a recording shows at once whether the Genos starts the fill somewhere else.
- Decision: the kit is for the Genos and Genos2 only. It plays keys down to E0 (MIDI 28), below the 61-key PSR-SX range, and we do not know whether an SX reads chord keys it cannot play. This is because a wrong reading would poison a reference.
- Decision: reference digests hold only the script steps in their bar lines, and hash every note of every part (`Take::render_reference`). This is because the instrument does not send its section changes, so our section timeline would be a guess, and it would make the first M2 change to section timing fail against hardware that never reported it.
- Decision: reference digests live in `tests/reference/<style file>.digest`, in the golden digest format, with a `.known` list of accepted differences. The test fails only on new differences, and reports fixed ones. This is because the engine will not match the hardware on day one, and the list should shrink visibly as M2 lands.

Closes #9


ryangavin:

Reviewed `origin/m2/capture-kit` @ 00cd96b with the corpus symlinked.

- `cargo test --release`: 209 passed, 1 ignored. All 9 `capture::` tests ran against the corpus; `import_round_trip` went through all 9 kit styles.
- `cargo clippy --release --tests`: nothing new in `capture.rs`. The one warning in `sim.rs:394` was already there.
- Ran `yahaha capture-kit` locally: 9 files. All tempos are integer bpm, 1920 ppq, except TickingAway (480).
- No readable style content is committed. The kit `.mid` files hold only our script plus the style name in a text meta. `tests/reference` holds only a README.
- No RT-thread code is touched, and the mixer is not involved.

Throwaway tests for the findings below were run and then removed; nothing was pushed.

### Blocking

**B1. Button presses and chord changes sit right on beat and bar lines, so hardware results depend on jitter and clock drift.**
- `src/capture.rs:215`: Section Control goes out at `t0 + s.tick - 1`. At 1920 ppq that is 0.27 ms before the beat (about 0.26–0.29 ms across the kit tempos). Chord keys go out exactly on the beat (`t0 + s.tick`), with note-ons in ascending order.
- The DAW plays the kit on the computer's clock, and the instrument runs its own clock from the moment Sync Start fires. USB MIDI jitter is around 1 ms, and the drift between the two clocks keeps growing over a 3–5 minute take. Together they decide which side of the boundary each event lands on:
  - A button pressed at 3.1919 can arrive on beat 4. That changes the fill start: a 1-beat fill becomes one that starts in the next bar.
  - A downbeat chord can arrive after the downbeat notes have fired. The part then plays the old chord, or a partial one (C alone, then C+E), and gets retriggered by its RTR instead of starting cleanly on the new chord.
- Our own `press_tick` comment in `sim.rs` names this hazard: "a press exactly on a beat or bar line would otherwise depend on float rounding". The kit carries the same hazard onto real hardware, where it depends on jitter instead of rounding.
- A reference digest made this way is not reproducible between owners, or between two takes by the same owner. That defeats the kit's purpose. Several things point at timing-sensitive behaviour here:
  - Sim already shows it is sensitive to the press position. Every fill and the Break in the script lasts exactly one beat (see headers below).
  - TickingAway (480 ppq) already behaves differently at the same script position. Bar 71 is `Fill In BA` with no `Fill In CC`, where every 1920-ppq style has `Main B > Fill In CC@4`.
- Fix:
  - Put button presses well inside a beat. Send them half a beat or a full beat early, and make sim press at the same tick.
  - Send chord keys a fixed lead ahead of the beat, e.g. 10–20 ms (or 1/64). Give sim the same lead, or else record a Decision on why the lead is not needed, backed by evidence.

**B2. Verification accepts tempo slip large enough to put the script's chords most of a beat away from the instrument's beat.**
- `src/capture.rs:623` allows `|a − 1| ≤ 0.2%`.
- The chords come from the DAW's clock and the style comes from the instrument's, so any difference in tempo shows up directly as chord-vs-beat slip.
- Throwaway test: `fake_recording(&ours, 1.0, 1.0019, …)` on AustinCityBlues. Result: `verified=true`, ratio 1.00190, all 8001 notes matched. That is 555 ms of slip by the end, about 0.76 beat at 82 bpm.
- Real crystal drift is about 50 ppm, but the check should bound what matters. For example:
  - require `|a−1| × take length` to stay under a few ms, or
  - better, check that the instrument's Chord SysEx timestamps line up with the script's chord ticks after the fit. Today they do not affect verification at all.
- As it stands, a recording where the DAW and the instrument disagree on tempo passes and can be turned into a reference.

**B3. One of the "factory styles" in the owner instructions does not exist on the Genos.**
- `KIT[8]` (`src/capture.rs:69`) is "RockShuffle".
- The Genos Data List style list (`Genos_data_list.txt` lines 3330–3420) has no preset of that name. It has `US RockShuffle` (line 3399), `PopRockShuffle` (3374) and `Rock&RollShuffle`.
- Nothing shows that the Tyros5 `RockShuffle.T162.prs` is the same data as any of these. Our copy's internal name is `RockShuffle.T162.prs`.
- The README tells owners to "load them from the Preset tab", so an owner will either not find it or pick a different style.
- `60s8Beat` (line 3340) and `SoulShuffle` (line 3397) do exist by name.
- Fix: drop RockShuffle, or name the exact Genos preset and say in a Decision that it may not be the same data as our copy.

### Nonblocking

**N1. The hardware take's bar headers and as-written split are ours, not the hardware's** (`src/capture.rs:554`, `hardware = ours.clone()`).
- The reference digest keeps yahaha's section timeline and chord steps, as of import day, in every `bar` header. It also counts or lists notes using *our* section-at-tick.
- The hardware's section timing is therefore never captured. As soon as an M2 fix changes section timing (for example the 1-beat Break), `reference_captures` will fail with `bar N header` differences that the hardware never reported.
- Suggest leaving headers out of the reference comparison, or storing only the script-derived part of the header.

**N2. The 25 ms tolerance is large for the behaviour M2 is about.**
- 25 ms is about 138 ticks at 172 bpm.
- A matched note takes our start, and our length when the ends agree within 25 ms. An unmatched note is snapped to the nearest same-channel note within 25 ms.
- Guitar NTR strum spacing (typically 5–20 ms per string) and RTR note-off timing differences are erased before hashing. USB/DAW jitter is about 1–2 ms.
- Suggest 5–8 ms by default, or at least a Decision that measures the jitter in a real recording.

**N3. Identical chords are re-struck.**
- Examples: `G7 C | C Am F G7`, and `Am D7 G7 C | C F …`.
- The kit sends note-off then note-on for the same keys at the same tick. That is a new press that no legato player makes.
- The instrument may not send a Chord SysEx for an unchanged chord, so the report will count these as "no chord message".
- Hold them (`-`) instead, unless re-striking is the thing being tested.

**N4. Section coverage is thinner than the script comments claim.** Sim headers for all 9 styles:
- Every fill and the Break lasts one beat, from `@4` to the bar line.
- JackDoesItAgain never reaches Main D (bar 94 is `Main C > Fill In AA`), so "Main D: Cancel, 1+5, 1+8 in context" is Main C for that style.
- Ending A is one bar for 8 of the 9 styles, and the last 3 bars are `stopped`.

Consider pressing fills earlier in the bar (this also helps B1), and trimming the stopped tail.

**N5. `.known` header leaks a local path** (`src/capture.rs:769`).
- The header embeds `{recording}`, i.e. whatever path was passed, e.g. `/Users/<name>/Downloads/…`, into a file that gets committed to a public repo.
- Use the file name only.

**N6. Tests.**
- `import_round_trip` is an identity test: our own take goes back in.
- Nothing tests:
  - that `--golden` refuses an unverified recording, or that `--force` overrides it;
  - `.known` generation;
  - a recording with notes before bar 1, a type 1 file with the tempo map in a separate track, or a missing section change.
- `reference_captures` passes trivially until digests exist. That is fine, but worth saying in the PR.

**N7. Owner instructions.**
- The README says the file table is "listed at the end of this file", but `docs/capture-kit/README.md` in the repo has no table (only the generated copy does).
- Consider adding to the settings list: Style Setting → Section Change Timing, Dynamics Control, and all style channels on in the Mixer.
- The forum post invites SX600/700 owners. Those are 61-key instruments (C1–C6), and the kit plays keys down to E0 (MIDI 28). Say that notes below C1 arrive over MIDI and have not been confirmed to be read, or drop the SX600 mention.

**N8. Churn.**
- The `sim::snapshot` split, and moving `fnv1a`/`part_key`/`digest` out of `golden.rs`, is a behaviour-neutral refactor. Golden digests are unchanged, which I verified.
- It will still conflict with any sibling M2 PR that touches `sim.rs`'s snapshot or `golden.rs`. Expect to rebase.

### Verified OK
- Section Control and Chord Control byte layouts match Data List pp.110–111:
  - switch numbers 00–03, 08–0B, 18 and 20–23;
  - cr in `0fffnnnn` form, with fff = 3 for natural;
  - ct 0–34, with 34 = cc;
  - bn = 127 means no bass.
- The `ct > NUM_TYPES` bound is correct, since NUM_TYPES = 34 = CANCEL.
- Velocity-0 note-ons are treated as note-offs (`sff::parse_track`).
- The E0 (MIDI 28) to F#2 (MIDI 54) key range matches the Genos 76-key range and the default split.

Verdict: **changes requested** (B1–B3).

ryangavin:
Round 1 review addressed in 50d2f8c.

**Blocking**

- **B1 (steps on the beat line): fixed.** Nothing in the kit goes out on a beat line now (`capture::plan`).
  - Buttons press half a beat ahead of their slot. The Genos changes sections on a beat or a bar line (RM p.12, Section Change Timing), so half a beat is as far from both as a press can be.
  - Chords and releases go out a per-style lead ahead of the slot. I tried a fixed lead first, and a scan over the kit showed it isn't enough: with any single lead, a 3 ms shift still changed notes in several styles (up to hundreds of notes for Austin City Blues and 90sDisco at a 64th). Those are notes that start or end next to the chord, or start 40 ms before it, which is our engine's late-chord window.
  - So, for each position in the beat the script puts chords on, the plan picks the shortest lead between a 24th and a quarter of a beat that is furthest from all of those points. The points come from the sections the script calls up, with the rhythm channels left out.
  - The plan depends only on the style's patterns and the script, never on what the engine does, so a kit file stays valid while M2 changes the engine.
  - `sim::perform_steps` plays the same ticks, and golden timing is unchanged (`sim::golden_act`; no digest changed).
  - Proof: `kit_timing_survives_jitter`. For every kit style, moving every step 3 ms earlier or later leaves the sections and notes the same, and every style has more than 3 ms of room (6 to 20 ms). The same test shows that golden timing, with steps on the line, does change the take when steps arrive 3 ms late. `kit_midi_plays_the_script` checks that no event goes out within 5% of a beat of a beat line.
- **B2 (0.2% tempo tolerance hides chord slip): fixed.**
  - The importer now measures how far each chord landed from where we play it. It uses the Chord SysEx and measures relative to the first chord, so the instrument's own delay in reading chords cancels out. Without Chord SysEx, it estimates the slip from the fitted drift.
  - A recording fails verification when any chord slips past the style's margin.
  - Proof: `import_rejects_clock_drift`. Your case, 1.0019 on Austin City Blues, now fails with and without chord messages, while the tempo check alone still passes it. 30 ppm verifies.
  - The fake recording now sends chords on the computer's clock and the style on the instrument's. The old fake sent both on one clock, which is why your throwaway test verified.
- **B3 (RockShuffle is not a Genos preset): fixed.** I replaced it with 90sDisco, a Genos preset with a Tyros5 copy of the same name, which has several source channels per part. `factory_kit_styles_are_genos_presets` checks every factory kit title against the Data List's style names, and checks that `RockShuffle` is not one of them. The test skips when the git-ignored manuals are missing.

**Nonblocking**

- **N1: done.** Reference digests use `Take::render_reference`. Bar lines hold only the script steps, and every note of every part is hashed. That removes both our section timeline and the as-written split, which depends on sections, from the reference. `reference_digest_ignores_our_sections` moves one of our section changes and checks that the golden listing changes but the reference digest doesn't. The readable report still uses our sections for context.
- **N2: done.**
  - The default tolerance is 8 ms.
  - A length snaps only within twice the tolerance, because an end can arrive a burst of traffic late.
  - The alignment line reports the measured jitter (median and 99%) so the tolerance can be tuned.
  - I didn't go down to 5 ms without real data: on a downbeat burst, USB MIDI can serialise dozens of events over a few ms.
- **N3: done.** The script now holds with `-` instead of striking the held chord again. The script test fails if any chord is struck again while held.
- **N4: done, partly by design.**
  - Main A's own fill, the fills into B and D, and the Break now play whole bars: the buttons press between bars.
  - One one-beat fill is kept, into Main A with G7 on its first beat, on purpose.
  - Jack Does It Again has a Fill DD but no Main D, so it falls back to Main C after the fill. The script comment now says so. That fallback is useful hardware data in itself.
  - The ending tail stays at four bars, which covers the longest Ending A in the kit (Ticking Away, 3 bars). The stopped bars after a shorter ending check that the style stops, and the comment says so.
- **N5: done.** `.known` names the recording by file name only. This is tested in `write_reference_files`.
- **N6: done.**
  - `write_reference_files`: `--golden` refuses an unverified recording, `--force` writes it, and the `.known` content is checked.
  - `import_type1_with_notes_before_bar_1`: a type 1 file with the tempo on its own track at another tempo, plus notes before bar 1.
  - `import_rejects_a_missed_section_change`.
  - `import_round_trip` stays as the identity baseline.
- **N7: done.**
  - The README now says that `capture-kit` adds the file table to the kit's copy.
  - Added to the settings: Dynamics Control off, Section Change Timing at Next Bar, Stop ACMP off, all eight style channels on.
  - It now tells owners to take the tempo from the file.
  - The kit, README and forum post are now Genos/Genos2 only. The kit plays keys down to E0, and we don't know whether a 61-key SX reads chord keys it can't play.
- **N8: declined.** The split is needed more now: `perform_steps` is how the kit's timing reaches the engine. The golden digests are unchanged, so any conflict with sibling PRs is textual only and easy to resolve.

The PR body's Decisions section is updated: new decisions for the kit's timing, the chord-slip check, Genos-only and section-free reference digests, plus revised ones for the kit styles, the tolerance and the verification rules. I'm posting the same decisions on #9.

ryangavin:

Reviewed `origin/m2/capture-kit` @ 50d2f8c with the corpus symlinked.

- `cargo test --release`: 216 passed, 1 ignored. All 16 `capture::` tests ran against the corpus.
  - Exception: `factory_kit_styles_are_genos_presets` returns early in a fresh worktree, because `docs/manuals` is git-ignored and missing there (see N1).
- `cargo clippy --release --all-targets`: nothing new. `sim.rs:434` (`map_or(false, …)`) was already there before the PR.
- No RT-thread code is touched and the mixer is not involved. No readable style content is committed.

I ran throwaway probes in `capture.rs`'s test module, then reverted them. Nothing was pushed.

### Round-1 fixes: verified

- **B1 (timing), fixed.** Probe: for every kit style, move every non-button step by ±(margin − 1) ticks. I did this two ways: as a uniform shift, and as a linear drift that grows to margin − 1 by the end of the take. The same sections and notes come out, allowing for notes that start or end with the chord and so move with it.
  - The margin really is the room the take has, so the plan is sound.
  - The closest any chord gets to a beat line is 9.3% of a beat (JackDoesItAgain), and 18–25% for every other style.
- **B2 (drift detection), fixed.** It rejects 0.19% drift with and without Chord SysEx. The capacity problem is a new issue, B2′ below.
- **B3, fixed.** `60s8Beat`, `90sDisco` and `SoulShuffle` are all in `Genos_data_list.txt` (lines 3340, 3376, 3397).
- **N1, N3 and N5, fixed and tested.** N2 is fine with the 8 ms default and the jitter report. N8 was declined with a sound reason, so I'm not raising it again.

### Blocking

**B1′. The "no press in a bar's first beat" guard checks the wrong position** (`src/capture.rs:1019`).
- The script is parsed with tpb = 1920, so ppq is 480 and half a beat is **240**. The guard subtracts **960** (half a bar):
  ```rust
  let t = s.tick.saturating_sub(960);
  assert!(s.tick == 0 || t % 1920 >= 480, ...)
  ```
  So it tests "is the slot on beat 3", not "does the press land in beat 1".
- Throwaway test, using the guard's own predicate:
  - `| C | C [MainB] - - - |`: slot at beat 2, press at beat 1.5, which is inside the first beat and so an immediate change under RM p.12. **Not flagged.**
  - `| C | C - [MainB] - - |`: slot at beat 3, press at beat 2.5, which is harmless. **Flagged.**
- The committed script happens to pass, because its presses are all at bar starts or `@4`. But the one assertion that protects the RM p.12 "within the first beat" rule does not test that rule.
- Fix: use `s.tick.saturating_sub(240)`, or better, check `plan`'s `acts` directly.

**B2′. The drift budget is too small for three of the kit styles, and the test that says 30 ppm is fine checks only the style with the most room.**
- `import_rejects_clock_drift` (`capture.rs:1255`, `:1288`) says "A drift a real pair of crystals shows (30 ppm) is fine", but it tests only AustinCityBlues, where the budget is 19.8 ms over 299 s, about 66 ppm.
- Budgets per style (the plan's margin divided by the take length), with `fake_recording(&ours, 2.0, 1 + ppm·1e-6)` checked on every kit style:

| style | margin | take | budget | 20 ppm | 30 ppm | 50 ppm |
|---|---|---|---|---|---|---|
| SoulShuffle | 6.1 ms | 240 s | ~26 ppm | ok | **not verified** | **not verified** |
| 90sDisco | 6.0 ms | 196 s | ~31 ppm | ok | **not verified** | **not verified** |
| TickingAway | 7.3 ms | 204 s | ~36 ppm | ok | ok | **not verified** |
| the other six | 10.9–20.5 ms | | 54–97 ppm | ok | ok | ok |

- The offset between a computer's clock and an instrument's crystal is fixed for that pair of devices, and ±30–100 ppm between them is ordinary. A second take drifts by the same amount. The failure message (`capture.rs:808`) still says "record again", which cannot help.
- The manual has the fix. RM p.151 says **Clock** can be set to **USB1**, so the instrument's style runs on the DAW's MIDI clock, and the "Clock External" template is described as for exactly this case. With the instrument slaved to the DAW, drift is zero by construction and every margin is available for jitter.
- Suggested fix:
  - Tell owners to set Clock = USB1 and have the DAW send MIDI clock (and check Start/Stop reception, RM p.151; Sync Start behaviour under an external clock should be written up as a Decision).
  - Keep Internal only as a fallback.
  - Make the drift test loop over every kit style, and change the message so it names the clock setting instead of "record again".
- If you keep Internal, at least:
  - fix the false test comment;
  - add the per-style ppm budget to the kit table, so owners of the three tight styles know in advance;
  - record in the Decisions that those styles may not verify on common hardware.

### Nonblocking

**N1. `factory_kit_styles_are_genos_presets` returns early without saying so** when `docs/manuals/` is missing, which is the case in every fresh worktree and in CI.
- The names check out by hand (lines above).
- Print a skip line like the other corpus-dependent tests, or look at `$YAHAHA_MANUALS`.

**N2. The manual doesn't say when a fill starts after a half-beat press.**
- RM p.12 Section Change Timing covers changes *to a Main* (and Auto Fill forces Next Bar). It says nothing about the beat or subdivision a fill (or the Break) starts on.
- `engine::queue_fill` starts it at the next beat. If the Genos starts fills on a finer grid, the "whole-bar" fills (pressed at 4.5) would begin at 4.5 on hardware.
- The drum check would catch this, but a fill is a small share of all drum notes, so the 98% threshold might not.
- Add a Decision stating the assumption, and a report line that compares the drums of each fill bar on their own, so this shows up as a finding rather than an unverified take or a quiet 2%.

**N3. Our engine emits zero-length notes, and many of them.**
- Counts over the full capture script: 367 in 60s8Beat, 330 in SoulShuffle, 269 in 90sDisco, 113 in JackDoesItAgain, 96 in BluesOrganTrio, 57 in SmoothItOver, 10 in AustinCityBlues, 3 in TickingAway, 0 in OrganCruise.
- Example: 60s8Beat, ch15, keys 57 and 62 at tick 337536. The chord `D` acts there and the key gets a note-on and a note-off on the same tick.
- This is engine behaviour that was already there, not this PR. But every one of these will show up as "yahaha only" against hardware, which is unlikely to send zero-length notes.
- Worth a separate M2 ticket, or a line in the import report that counts them, so they are not mistaken for chord-following differences.

**N4. `find_file` duplicates the corpus walker in `sim.rs`.** Minor.

Verdict: **changes requested** (B1′, B2′).

ryangavin:

I reviewed `origin/m2/capture-kit` @ 50d2f8c with the corpus symlinked. This is the same head as round 2: nothing has been pushed since then, and nobody has replied to the round-2 review. So none of the round-2 blocking items have been addressed.

- `cargo test --release`: 216 passed, 1 ignored. The `capture::` corpus tests ran.
- `cargo clippy --release --all-targets`: nothing new in `capture.rs`.
- No readable style content is committed, no RT-thread code is touched, and the mixer is not involved.

### Blocking (carried over from round 2, re-verified on this head)

**B1′. The first-beat press guard tests the wrong position** (`src/capture.rs:1019-1020`).
- The code is `s.tick.saturating_sub(960)` with `% 1920 >= 480`. That subtracts half a bar, when the press actually goes out half a beat (240 ticks) before its slot.
- `| C | C [MainB] - - - |`: the slot is tick 2400, so the guard computes t=1440 and 1440 % 1920 = 1440, and the step is **not flagged**. But the press goes out at 2160, which is inside beat 1. That is an immediate change under RM p.12.
- `| C | C - [MainB] - - |`: the slot is tick 2880, so t=1920 and 1920 % 1920 = 0, and the step **is flagged**. But the press goes out at 2640, in beat 2, which is harmless.
- Fix: use `saturating_sub(240)`, or assert on `plan`'s acts directly.

**B2′. The clock-drift budget is too small for three kit styles.** The test and the error message also contradict each other.
- I re-ran a throwaway probe, `fake_recording(&ours, 2.0, 1 + ppm·1e-6)`, over every `KIT` style on this head:
  - SoulShuffle: 20 ppm verifies. **30 ppm and 50 ppm do not verify.**
  - 90sDisco: 20 ppm verifies. **30 ppm and 50 ppm do not verify.**
  - TickingAway: 20 ppm and 30 ppm verify. **50 ppm does not verify.**
  - The other six styles verify at 20, 30 and 50 ppm.
- `import_rejects_clock_drift` (`capture.rs:1255`) says "A drift a real pair of crystals shows (30 ppm) is fine". It checks only AustinCityBlues, which is the roomiest style.
- `capture.rs:808` tells the owner to "record again". That cannot help: the drift between a given computer and instrument is fixed, so a second take drifts the same way.
- Fix (the same as round 2):
  - Have owners slave the instrument to the DAW's MIDI clock. RM p.151 has Clock = USB1 and the "Clock External" template. Record Sync Start under external clock as a Decision.
  - Or keep Internal, but then:
    - print the per-style ppm budget in the kit table;
    - fix the test comment and loop the test over every style;
    - change the message so it names the clock setting instead of "record again";
    - add a Decision that SoulShuffle and 90sDisco may not verify on common hardware.

### Nonblocking (open from round 2)
- **N1.** `factory_kit_styles_are_genos_presets` returns silently when `docs/manuals/` is missing, which is every fresh worktree and CI. Print a skip line.
- **N2.** Record the fill-start assumption for a half-beat press as a Decision: RM p.12 is silent on it, and `engine::queue_fill` starts the fill on the next beat. Report the drums of each fill bar on their own line.
- **N3.** Our engine emits zero-length notes; 60s8Beat alone has 367 over the script. They will show up as "yahaha only" differences against hardware. Count them in the report, or open a ticket.
- **N4.** `find_file` duplicates the corpus walker in `sim.rs`.

Verdict: **changes requested** (B1′, B2′). Please push fixes and reply to round 2 before the next round.

ryangavin:
Review round 3, addressed in dee1db3.

**Blocking**

- **B1′ (fixed).** The first-beat guard now checks where the press actually goes out: `press_tick(slot) = slot - ppq/2`, then `in_first_beat(t) = t % tpb < ppq`. `plan` and the script test share both helpers. `kit_midi` now also refuses to write a kit that presses in a bar's first beat.
  - Test `first_beat_presses_are_refused` uses your two examples. `| C | C [MainB] - - - |` presses at 2160 and is refused. `| C | C - [MainB] - - |` presses at 2640 and is fine. A press for a downbeat slot (it goes out in beat 4 of the bar before) is also fine.
  - The real script still passes.
- **B2′ (fixed, by publishing budgets plus a correction path, not by clock slaving).**
  - Each chord is now checked against its own room: the distance to the nearest note edge or late-chord point in the sections playing around it, or to its slot. The kit-wide minimum is gone. With that, the budgets are:

    | Style | Budget (ppm) |
    |---|---|
    | 90sDisco | 32 |
    | TickingAway | 38 |
    | SoulShuffle | 43 |
    | OrganCruise | 56 |
    | SmoothItOver | 80 |
    | BluesOrganTrio | 80 |
    | AustinCityBlues | 81 |
    | JackDoesItAgain | 100 |
    | 60s8Beat | 113 |

    The kit README gets a "Clock tolerance" column with these, and the report prints the style's budget next to the drift it measured.
  - The failure message no longer says "record again". It says re-recording on the same setup won't help, and it names the fix, using the drift it measured:
    - `yahaha capture-kit <dir> --clock-ppm N` writes the kit with its tempo N ppm faster, so the chords keep to the instrument's clock;
    - `capture-import --clock-ppm N` imports that recording.
  - `import_clock_drift_over_every_style` covers every kit style, with and without Chord SysEx:
    - at half the budget, the recording verifies;
    - at twice the budget, it fails, and the report gives the measured N (within 3 ppm);
    - a kit corrected with that N verifies again and gives exactly our reference digest.
  - `kit_clock_correction_sets_the_tempo` checks the tempo meta event.
  - Why not clock slaving: RM p.151 says that with Clock set to anything other than Internal, the Style "cannot be controlled by the buttons on this instrument". So Sync Start is gone, and the style would start on the DAW's FA at the top of the file. The lead-in, the Intro press and the first-chord start would all need redesigning, and nothing tested on hardware backs that.

**Nonblocking**

- **N1 (done).** `factory_kit_styles_are_genos_presets` now prints a skip line when the manuals are missing.
- **N2 (done).**
  - The fill-start rule is recorded as a Decision: a fill starts on the beat after the half-beat press, per `engine::queue_fill`, because RM p.12 is silent on it.
  - The report adds a "fill drums" block with one line per fill: its bar, section and start beat, and whether its drums match.
  - `import_round_trip` checks that every fill line reads "same".
- **N3 (done in the report).** The report counts our zero-length notes and says how many of them are "yahaha only". Whether the engine should emit them at all is an engine question outside #9, so I did not change the engine or open a ticket from here.
- **N4 (done).** `golden.rs::find_style` duplicated `capture::find_file` exactly, and now calls it. I left the test-only `corpus()` in sim.rs alone: it lists every style file by extension, which is a different job from finding one file by name.

**Gates**

- `cargo test --release`: 218 passed, 1 ignored (pre-existing). The corpus and docs/manuals were symlinked, so the corpus and Data List tests ran.
- `cargo clippy --release --all-targets`: no new warnings in capture.rs, golden.rs or main.rs. The two main.rs hits are on lines I didn't touch.

The PR body's Decisions and How-tested sections are updated: per-chord room and budgets, `--clock-ppm` versus clock slaving, and fill start.

ryangavin:
## Final adversarial review: **approve**

Reviewed `dee1db3`. I then merged `integration/m2-chord-conversion` @ `44ccb0a` into the branch: #55–#59, #61 (keyboard parts) and #66 (one corpus walker). I pushed two commits: `b40c19b` (the merge) and `42e6330` (fixes). The head is now **`42e6330`**. I did not merge the PR.

### Round-3 blockers: verified fixed
- **B1′ (first-beat guard): fixed.**
  - `press_tick(slot) = slot − ppq/2` and `in_first_beat(t) = t % tpb < ppq` are shared by `plan`, the script test and `kit_midi`.
  - `kit_midi` refuses a press in a bar's first beat.
  - `first_beat_presses_are_refused` covers both of the round-2 examples plus a downbeat slot.
  - The rule matches RM p.12 (Next Bar changes at once "within the first subsequent beat"). AUTO FILL IN forces Next Bar anyway.
- **B2′ (clock budget): fixed.** I checked it end to end through the CLI, not only in unit tests:
  - A 90sDisco take drifting +80 ppm (budget 32 ppm) fails verification. The message names `capture-kit --clock-ppm 80` and says re-recording won't help. `--golden` refuses to write the digest.
  - A `--clock-ppm 80` kit, recorded on the same +80 ppm clock and imported with `--clock-ppm 80`, verifies with 0 of 102 bars differing.
  - `import_clock_drift_over_every_style` loops over all nine styles.
  - The kit table prints each style's tolerance: 32 ppm (90sDisco) to 113 ppm (60s8Beat). These are unchanged by the merge.
  - I agree with not using clock slaving. RM p.151 says that with an external clock the Style cannot be controlled from the panel, which rules out Sync Start.

### The merge
- **Conflicts:**
  - `README.md` and `main.rs` keep both the capture commands and the oracle command.
  - `golden.rs` takes #66's `find_style`.
- **Guard fix:** #66's `only_the_library_walks_folders` would have failed on `capture.rs` (three hand-rolled folder listings). So:
  - `find_file` now goes through `library::style_files`;
  - the two `tests/reference` checks use a new test-only `library::files_in`.
- **Sim still lines up after the merge:**
  - The kit `.mid` files are byte-identical before and after the merge.
  - Every kit style's reference digest still equals `import_round_trip`'s.
- **Channels:**
  - The importer keeps only the recording's ch 9–16. The keyboard parts on ch 1–4 (#61, All Parts template) carry the Left voice echoing the kit's keys, and are ignored.
  - Our side was not filtered by channel. `import_round_trip` now asserts that our take plays only on ch 9–16, so a future engine change that makes keyboard parts sound in sim can't silently skew the matching.

### End to end, from the owner's side
- **Kit contents:** `yahaha capture-kit` writes 9 files plus the README. Each file holds only:
  - ch 1 chord keys (MIDI 28–54, below the default F#2 split);
  - Section Control SysEx on/off pairs;
  - bar markers, tempo and time signature.

  There is no style data.
- **Round trip:** for every kit style I wrote a fake recording from our own sim take and imported it through the CLI (`capture-import`). Each one was:
  - type 1, recorded at another tempo;
  - 3 ms jitter and a +15 ppm instrument clock;
  - notes before bar 1;
  - Chord SysEx and the ch 4 keyboard echo.

  All 9 came out **verified, 0 of 102 bars differ**, with 140 of 140 chords read as meant.
- **Nothing readable committed:** no readable style content is committed, and `tests/reference` holds only its README. The one `/Users/` string in the diff is the test fixture that checks a local path is stripped.
- **Owner instructions:** I checked the MIDI settings against RM pp.150–153 (All Parts template, Receive "Keyboard", Chord SysEx Tx/Rx, Clock Internal, Start/Stop). A Genos owner following the README should produce a recording the importer can align: it finds bar 1 from the drums and ignores notes before it and anything on ch 1–8.

### Blocking issue found and fixed in `42e6330`: recordings were to be posted publicly
- **The problem:** the kit README said "Post them in the forum thread, or attach them to an issue", and that a recording "is yours to share". The forum post said "reply here … or attach them to a GitHub issue".
  - A recording is a note-for-note performance of the style (Yamaha's, for the three presets).
  - This PR's own `tests/reference/README.md` keeps recordings off the repo for exactly that reason.
- **The fix:**
  - Both docs now ask for recordings by private message, not in the thread or an issue, and say why.
  - The kit never asks anyone to share style files: owners load the free MOX pack or a preset themselves.
  - `owner_docs_ask_for_recordings_privately` pins this.
- Please add a line to the PR's Decisions about sending recordings privately. I did not edit the PR body.

### Gates (at `42e6330`)
- `cargo test --release`: **276 passed, 0 failed, 1 ignored**.
  - All 20 `capture::` tests ran against the corpus, and the Data List check ran with the manuals present.
  - Three synth tests print "soundfont missing; skipping" on this machine, because `soundfonts/GeneralUser-GS.sf2` is renamed to `.bak` locally. That is unrelated to this PR.
- `cargo clippy --release --all-targets`: nothing new against integration @ `44ccb0a`. One warning is gone: `library::style_files` is now used.

### Remaining concerns (nonblocking)
- Every round trip here is self-consistent: fake recordings built from our own take. The first real Genos recording is still the real test. Things only hardware can confirm:
  - fill start after a half-beat press (the N2 Decision);
  - Sync Start with a chord that arrives a lead early;
  - whether "Keyboard" receive echoes on ch 1–4 as assumed. The importer tolerates it either way.
- The forum post still has the `[Ryan / contact]` placeholder, as intended, for Ryan to fill in before posting.
- N3 (zero-length notes from our engine) is only counted in the report. It still wants an engine ticket.

Issue #49, opened 2026-09-23
Zero-length retrigger notes at chord changes


ryangavin:
Found in review of #45. When a chord change retriggers a note whose pattern note-off falls on the same tick, a 0-length note is emitted (an audible blip). The golden listings showed about 72 of them across 6 styles. Fix this alongside the RTR work in #10.

---
Milestone: **M2 · Chord-to-note conversion 1:1**

ryangavin:
Decision made in #63 (the owner may overrule):

- **Decision: a note whose pattern note-off falls on the chord change is left to end. It is not retriggered or bent, because a new attack there would last zero ticks.**
  - When the section boundary itself falls on the change, nothing is revoiced, because the boundary's all-off cuts every note anyway.
  - `catch_up` uses the same test, so it no longer starts such notes either.

The new sim test asserts that no note on a chord-following part lasts zero time across the corpus.

ryangavin:
Decision updated in #63, review round 1 (the owner playtests and may overrule):

- **Decision: a chord that lands on, or up to 40 ms (`EARLY_CHORD_NS`) before, a note's pattern note-off does not attack that note again. The same applies before the pattern strikes the note's new pitch again, and before the section ends. Where the note would be retriggered, it plays out at its old pitch, or stops if the part's bend moves. Reasons:**
  - a new attack there is a blip;
  - 40 ms mirrors the existing late-chord window;
  - a chord just before the beat belongs to the next bar.

  A note brought in by a late chord starts only if no more of it is lost than is still to come.
- **Evidence:** with the window off and every corpus style played with chords 1 ms before the beat, 1,208 retriggers last under 40 ms (5,488 at 20 ms early). With it on, none do.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #46, opened 2026-09-23
Revoice can send two note-ons for the same key; first note-off cuts both


ryangavin:
When a chord change makes two voices of one part land on the same pitch, `Engine::revoice` sends a second note-on for a key that is already sounding on that channel. `note_on` steals identical notes when *starting* a note, but `revoice` bypasses it. On/off counts stay balanced, but on external gear (and most synths) the first note-off releases both voices, so the part goes quiet early.

This already happened with chords that fold voices, but 1+8 (#4, PR #43) folds every voice to the root, so it is now much more common: in the corpus CoolRevibed goes from 4 to 10 cases and SlowWalker from 2 to 7.

Suggested fix: in `revoice`, when the target key is already sounding on the same dest channel, drop (release) the duplicate voice instead of sending a second note-on, and keep the sounding table consistent. Add a sim test that asserts no channel ever has two note-ons for one key without a note-off between them.

Found in review of PR #43.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

ryangavin:
Decision made in #63 (the owner may overrule):

- **Decision: two voices of a chord-following part landing on one key keep one, because a second note-on for a sounding key is cut by the first note-off on most synths.** How that plays out:
  - On revoice, a retriggered voice whose key is already sounding is dropped. A note that is ending on that tick makes way instead.
  - The same thing happens outside revoice when two voices start on one key at the same instant (1+8 folding, or a pattern note on a key the chord change just retriggered). The later voice takes over the sounding note, so its note-off ends it. Before, `note_on` stole the first voice and left a 0-tick note.
  - Rhythm parts are left as written, so doubled drum hits stay doubled.

The new sim test `corpus_chord_changes_leave_no_blips_or_doubled_keys` asserts this across the whole corpus. It changes chord every half bar with 1+8/1+5.

ryangavin:
Decision updated in #63, review round 1 (the owner playtests and may overrule):

- **Decision: two voices of a chord-following part that land on one key sound once, and the second is kept as a muted twin. It sends nothing, but the key sounds until both voices have ended, and the next chord parts them again. For example, 1+8 → major brings back the third and fifth. Reasons:**
  - a second note-on for a sounding key is cut by the first note-off on most synths;
  - on Genos both voices exist inside the tone generator.

[...]

PR #63, opened 2026-09-23
RTR: real Pitch Shift, Note Generator, and revoice fixes (#10, #46, #49)

ryangavin:
## What

This PR reworks the chord-change revoice path in `src/engine.rs`.

**Pitch Shift / Pitch Shift to Root now bend (#10).** A sounding note moves to its new pitch with the part's pitch bend, with no new attack.
- Each part that follows chords (ch 11–16) gets a pitch bend range of at least 12 semitones. It is sent as RPN 0 at the end of the style's channel setup. A part whose patterns bend on their own gets 12 more than its widest pattern bend, up to 24, and never shifts by more than that leaves over the pattern's bend, so the two together never clamp (review B2).
- The style's own RPN 0 (in SInt or in a pattern) is kept as the *source* range. The pattern's own bends (slides) are rescaled from it, so they sound as written. When a style already sets 12 or more, its bends go out byte for byte.
- Pitch bend is per channel, so every note on a part bends together. The part bends by the shift that most of its continuing notes need, and a held note counts as a vote for no shift.
  - A Pitch Shift note that needs a different shift is retriggered at its new pitch.
  - A held note that the chosen bend would detune is also retriggered.
- A note started while its part is bent goes out that much lower, so it sounds at the right pitch.
- The bend returns to centre at the next note-on once the part has fallen silent, also when no note continues through a change, so release tails keep their pitch. It also resets on stop, section change and style change (`all_off`).
- **Note Generator** plays as Retrigger.

**Revoice fixes (#46, #49).** `revoice` now works one part at a time, in phases: note-offs, then the bend, then note-ons.
- **#46:** when a retriggered voice lands on a key that is still sounding, it gets no second note-on: it joins the key as a muted twin. If the note on that key is about to end, it gives way instead.
- **#46 (same cause, outside revoice):** two voices of a chord-following part that start on one key at the same instant now sound once. This covers 1+8 folding and a pattern note landing on a key the chord change just retriggered. Before, `note_on` stole the first voice and left a 0-tick note. The second voice is kept muted beside the first: it sends nothing, but the key sounds until both voices have ended, and the next chord parts them again (for example 1+8 → major brings back the third and fifth). Rhythm parts are unchanged, so doubled drum hits stay doubled.
- **#49:** a chord that lands on, or up to 40 ms before, a note's pattern note-off does not attack that note again: where it would be retriggered, it plays out as it is (or stops, if the part's bend moves). The same holds when the pattern strikes the retriggered pitch again within 40 ms, and when the section ends within 40 ms (nothing is revoiced; `all_off` cuts everything anyway). Notes about to end have no vote on the part's bend. `catch_up` starts a note only if no more of it is lost than is still to come.

**Sim listing.** `Sink` gained a default no-op `retune(ch, semis)` hook, which only the `Recorder` implements.
- The listing now shows the pitch that sounds, and each Pitch Shift as `beat.tick C2>F2`. The format is documented in `tests/golden/README.md`.
- `Recorder::pitch_ons` and `Recorder::sounding_at` give tests the sounding pitch.

Real time: nothing new allocates, locks or panics. Revoice uses fixed stack arrays sized `MAX_SOUNDING`, and the bend math is a few float operations.

## Corpus evidence

RTR usage across the 208 corpus styles (zones; SFF2 counts all three zones):

| RTR | zones | styles |
|---|---|---|
| Pitch Shift | 18,992 | 206 |
| Pitch Shift to Root | 1,909 | 196 (1,291 of them on Root Trans bass) |
| Retrigger | 371 | 36 |
| Retrigger to Root | 487 | 38 |
| Stop | 672 | 32 |
| Note Generator | 0 | 0 |

Style bend ranges (RPN 0 data entry):

| Where | Ranges set |
|---|---|
| SInt | 12 ×70, 24 ×16, 8 ×8, 4 ×2, 3 ×2, 0 ×3 |
| Patterns | 12 ×6, 2 ×4 |

132 styles bend in their patterns (32,771 bend events).

I ran a performance with a chord change every half bar across the corpus: the sim test script, with chords both on and off the beat.
- 9,159 bends were sent across 203 styles.
- Pitch Shift notes at the changes: 15,797 bent without a new attack, and 11,140 fell back to a retrigger because another note on the part needed a different shift. 526 held notes were re-attacked because of the bend.
- New attacks at chord changes on ch 11–16 fell from 42,303 (everything retriggered) to 27,119.

## How tested

All gates pass:
- `cargo build --release`
- `cargo test --release`: 209 passed, 1 ignored (the existing M1 row). The corpus was linked and the corpus tests ran.
- `cargo clippy --release`: no new warnings. The two in `engine.rs`/`sim.rs` under `--tests` are older code.

New tests from review round 1:
- `corpus_early_chords_start_no_blips`: every style, chords 1, 3, 20 and 39 ms before the beat. No note a chord change retriggers lasts less than 40 ms. With the window turned off, there are 1,208 such notes at 1 ms early, 1,450 at 3 ms and 5,488 at 20 ms.
- `chord_a_hair_before_the_beat_retriggers_nothing_that_ends_on_it`: the reviewer's Merengue repro, with the bar length truncated to whole ns, 1 ns early and 1 ms early. It failed before the fix (ch12 key 69 was a blip).
- `corpus_pitch_shift_leaves_room_for_pattern_bends`: a busy performance (a chord change every quarter bar, any root and type, slash chords, Main A–D; 3 seeds) through every style, run with a test-only clamp counter. There are 0 clamped bends. With the headroom check off, it clamps in 4 styles (AnalogBallad, ElectroRock, EuroPopMambo, Reggaeton2). It also checks that AnalogBallad's octave-sliding bass goes out with RPN 0 = 24.
- `folded_voices_part_again`: 4 pad-heavy styles, with every zone set to Retrigger, going C1+8 → C. The notes held through the change match a run in C all along, including E and G. It fails without the muted twins.
- `engine::tests`: NRPN deselects RPN 0, and the output-range formula.
- `late_chord_after_no_chord_keeps_the_downbeat` now leaves out keys the on-beat run strikes or releases within the lateness after the check point. `corpus_pitch_shift_sounds_what_retrigger_would` leaves out notes about to end or be struck again. The two runs treat those differently on purpose.

New tests (`src/sim.rs`, `mod rtr`):
- `corpus_chord_changes_leave_no_blips_or_doubled_keys` runs every corpus style through Main A–D, changing chord every half bar on and off the beat, with 1+8, 1+5 and slash chords. On the chord-following parts:
  - no key gets a second note-on while it is sounding (#46);
  - no note lasts zero time (#49).

  On every part:
  - note-ons and note-offs balance;
  - nothing is left sounding;
  - no part is left bent or pitch-shifted.
- `corpus_pitch_shift_sounds_what_retrigger_would` plays every style twice: as written, and with Pitch Shift (to Root) swapped for Retrigger (to Root). It checks that:
  - for each chord pair, the same pitches sound right after the change;
  - through the whole performance, every note started between changes sounds at the same pitch in both runs.

  This covers the bend arithmetic, the vote, the fallback retriggers and the compensated note-ons.
- `pitch_shift_bends_the_held_bass` uses TickingAway's 4-bar bass note. It checks:
  - RPN 0 is set to 12;
  - G1+8 → D1+5 sends no note-off/on, only the exact −5 semitone bend value;
  - D1 sounds;
  - the next note after the chord change to C is sent straight, with the bend recentred.
- Existing tests now look at the sounding pitch: `ons`/`held_at` and the 1+8/1+5 test, which also checks notes held into the window.

## Golden changes (review round 1)

Regenerated again with `UPDATE_GOLDEN=1` on top of the round-0 digests. SmoothItOver and AustinCityBlues are unchanged.

| Style | Bars (parts) | What changed |
|---|---|---|
| BluesOrganTrio | 15, 16, 18, 23 (Chord1) | At the mid-bar change the Chord1 notes end 128 ticks (≈33 ms) after the chord. They used to be retriggered for those 128 ticks; now they play out at their old pitch (#49 window). |
| BluesOrganTrio | 25–27, 29–31 (Bass) | The same for the beat-4 bass note over the bar line: it plays out its last 128 ticks instead of a 128-tick retrigger in the next bar. |
| BluesOrganTrio | 28 (Chord1), TickingAway 27 (Chord2) | Two voices folded on one key: the key now sounds until the later of the two ends (muted twin), 32 and 75 ticks longer. |
| JackDoesItAgain | 25–27, 29, 30, 33, 34, 37–39 (Chord1) | 24–28-tick retriggers on the downbeat are gone; the note from the bar before plays out those ticks instead (#49 window). |
| JackDoesItAgain | 10 (Chord2) | A note whose note-off falls 56 ticks after the bar-line change plays out instead of ending at the change. |
| JackDoesItAgain | 3 (Chord1), 4, 8, 25, 27, 39 (Chord2) | Folded voices: the key now sounds until the later voice ends (20–120 ticks longer). |
| JackDoesItAgain | 16 (Chord1, Chord2) | Asus2 → A: Chord2 now sounds A2 C#3 E3. The voice folded onto E3 under Asus2 is kept, so the group's voice leading is whole again and B2 moves up to C#3. Before, it played E2 and A2 and lost the third. Chord1: an A3 plays out its last 80 ticks. |
| JackDoesItAgain | 13 (Chord1) | Same notes and lengths. Only the order of three note-ons sent at the same instant changed, and the listing prints in send order. |
| JackDoesItAgain | 17 (Chord1) | An 80-tick G3 retrigger on the C1+5 downbeat is gone (#49 window). |
| OrganCruise | 3–40, 22 bars (Bass) | The bass note from beat 4 ends 64 ticks (≈17 ms) after the bar line. It no longer bends the whole part for that moment, so the `A0>F0`-style bends are gone. Where the next bass note (just before the bar line, corrected outright) lands on the same key, the old note makes way at the bar line (for example 1108 → 1044 ticks). |

## Golden changes (round 0)

I regenerated with `UPDATE_GOLDEN=1`. SmoothItOver is unchanged. The other styles changed by bar group:

| Style | Bars (parts) | What changed |
|---|---|---|
| AustinCityBlues | 26–27, 30–31, 38–39 (Bass) | A bass note held over the bar line now bends to the new chord's root (Pitch Shift to Root) instead of being cut and struck again. The held note is longer and gets a `X>Y` bend. |
| BluesOrganTrio | 3, 5, 7, 11–15, 17, 23, 26, 28, 30, 31, 33, 35, 38, 40 (Chord1) | Zero-length notes are gone. Two Chord1 voices landing on one key on the same tick used to be stolen into a 0-tick note plus the real one; they now sound once. |
| BluesOrganTrio | 18 (Chord1) | A doubled note-on from revoice was dropped (#46). |
| BluesOrganTrio | 27–32 (Bass) | Zero-length bass retriggers on the downbeat were removed; the pattern's note-off fell on the change (#49). |
| JackDoesItAgain | 2–39, 30 bars (Chord1/Chord2) | Mostly zero-length notes on the chord change removed (#49) and doubled voices dropped (#46), for example bar 3. |
| JackDoesItAgain | 13, 18, 23 (Chord2) | Held voicings now hold or bend through the mid-bar chord change instead of being cut and restruck. |
| OrganCruise | 3–40 (Bass) | Bass notes held into the next bar now bend to the new root instead of retriggering. A few 0-length retriggers were removed. |
| TickingAway | 2–41 (Bass, Chord1, Chord2) | This style holds long notes: a 4-bar bass, whole-bar chords and pads. They now bend to follow each chord instead of being restruck every bar. Parts whose notes need different shifts bend by the majority shift and retrigger the rest. |

## Decisions

- **Decision: Pitch Shift bends with MIDI pitch bend. Each chord-following part (ch 11–16) gets an output bend range of max(style's own, 12, min(widest pattern bend + 12, 24)) via RPN 0, because:**
  - bend is the only way to move a sounding note without a new attack on a MIDI channel;
  - 12 covers every to-Root move (±6) and nearly every Pitch Shift move;
  - 12 is the range styles set most often (70 SInt settings);
  - a Genos Style part takes RPN 0 up to 24 (DL p.98, Pitch Bend Sensitivity 00H–18H; MIDI Implementation Chart 0–24), and 16 SInt settings use 24.

  The pattern's own bends are rescaled from the style's range, and a shift that would not fit over the pattern's widest bend, in the narrowest range the part can have (a pattern may set a narrower RPN 0 than the SInt), falls back to a retrigger. The RPN 0 is sent with the part setup, so a section change (which re-sends the SInt, #15) sets it again, and SInt bends re-sent there, or controls a mid-bar Fill catches up on, are rescaled the same way. (Round 0 cited RM p.140 for "12 is the Genos maximum". That page is the keyboard parts' wheel range, not style parts, so the citation is withdrawn.) External instruments that ignore RPN need their range set by hand: 12, or 24 to be safe. The README says so.
- **Decision: when the notes on one part need different shifts, the part bends by the shift the most continuing notes need, and the others are retriggered at their new pitch, because:**
  - pitch bend is per channel;
  - the manual doesn't say what a MIDI stream would carry;
  - this keeps every pitch right and attacks as few notes as possible.

  Ties go to the smaller shift. A note about to end (next decision) has no vote, and a key two voices share (a muted twin) votes once.
- **Decision: a chord that lands on, or up to 40 ms before, a note's pattern note-off does not attack that note again. The same goes for the section end, and for the pattern striking the note's new pitch again within 40 ms. Where the note would be retriggered, it plays out at its old pitch, or stops if the part's bend moves. Reasons:**
  - a new attack there is a blip (#49, review B1);
  - 40 ms mirrors the existing late-chord window (`LATE_CHORD_NS`);
  - a player landing a chord just before the beat means the next bar, not the tail of this one.

  A note brought in by a late chord (`catch_up`) starts only if no more of it is lost than is still to come. That keeps the downbeat for a chord a little late, and drops only notes that would be mostly gone.
- **Decision: two voices of a chord-following part that land on one key sound once, and the second is kept as a muted twin. The key sounds until both have ended, and the next chord parts them again (#46, review). Reasons:**
  - a second note-on for a sounding key is cut by the first note-off on most synths;
  - on Genos both voices exist inside the tone generator;
  - keeping the twin makes 1+8 → major bring back the third and fifth.

  A voice landing on a sounding key at a different moment still steals it, as in M1. Rhythm parts play doubled hits as written.
- **Decision: a bent part returns to centre at its next note-on once it has fallen silent, not at the last note-off, and it keeps its bend when nothing continues through a change. This keeps the release tail at its pitch.** (Round 0 sent centre right after the note-offs when nothing continued, which contradicted this. Fixed.)
- **Decision: Pitch/Retrigger to Root goes to the slash bass on channels that follow the slash bass (Bass On, or an NTT Bass zone, the same test `theory::transpose` uses) and to the chord root elsewhere.** Since #57 this reads the zone's Bass On (`zone.bass_on`), as `theory::transpose` does.
- **Decision: to-Root keeps the nearest root (±6), as before this PR, rather than a literal "same octave" (RM p.31: "The octave of the new pitch remains the same"). Reasons:**
  - the manual doesn't say whether "octave" means C to B or the note's own register;
  - the nearest root never moves a note more than a tritone, which is always within the note's octave in the second sense;
  - a C-based reading would jump an 11-semitone bend or retrigger on B → C.

  The owner may overrule.
- **Decision: Note Generator plays as Retrigger, because:**
  - it is not in the Genos RTR list (RM p.31);
  - no corpus style uses it (0 zones);
  - its name suggests a new attack.
- **Decision: an NRPN select (CC 98/99) deselects RPN 0, so a later data entry is not read as a bend range (0 corpus cases; the MIDI spec).**

Closes #10
Closes #46
Closes #49


ryangavin:

Checked out `origin/m2/rtr-revoice` (6766a43) with the corpus linked. I ran:
- `cargo test --release`: 203 passed, 1 ignored. The corpus tests did run: `sim::rtr::corpus_*`, `corpus_full_performance_no_stuck_notes` and `theory::tests::corpus_every_rule_every_chord` all executed and passed.
- `cargo clippy --release`: no new warnings in the lines this PR touched.

I proved the findings below with throwaway sim tests (not pushed). **Verdict: changes requested (2 blocking).**

### Blocking

**B1. #49 is only fixed when the chord lands exactly on the tick. A 1 ns early chord still gives a 1 ns note, and the new test can't catch it.**
- `due_now` (`src/engine.rs:1257`) uses `target = tick_at(now) + 1e-6` ticks. A chord that lands a hair before the tick of a pattern note-off doesn't see that note-off. The note gets retriggered, and the pattern ends it a moment later.
- **Repro (corpus, no offset at all):** Merengue.T158.prs. Chords are placed at `i * bar_ns`, where `bar_ns` is truncated to 1846151999 ns.
  - At t=1846151999, ch12 keys 65 and 69 are retriggered.
  - Their note-offs arrive 1 ns later.
  - That is a zero-length note in every audible sense.
- **Live play is worse.** In live play the player usually strikes the chord just before the beat. Across the corpus, with chords every bar and a C/F/G/Am cycle:
  - 1 ms early: 349 retriggered notes lasted ≤1 ms, in 35 styles (shortest 23 µs).
  - 3 ms early: 523 notes lasted ≤3 ms, in 48 styles.

  These are exactly the "audible blip" #49 describes.
- **The test is too weak.** `corpus_chord_changes_leave_no_blips_or_doubled_keys` asserts `s < *t` (`src/sim.rs:756`), so a 1 ns note passes. Its chords are also either exactly on the beat or `bar/11` late, never slightly early.
- **Fix:** give `due_now`/`ends_now` a tolerance. Treat a pattern note-off (or a section end) due within a small window after `now` as ending now: "leave it". A few ms is enough; an `EARLY_CHORD_NS` next to `LATE_CHORD_NS` would do.

  Then assert a minimum length, for example no retriggered note shorter than that window, with chords a few ms before the beat as well as after it.

**B2. The pattern's own bends and the RTR shift can together go past the output range, so the pitch sent is clamped and wrong.**
- `send_bend` (`src/engine.rs:~1219`) clamps `centre + pat + rtr` to 0..16383. The vote (`src/engine.rs:1490`, `(bend + d).abs() > range`) checks only the RTR shift. It ignores the headroom the pattern's own bends need.
- For a style whose range is already ≥12, `out_bend_range` adds no headroom at all.
- **Repro:** the half-bar chord-change run over the corpus. I instrumented the clamp:
  - 59 clamped bends in 6 styles.
  - Example: TickingAway ch11 (bass). The style range is 12 and the pattern slides the bass to −12. The part is already shifted −2/−3/−9, so the pitch stops at −12 when it should reach −14/−15/−21.
  - Also: CountryPopDuo ch15, ch13.
- On Genos the shift is internal, so the style's slide always sounds as written. This is a fidelity error that the Decision's claim that "12 covers nearly every Pitch Shift" does not account for.
- **Options:**
  - (a) At load time, work out each part's maximum pattern bend excursion. Refuse a bend (fall back to retrigger) when `|rtr| + max_pat` exceeds the range.
  - (b) Send a wider output range, such as style range + 12 capped at 24, and rescale.

  Either way, add a test that no bend is clamped while a part is shifted.

### Nonblocking

1. **Wrong citation.** "12 is the Genos maximum (RM p.140)" (the PR body, the `RTR_BEND_RANGE` doc at `src/engine.rs:18-21`, and the #10 comment).
   - That page is the *keyboard part* Pitch Bend Range (Left/Right1-3) for the wheel and Glide.
   - It is not a limit on style parts: 16 corpus SInt settings use 24, and the code keeps them.
   - Please fix the wording.
2. **Release tails get bent.** Cut and Retrigger notes get their note-off, then the new channel bend (`src/engine.rs:1501`).
   - When nothing continues, `new_bend = 0` is sent straight after the cuts. That is Chord Cancel, all-Stop parts, and parts that were previously bent and are now all-Retrigger.
   - So the release tails of the notes just released jump by the old bend.
   - This contradicts the PR's own Decision that the bend recentres at the next note-on "because the release tail keeps its pitch". When `!continuing` and nothing is retriggered, leave the bend until the next note-on.

   The same applies to "Leave" notes. Their pattern note-off is processed after `revoice` has already sent the new bend, so their tail is bent too.
3. **#46 dedupe loses voices for later chord changes.** When two voices merge on one key (the `note_on` hijack at `src/engine.rs:1198`, and the dropped retrigger voice), the dropped voice is gone from `sounding`.
   - Example: after 1+8 → C major, with Pitch Shift, Genos would split the merged voices back into root and fifth. yahaha moves only the one survivor.
   - Consider keeping the dropped voice as a "shadow" entry (active, no note-on sent) so a later revoice can bring it back.
   - The equivalence test can't see this, because both of its runs drop the voice.
4. **Scope and conflict with #57.** The to-Root target now uses `rule.bass_on` only (`src/engine.rs:1455`). The transposer (`theory.rs:597`) uses `rule.bass_on || z.ntt == Ntt::Bass`.
   - Please use the same predicate.
   - #57 (per-zone Bass On) is changing exactly this, so this bit will conflict. Either drop the change here or coordinate with #57.
5. **Pitch Shift to Root octave.** The manual says "The octave of the new pitch remains the same" (RM p.31). The code keeps the nearest ±6. The comment even says "in the same octave or the next".
   - This predates the PR, but the Decision says the rule was kept without citing the manual line.
   - Please record it as a Decision with a reason, or match the manual.
6. **RPN tracking ignores NRPN.** `Prepared::new` and `pattern_rpn` track only CC101/100. A CC6 after NRPN 99/98 (with RPN 0 still latched) would be read as a bend range and rewritten.
   - It happens 0 times in the corpus today, and XG SInt does use NRPN. Clear the tracked RPN on CC98/99.
7. **Corpus counts.** I reran them. I get Pitch Shift 19,128 zones / 206 styles, and to Root 1,929 / 196, counting SFF1 rules as 3 zones. The rest match exactly.
   - The small difference is only SFF1 zone counting. The conclusions stand.
   - Note Generator: 0 zones confirmed, and it is not in the RM p.31 RTR list. That Decision is fine.

Checked and fine:
- Mixer principle: no gain or CC7 changes.
- RT safety: fixed stack arrays, no alloc, lock or panic path in `revoice_part`, `due_now`, `ends_now` or `send_bend`.
- Golden changes are digest-only, and no readable note content is committed.
- `all_off` recentres the bend on stop, section change and style change.

ryangavin:
## Review round 1: addressed in 9cccff1

**B1 (#49 only fixed exactly on the tick): fixed.**
- `due_now` now looks `EARLY_CHORD_NS` = 40 ms ahead. This mirrors `LATE_CHORD_NS`.
- A note whose pattern note-off falls in that window is not attacked again. Where it would be retriggered, it plays out at its old pitch, or stops if the part's bend moves.
- The same holds when the pattern strikes the retriggered pitch again inside the window (new `struck_now`), and when the section ends inside it.
- Notes about to end have no vote on the part's bend.
- `catch_up` starts a note only if no more of it is lost than is still to come. That keeps the downbeat for a slightly late chord.
- Tests:
  - `corpus_early_chords_start_no_blips` runs every style with chords 1, 3, 20 and 39 ms early. No retriggered note lasts under 40 ms. With the window off, 1,208 do at 1 ms early and 5,488 at 20 ms.
  - `chord_a_hair_before_the_beat_retriggers_nothing_that_ends_on_it` is your Merengue repro: 2 × bar in whole ns, minus 1 ns, minus 1 ms. It failed before the fix (ch12 key 69).

**B2 (pattern bend + shift clamped): fixed.**
- `Prepared` precomputes each part's widest pattern bend (`pat_bend_max`), in semitones of the widest range the style gives the part.
- The output range becomes max(style, 12, min(pat_max + 12, 24)). A Genos Style part takes up to 24: DL p.98, RPN 0 00H–18H.
- The vote refuses a shift beyond range − pat_max, and such notes fall back to a retrigger.
- Test: `corpus_pitch_shift_leaves_room_for_pattern_bends` counts clamped bends through a test-only counter in `send_bend`, over a busy performance (a change every quarter bar, any root and type, 3 seeds) × every style.
  - Result: 0 clamps. With the headroom check off, it clamps in AnalogBallad, ElectroRock, EuroPopMambo and Reggaeton2.
  - AnalogBallad's bass, which slides a full octave, now goes out with RPN 0 = 24.
  - TickingAway's bass has no pattern bends (its range is the GM 2). Chord1 bends ±1, so it gets 13.

**Nonblocking**
- RM p.140 citation: fixed. It is withdrawn in the `RTR_BEND_RANGE` doc, the PR body and the #10 comment. The 24 cap cites DL p.98 and the MIDI Implementation Chart.
- Centre sent right after the note-offs when nothing continues: fixed. The bend is kept until the part's next note-on after it falls silent, which matches the Decision. Notes that would be left bent out of tune are cut rather than bent.
- #46 dedupe dropping the merged voice: done. The second voice is kept as a muted twin (`Sounding::muted`).
  - It sends nothing, and `off_where` hands the key to the twin if the sounding voice ends first.
  - `revoice_part` swaps roles when the sounding voice moves but the twin stays, and lets a twin retrigger onto its own key.
  - Test: `folded_voices_part_again` (C1+8 → C with Retrigger on 4 pad styles). The held notes match a run in C all along, and it fails without twins.
  - Scope: this covers voices that meet at the same instant. A voice that lands on a sounding key at a different moment still steals it, as in M1. I tried turning those steals into twins too. It moved 31 more golden bars (AustinCityBlues and more of Jack), so I left it out as a separate behaviour change. Tell me if you want it.
- To-Root target: it now uses `rule.bass_on || zone.ntt == Ntt::Bass`, like `theory::transpose`. #57 will touch both places.
- Pitch Shift to Root, nearest ±6 versus "octave remains the same": declined as a code change and recorded as a Decision. The manual doesn't define the octave boundary, and the nearest root never moves more than a tritone. It predates this PR.
- NRPN: fixed. CC 98/99 deselect RPN 0 in SInt parsing, in the precompute and in patterns (`select_rpn`), with a unit test.
- RTR counts and Note Generator: nothing to change.

**Golden changes:** BluesOrganTrio, JackDoesItAgain, OrganCruise and TickingAway were regenerated. Every bar group is explained in the PR body under "Golden changes (review round 1)". Most are the 40 ms window (short retriggers gone, notes play out their last ticks). A few are twins keeping a key sounding until the later voice ends. One (Jack bar 16) is a twin restoring a lost third.

**Decisions updated** in the PR body: bend range and citation, the 40 ms window, muted twins, keeping the bend, to-Root on NTT Bass, nearest-root octave, and NRPN.

**Gates:** `cargo build --release` passes. `cargo test --release` gives 209 passed and 1 ignored, and the corpus tests ran. `cargo clippy --release` shows no new warnings; the one in `engine.rs` is older Break-button code.

ryangavin:

I checked out `origin/m2/rtr-revoice` (9cccff1) with the corpus linked. Results:
- `cargo test --release`: 209 passed, 1 ignored. The corpus tests ran (`sim::rtr::corpus_*`, `corpus_full_performance_no_stuck_notes`, `theory::…corpus_every_rule_every_chord`).
- `cargo clippy --release`: nothing new in the touched code. `engine.rs:991` is the old Break code.

I checked the round-1 fixes with throwaway sim tests, none of them pushed.

**Verdict: changes requested (1 blocking).** The round-1 fixes themselves hold up. The blocker is that the PR can't be merged into its base as it stands, and the conflict affects how it behaves.

### Round-1 fixes I checked
- **B1 (early chords).** `due_within`/`ends_now`/`struck_now` do the job.
  - I ran a chaos run: every corpus style, 2 seeds, 47 changes each. The changes fell up to 40 ms before or after the beat, or anywhere in the bar. They included Chord Cancel, slash chords, any chord type, a second change 5 ms after some of them, part toggles, Manual Bass, Keyboard/Master transpose and Main changes.
  - Result: no zero-length notes, no second note-on on a sounding key (ch 9–16 pitched parts), no stuck notes, and no part left bent.
  - I also ran chords every half bar, 1, 10 and 30 ms late. The only notes under 5 ms were 2 in 80sSynthDisco. That style writes the same ~5.5 ms restrike itself, and the on-beat run has it too, so those are style content.
- **B2 (clamps).** `bend_clamps` stayed at 0 through the whole chaos run, transposes included.
- **Muted twins.** In the chaos run they never leaked a note-on, never doubled a note-off (`all_off` goes through the twin logic, so `!f(o)` skips the handoff), and never left anything stuck.
- **Other round-1 items.** The NRPN deselect, the to-Root predicate, the citation fix, and keeping the bend on a silent part all check out.
  - The owner's Manual Bass plays through the synth voice slot, not ch 11, so a bass part left bent while Manual Bass is on doesn't detune the player.
- **SInt bends.** They are rescaled in `send_init`. I found 4 off-centre SInt bends on parts in the corpus: DownTime ch15 full up, Norteno ch12 +525, CelticTrance ch9, Calypso ch9.
  - DownTime ch15 goes out as 9362 at range 14, which is +2.0 st as written.
  - Norteno ch12 goes out as 8273 at range 13, which is +0.128 st as written.

  Both are correct.
- **Committed content.** Golden changes are digests only. No readable note content is committed, and every bar group is explained.

### Blocking

**B1. The PR conflicts with `integration/m2-chord-conversion`, and resolving it changes behaviour.** GitHub reports `mergeable: CONFLICTING`, and `git merge-tree` shows the conflict is in `src/engine.rs`. The base has moved on with #59 (structured SInt, re-applied on section change) and #55/#58. Once the conflict is resolved, the base's new code would undo this PR's bend handling unless it's ported:
1. **`Prepared.init` is now `Msgs`.** `init_len` is gone. This PR's RPN-0 rewrite (`init.iter_mut()` over `[u8; 3]`) and its RPN block appended at the end both have to be ported.
   - The base's `reapply_init` re-sends only `init[..init_resend]`, the parts' setup, on **every section change**.
   - The SInt's own RPN 101/100/6 sits in `SintChannel::other`, inside that resent prefix. If it's left raw, every section change resets the part's range to the style's value (for example 2). The engine still scales for `out_range` (12–24), so every later Pitch Shift bend and every rescaled pattern slide comes out 6× too small.
   - The appended "raise to out range" block sits after the effect SysEx, which is *past* `init_resend`, so it would not be re-sent.
2. **`reapply_init` re-sends SInt pitch bends raw.** It doesn't go through `send_bend` and doesn't set `pat_bend`. `send_init` in this PR does both. At range 14, DownTime ch15's full-up SInt bend would be +14 st after every section change instead of +2.
3. **`chase()`/`emit_control()` (base) send pattern `Bend` and RPN CCs raw.** They bypass `pattern_rpn`/`send_bend`, so a Fill entered mid-bar replays unscaled slides and the style's own RPN-0 value.

**Fix:** rebase onto the integration branch. Route `emit_control`'s Bend and RPN/NRPN cases through `send_bend`/`pattern_rpn`, which covers `chase` too. Keep the RPN-0 rewrite inside the resent prefix, and make `reapply_init` rescale SInt bends the way `send_init` does. Then rerun every corpus test.

Add a sim test: after Main A → Main B, and after a Fill entered mid-bar, the last RPN-0 data entry on each ch 11–16 must equal `out_range`. The test should also show that a Pitch Shift after the change still sends the exact bend value (like `pitch_shift_bends_the_held_bass`, but across a section change).

### Nonblocking

1. **Pitch Shift notes about to end keep the old pitch for up to 40 ms.** An ending note has no vote, so a lone Pitch Shift note just plays out unbent (`ends[a]` → `Leave`, `src/engine.rs` ~1662).
   - The golden change for OrganCruise shows it: its Pitch-Shift-to-Root bass (the Root Trans case, 1,291 zones) now sounds the old root about 17 ms into the new bar, while the pattern's next bass note is already sounding the new root.
   - The #49 reason (don't make a blip) applies to new *attacks*. A bend has no attack. RM p.31 says Pitch Shift "bends without a new attack", straight away.
   - Suggestion: let ending notes vote when no continuing note does. Then a lone ending Pitch Shift note still bends, and nothing else has to be retriggered because of it. This is a Decision, so it's the owner's call; flagging it because round 1 moved OrganCruise *away* from the manual.
2. **Rolled or partly fingered chords still blip.** Chords change key by key because the recognizer updates on each key.
   - I scripted F and then F7 3 ms later, at 2⅓ bars, across the corpus. 188 notes in 114 styles were attacked at F and cut 3 ms later: the `LATE_CHORD_NS` "correct outright" path retriggers them again.
   - This predates the PR and is outside #49 (which is about same-tick note-offs), but it's the same audible symptom in live play. Worth a follow-up issue: coalesce chord changes within a few ms, or let a note attacked less than N ms ago move silently.
3. **`MAX_BEND_RANGE` headroom assumes the style's range can't shrink mid-song.**
   - Example: an SInt range of 24 with no pattern bends gives `pat_max` 0, `out` 24 and `room` 24, so the accumulated shift can reach ±24.
   - If a pattern then sets RPN 0 = 2, `out_range` drops to 12 and `send_bend` clamps.
   - This happens 0 times in the corpus (pattern RPNs are 12 ×6 and 2 ×4). The fix is cheap: compute `room` from `min(out over the style's ranges)`, or make `out_bend_range` use the widest range the part ever has. `widest` already exists.
4. **Muted twins count twice in the vote.** `shift_of` counts muted voices, so a 1+8 fold outvotes one real voice with two copies of the same sounding pitch. That's harmless for pitch, but it's a hidden bias. Consider skipping `muted` in the vote.
5. **`folded_voices_part_again` never checks that it ran anything.** It only `eprintln!`s `ran`, so it passes when all 4 style files are missing. Assert `ran > 0` (or `== 4`) when the corpus dir exists, as the other corpus tests assert their counts.

### Checked and fine
- **Mixer principle.** No gain and no CC7 path touched. The only bend math is RPN/bend scaling, which isn't volume.
- **RT safety.** `revoice_part`, `struck_now`, `due_within`, the twin logic in `off_where`, `select_rpn` and `send_bend` use stack arrays and bounded loops, with no alloc, lock or panic path. The `retriggered: Vec` and `bend_clamps: Cell` hooks are `#[cfg(test)]` only.
- **Corpus counts.** They match round 1's rerun, and the Decisions cite them correctly.
- **Scope.** It stays within the revoice/RTR path, plus the documented `note_on` fold.

ryangavin:

I checked out `origin/m2/rtr-revoice` with the corpus linked.

**The branch has not changed since round 2.** The head is still 9cccff1, pushed at 20:29, before the round-2 review. There is no reply to round 2. GitHub still reports `mergeable: CONFLICTING`.

Gates on 9cccff1:
- `cargo test --release`: 209 passed, 1 ignored. The corpus tests ran: `sim::rtr::corpus_*` (4), `corpus_full_performance_no_stuck_notes`, `corpus_every_chord_type`, `corpus_one_plus_eight_and_one_plus_five` and `theory::…corpus_every_rule_every_chord`.
- `cargo clippy --release`: nothing new. `engine.rs:991` is the old Break code.

**Verdict: changes requested (1 blocking).**

### Blocking

**B1 (round 2 B1, not addressed, and now worse). The PR can't be merged into `integration/m2-chord-conversion`, and a merge doesn't compile.**

Since round 2, the base has also taken #57 (a1daaf5, "Bass On per SFF2 zone"). A trial merge (local only, aborted afterwards) gives:
- **Textual conflicts in `src/engine.rs`:**
  - `send_init`: the base loops over `Msgs`, and this PR loops over `init[i]`/`init_len[i]`.
  - The pattern `Cc/Pc/Bend` arm: the base calls `emit_control(dest, e.kind, sink)`, and this PR matches `cc`/`val` inline.
- **Keeping this PR's side of both hunks still gives 12 compile errors:**
  - `init_len` is not found (the `Prepared::new` RPN block, ~engine.rs:358).
  - `Msgs` has no `iter_mut` and can't be indexed (the RPN-0 rewrite, and `send_init`).
  - `cc`/`val` are not in scope in the combined `PKind` arm.
  - **New since round 2:** `E0609 no field bass_on on type &ChannelRule` at this PR's `engine.rs:1601`:

    ```rust
    let bass_on = rule.bass_on || zone.ntt == Ntt::Bass;
    ```

    #57 moved Bass On to `Zone`, so the to-Root target now has to be `zone.bass_on || zone.ntt == Ntt::Bass`. That is the same predicate `theory::transpose` uses on the base (`z.bass_on || z.ntt == Ntt::Bass`).

Resolving the conflict isn't mechanical. Everything in round 2's B1 still applies, and it's how this PR's bend handling would be lost after the merge:
1. **Port the RPN-0 rewrite so it sits inside `init[..init_resend]`.**
   - Otherwise `reapply_init` resets every chord part to the style's own range on each section change, while the engine keeps scaling for `out_range`. Pitch Shift bends and rescaled slides would come out up to 6× too small.
   - The appended "raise to out range" block sits after `init_resend`, so it would never be re-sent.
2. **Make `reapply_init` route SInt `0xE0` on ch 11–16 through `pat_bend` + `send_bend`, as `send_init` does.**
   - Otherwise DownTime ch15's full-up SInt bend goes out as +14 st at range 14 after every section change, instead of +2.
3. **Route `emit_control`'s `Bend` and CC 6/98–101 cases through `send_bend`/`pattern_rpn`.** This also covers `chase()`, so a Fill entered mid-bar doesn't replay raw slides or the style's raw RPN 0.
4. **Fix the `rule.bass_on` predicate above.**

Then:
- rerun every corpus test and regenerate the goldens on top of the base's new digests, explaining each changed bar group again;
- add the sim test round 2 asked for:
  - after Main A → Main B, and after a Fill entered mid-bar, the last RPN-0 data entry on each ch 11–16 equals `out_range`;
  - a Pitch Shift after the change still sends the exact bend value;
  - a style with an off-centre SInt bend (DownTime ch15) sounds +2 st after a section change.

### Nonblocking (carried over from round 2, none addressed yet)
1. A lone Pitch Shift note that is about to end keeps the old pitch for up to 40 ms: it has no vote. The OrganCruise golden shows the old root about 17 ms into the new bar. A bend has no attack, so the #49 reason doesn't apply. Let ending notes vote when no continuing note does. This is the owner's call.
2. A rolled chord (F, then F7 3 ms later) still attacks and cuts 188 notes across 114 styles. This predates the PR. Worth a follow-up issue.
3. `room` should come from the narrowest `out_range` the part can reach, so a pattern that shrinks RPN 0 mid-song can't make `send_bend` clamp. This happens 0 times in the corpus.
4. Muted twins vote twice in `shift_of`.
5. `folded_voices_part_again` should assert `ran > 0` when the corpus is present. At the moment it passes with all 4 styles missing.

### Checked and fine (unchanged code, re-verified on 9cccff1)
- Mixer principle: no CC7 or gain path touched.
- RT safety: stack arrays and bounded loops in the revoice path. The test hooks are `#[cfg(test)]`.
- Only golden digests are committed, and no readable note content.

Please rebase (or merge the base in), port the four points above, and reply to rounds 2 and 3.

ryangavin:
## Review round 3: response (344ca20)

The branch now merges `integration/m2-chord-conversion` (through a1daaf5, #57), so it is up to date with the base. I used a merge commit rather than a rebase, so the push was not a force-push. A trial merge back into the base is clean.

### Addressed

**B1: conflicts and the four ports.**
1. **RPN 0 inside `init[..init_resend]`.** `Prepared::new` now works out each part's bend range and widest pattern bend from the structured SInt before it builds `init`. As it builds, it raises the SInt's own RPN 0 data entries to the output range. It then pushes the RPN 0 set for ch 11–16 after the part setup and before the drum setup SysEx. All of that is inside `init_resend`, so a section change sends it again.
2. **`reapply_init` and SInt bends.** `send_init` and `reapply_init` now share `send_setup(n)`. It restores the style's bend range, clears the selected RPN, and sends SInt bends on chord-following parts through `pat_bend`/`send_bend`, which keeps any active pitch shift on top.
3. **`emit_control` and chase.** On chord-following parts, `emit_control` sends `Bend` through `send_bend` and CC 6/98–101 through `pattern_rpn`. `emit_at_index` and `chase` both use it.
4. **Bass On.** The to-Root predicate is now `zone.bass_on || zone.ntt == Ntt::Bass`.

**Tests for B1.**
- New `sim::rtr::section_change_keeps_the_output_bend_range` and `sim::rtr::mid_bar_fill_keeps_the_output_bend_range` use a synthetic style. Its SInt sets RPN 0 = 2 and bends +1 semitone on ch 14, and its patterns set RPN 0 = 4. Each test asserts that every bend range sent on ch 14 is 14, the output range, never 2 or 4. It also asserts these bend values:
  - after Main B: +1 semitone, the SInt's range again;
  - in the Fill: +2 semitones, the SInt at the entry, then the Fill's chased RPN;
  - after a C→F Pitch Shift: exactly +5 on top, with no retrigger.

  Both tests fail if either port 2 or port 3 is reverted.
- `init_is_structured_and_mixer_owns_volume` now asserts that the 36 RPN messages come before `init_resend`.
- `corpus_chord_changes_leave_no_blips_or_doubled_keys` now also asserts, for every corpus style and through the Main A–D section changes, that no RPN 0 below 12 ever goes out on a chord-following part.

**Goldens.** I regenerated them with `UPDATE_GOLDEN=1` on top of the base, and nothing changed. The per-bar explanations in the PR body still hold.

**Nonblocking, done:**
- **Bend headroom.** It now comes from the narrowest range a part can reach: the new field `Prepared::shift_room`, from the minimum of the SInt and pattern RPN 0. Test: `shift_room_fits_the_narrowest_range`. The corpus bend-clamp test still counts 0 clamps.
- **Muted twins.** A muted twin no longer votes when its sounding twin votes for the same shift, so each key gets one vote. Test: `muted_twin_votes_once` holds C3 twice, plus E3 and E4, through C→Cm. The part now bends by −1 and only C is struck again. Before this change it was a 2–2 tie that went to no bend.
- **`folded_voices_part_again`.** It now asserts that it ran whenever the corpus is present.

### Declined
- **Letting ending notes vote when no continuing note does (OrganCruise).** Not done. That window exists for a chord played up to 40 ms before the beat on which the note ends (#49). The note belongs to the chord that is ending, so it keeps its pitch. Retrigger mode leaves it alone the same way, and bending it for its last ≤40 ms would be the pitch glitch #49 removed. This is the owner's call: the Decisions section records it, and it is a one-line change if the owner prefers the other behaviour.
- **Rolled chord (F, then F7 3 ms later).** This happens before this PR and is outside #10/#46/#49. I did not file an issue from this branch. It should get its own follow-up issue (188 notes across 114 styles, per the review).

### Gates
- `cargo test --release`: 255 passed. The corpus is present, and no test printed a skip.
- `cargo clippy --release`: no new warnings in the code I touched.

The PR body's Decisions section is updated for the narrowest-range headroom, the one-vote-per-key rule and the `zone.bass_on` predicate.

ryangavin:
## Final adversarial review: approve

**Verdict: approve.** No blocking issues remain. Head is 882126b. I did not merge.

### Round 3 blockers: verified in 344ca20

I read the code and confirmed each fix by reverting it and rerunning the tests (throwaway, not pushed):

| Round 3 item | Fixed? | Tests that fail when it is reverted |
|---|---|---|
| 1. RPN 0 inside `init[..init_resend]`: the SInt's own RPN 0 raised in place, plus the ch 11–16 block before the drum setup SysEx | yes | `section_change_keeps_the_output_bend_range`, `mid_bar_fill_keeps_the_output_bend_range`, `corpus_chord_changes_leave_no_blips_or_doubled_keys` |
| 2. `reapply_init` sends SInt bends through `pat_bend` + `send_bend` (shared `send_setup`) | yes | the two synthetic section-change tests |
| 3. `emit_control` sends CC 6/98–101 through `pattern_rpn` (also covers `chase`) | yes | the two synthetic section-change tests |
| 3. `emit_control` sends `Bend` through `send_bend` | yes, but untested | **nothing** (see below) |
| 4. To-Root predicate is `zone.bass_on \|\| zone.ntt == Ntt::Bass`, as in `theory::transpose` | yes | — |

**Test gap, now fixed.** Sending pattern bends on chord-following parts unscaled undoes round 1's B2 rescale, so slides come out up to 6× too wide. On 344ca20 that still passed all 255 tests. None of them checks what the receiver actually hears.

### Merge with the current base (#61, keyboard parts)

3985ab9 merges `integration/m2-chord-conversion` (0957035) with a merge commit. There were no textual conflicts. I checked the semantics:
- **Style channels only.** Every engine bend and RPN path is gated on `follows_chords` (ch 9–16, drums excluded).
  - `all_off` recentres only ch 9–16.
  - The player's wheel and pitch bend reach the keyboard parts through `live.rs`, a separate path.
  - The #61 assertion (the Style never sends on ch 1–4) passes, and the new chaos test checks it again under every action.
- **Mixer.** Unchanged. CC7 is still the only per-part volume, and moved faders win on a section change.
- **Goldens.** `UPDATE_GOLDEN=1` regenerates every digest byte for byte, so none changed.

### Integration attack on the bend-based Pitch Shift

882126b adds `sim::rtr_chaos::corpus_chaos_leaves_no_note_detuned`.

**What it plays.** Every corpus style (208 files), 4 seeds each, 22 bars:
- chords of every type, on the beat, within 40 ms either side, or anywhere in the beat;
- slash chords, Chord Cancel, and rolled chords (a second change 3 ms later);
- Main A–D, Break, Intro, Ending and Auto Fill;
- part toggles, Manual Bass, Keyboard/Master transpose, tempo, Sync Start/Stop, Stop Accompaniment and chord release;
- always: a stop and restart mid-pattern, and a **style change while running**, 2 ms after a chord change.

**How it checks.** It replays the output through a **model of the receiver** (RPN select, CC 98/99 deselect, RPN 0 range, 14-bit bend per channel). It does not use the engine's own bookkeeping.

**What it asserts** on the chord-following parts:
- At every note-on, the receiver's offset (bend × range) equals the engine's pitch shift, give or take no more than the part's own pattern bends. The bend range is at least 12.
- No bend is clamped.
- No key is started twice.
- **No note has zero length.**
- Nothing is stuck.
- No part is left bent after Stop. The one exception is a style loaded while stopped, which sends its own SInt bend.
- The Style never sends on ch 1–4.

**Results:** 832 runs, 419,278 note-ons and 43,668 off-centre bends, with 0 failures and a worst excess of 0.0015 semitones (rounding).
- It fails on the unscaled-bend mutation (1,695 detuned note-ons).
- It also fails when `reapply_init` sends SInt bends raw.

**Findings against the brief:**
- **Bends left hanging after stop, style change or section change:** none. `all_off` runs before `reapply_init`, before `load`'s `send_init`, and on every stop, and it clears `rtr_bend` and `pat_bend`.
- **#59 SInt re-apply:** the range is at least 12 at every note-on after every section change, Fill chase and style change. SInt bends come out in the right place (as in round 2, DownTime +2 st and Norteno +0.128 st).
- **Real-time safety:** checked the merged `send_setup`, `emit_control`, `pattern_rpn`, `send_bend` and `revoice_part`. They use stack arrays and bounded loops, with no allocation, lock or panic path. `dest_ch` is masked to 0–15 at parse time. `retriggered` and `bend_clamps` are `#[cfg(test)]`.

### Gates (882126b)
- `cargo test --release`: **269 passed, 1 ignored**, with the corpus present. I confirmed the corpus tests run over the 208 files. 255 of these are the tests from 344ca20, 13 came in from #61's merge, and 1 is new.
- `cargo clippy --release`: 19 warnings, against 21 on the base. None are new; the `engine.rs` one is the old Break code.

### Remaining concerns (non-blocking, both predate this PR)
1. **Two inputs in one engine wake-up can still make a zero-length note.**
   - **Cause:** a chord change and a command handled with the same `now` (Keyboard transpose, Start/Stop, Break, chord release, Sync Start, Stop Accompaniment). The chord restrikes a note and the command moves or cuts it at the same instant.
   - **Probe:** chord + Keyboard transpose in one wake-up, every corpus style, 8 moments each:

     | | same instant, chord first | same instant, transpose first | 1 µs apart |
     |---|---|---|---|
     | base | 16,159 | 17,130 | 8,599 |
     | this PR | 3,367 | 1,926 | 0 |

   - **Why it isn't blocking:** this PR greatly reduces it, and live play rarely triggers it, because the thread wakes for each input as it arrives.
   - **Follow-up:** coalesce the chord and the commands per wake-up. That is the same work as round 2's rolled-chord item.
2. **Rolled chords (round 2, NB2)** are still open, as a follow-up issue. So is the OrganCruise "ending note keeps its pitch" Decision, which is the owner's call and is recorded in the PR body.

ryangavin:
## Final review, follow-up: approval withdrawn; blocked on #62

**My approval above no longer stands.** The base moved while I was reviewing: `integration/m2-chord-conversion` took #66 (44ccb0a) and #62, the capture kit (cc20044), after I merged 0957035. GitHub now reports this PR as `CONFLICTING`. Everything I verified about 882126b still holds, but 882126b doesn't merge into the current base.

I did not push a second merge, because it can't be made green without an owner decision. The PR branch stays at **882126b**.

### What merging cc20044 involves

**1. `src/sim.rs` textual conflict: resolved (in a local WIP branch).**
- #62 moved the golden listing into `Take`/`PlayedNote`. This PR's listing shows the sounding pitch and `From>To` shifts.
- Port:
  - `PlayedNote.key` stays the **key sent**, because the capture importer pairs note-offs by it.
  - A new `Take::pitch` holds the pitch each note sounds at its start.
  - A new `Take::shifts` holds the pitch shifts of sounding notes.
  - Both are used only by `render` (`part_line`). `part_items`, `render_reference` and the capture comparison stay notes-as-sent.
  - The importer's `hardware` take clears both.
- Result: `UPDATE_GOLDEN=1` regenerates this PR's digests byte for byte, and every other test passes (289 passed).
- I tried making `key` the sounding pitch first. That broke `capture::tests::import_round_trip`: overlapping Chord2 notes started on the same sounding pitch, one bent and one sent compensated, and the importer closed the wrong one.

**2. Semantic conflict: #62's kit timing against this PR's 40 ms early window. Blocking, and it needs a decision.**
- **The kit's promise.** `capture::plan` times every kit chord away from the points where a few ms decide the outcome: note starts and ends, and `LATE_CHORD_MS` after a start. `kit_timing_survives_jitter` then asserts that ±3 ms of step jitter doesn't change our output.
- **The new points.** This PR adds `EARLY_CHORD_NS` (`due_within`/`struck_now`): a chord up to 40 ms before a pattern note-off, or before a note-on of the retriggered pitch, leaves the note alone. Every note end and start minus 40 ms is therefore a new cliff.
- **Failure on the merged tree:**
  - `JackDoesItAgain: steps -11 ticks late: key 48 on ch12 at tick 22080: ends (Some(23000.0), Some(22850.0))`.
  - A kit chord lands about 38 ms before a Chord1 note-off, and 3 ms earlier it is past 40 ms.
- **Adding the early points to the kit's `edges` isn't enough.** I tried adding them before starts, before ends and before the section end, with chord leads up to 3/8 of a beat:
  - 90sDisco's best room is **2.5 ms**, under the 3 ms budget;
  - with the margin assertion set aside, its output still changes under 3 ms of jitter: `key 73 on ch14 at tick 581676: ends (583576, 583478)`.
  - Margins, quarter-beat leads:

    | edges | 90sDisco | SoulShuffle | BluesOrganTrio | TickingAway |
    |---|---|---|---|---|
    | base kit edges | 6.0 ms | 6.1 ms | 10.9 ms | 7.3 ms |
    | plus the early points | 2.5 ms | 2.8 ms | 3.6 ms | 4.2 ms |

**Options (owner's call):**
- (a) Keep the 40 ms window. The kit adds the early points and accepts a smaller jitter budget, or drops or retimes 90sDisco, or bounds clock drift more tightly.
- (b) Narrow `EARLY_CHORD_NS`. That moves the cliffs; it doesn't remove them. It also brings back part of the live-play blips #49 is about (round 1 measured 1,208 at 1 ms early and 5,488 at 20 ms with no window).
- (c) Have the kit compare chords near a cliff as "either outcome" rather than requiring equality.

My recommendation is (a). The window is what fixes #49 in live play, and the kit is a test harness whose guarantees can be scoped per style.

### Where the work is
- Local branch `final63-merge2-wip` (2853205, not pushed) holds the resolved merge. Its kit `edges` include the early points before note starts and ends, not yet before the section end. `kit_timing_survives_jitter` is its only failing test.
- Once the kit timing is decided, the implementer can take it from there. Please rebuild the merge on the base as it is then.

The non-blocking concerns from my comment above still stand: same-instant input collisions, rolled chords, and the OrganCruise Decision.

ryangavin:
## Final review: approve (head a948f19)

**Verdict: approve.** This replaces my follow-up that withdrew approval. GitHub reports `MERGEABLE`. I did not merge.

a948f19 merges the current `integration/m2-chord-conversion` (cc20044: #66 and #62) with a merge commit, following the coordinator's decision to **keep the 40 ms early window and weaken the capture kit's guarantee**.

### `src/sim.rs` conflict (#62's `Take` refactor)
- `PlayedNote.key` stays the **key sent**, because the capture importer pairs note-offs by it.
- The golden listing's sounding pitch and `From>To` shifts move to `Take::pitch` and `Take::shifts`. Only `render` uses them; `part_items`, `render_reference` and the capture comparison are unchanged. The importer's `hardware` take clears both.
- `UPDATE_GOLDEN=1` regenerates every digest byte for byte.

### Capture kit against the early window
**(a) Placement.**
- Decision points are now per part: note starts and ends, `LATE_CHORD_MS` after a start, and the new points `EARLY_CHORD_MS` before a start, an end or a section end.
- `plan` picks each chord's lead to leave the **fewest parts** within `JITTER_MS` (3 ms) of a decision point, then the most room.
- Styles that clear every part get the same leads as before.

**(b) Timing-sensitive cells.**
- A chord still within 3 ms of a part's decision point is timing-sensitive **on that part only**.
- Its cells (bar, part) run from just before the chord to where that part next falls silent. That end point is there because a pitch-shift bend carries into later notes. The cells also include the bars where the notes it could cut started.
- How each consumer treats them:
  - **Importer:** lists their differences under their own heading and doesn't count them in "bars that differ".
  - **`--golden`:** leaves them out of `.known`.
  - **`reference_captures`:** ignores them.
  - **Clock budget and chord timing check:** still hold each chord to its room on its other parts.
- Per style, now also in the new "Timing-sensitive" column of the kit table:
  - 90sDisco: Chord1, 62 bars. Before per-part marking and placement it was Bass, Chord1, Chord2, Pad and Phrase1 in 98 bars.
  - SoulShuffle: Chord1, 3 bars.
  - The other 7 kit styles: none.

**(c) Tests and docs.**
- `kit_timing_survives_jitter` now asserts this guarantee exactly. With steps 3 ms late, 3 ms early, or each jittered within ±3 ms, no section and no note changes outside the timing-sensitive cells. It also asserts that the cells are needed somewhere, and that steps on the beat line are still unsafe.
- New `timing_sensitive_bars_accept_either_outcome` imports a recording of the other outcome and checks that it:
  - comes out verified;
  - shows 0 differing bars;
  - lists the differences under the timing-sensitive heading.

  A wrong note outside the cells still counts as a difference. The test fails if the importer stops setting the cells aside.
- The kit README ("Timing-sensitive bars") and `tests/reference/README.md` document the guarantee.

**Side effect:** clock budgets are now 19–61 ppm (they were quoted as 30–110). The early points narrow the rooms. `--clock-ppm` covers any recording that goes past its budget, as before.

### Gates (a948f19)
- `cargo test --release`: **291 passed, 1 ignored**, with the corpus present (208 styles).
  - That includes `sim::rtr_chaos::corpus_chaos_leaves_no_note_detuned`: 832 runs, 0 failures.
  - It also includes every `sim::rtr::corpus_*` test and all capture tests.
- `cargo clippy --release --tests`: 19 warnings, against 20 on the base. None are new.
- My earlier verification still holds: round-3 blockers fixed, #61 keyboard channels untouched, mixer untouched, real-time safety.

### Remaining concerns (non-blocking)
1. **Same-wake-up inputs.** A chord and a command handled in the same engine wake-up can still make a zero-length note. This predates the PR, which greatly reduces it.
2. **Rolled chords** are still open, as a follow-up.
3. **OrganCruise.** The "ending note keeps its pitch" Decision is the owner's call.
4. **Compensated keys in captures.** On a bent part, yahaha sends compensated keys, and the capture compares keys as sent. Whether a Genos transmits the shifted pitch or the original key for Pitch Shift notes will only show once real recordings arrive.

PR #66, opened 2026-09-23
Corpus tests and tools scan every style type recursively

ryangavin:
Some corpus tests and tools listed corpus files themselves, and some of them only looked at `.sty` files or at one folder. They now all use one helper: `library::style_files(root)`. This is the scan the style browser already uses. It walks folders recursively, matches every extension in `EXTENSIONS` (sty, prs, sst, bcs, pcs, pst, fps) in any case, skips hidden files and folders, and follows each symlinked folder once. Tests call it through `library::corpus_styles()`.

## Files touched
- `src/library.rs`: the folder walk moved out of `Library::scan` into `walk`, and `style_files` and `corpus_styles` (test only) were added. There are two new tests:
  - `style_files_finds_every_style_type_in_subfolders`: in a temp dir holding `a.sty`, `Registrations/b.prs` and `Registrations/deeper/c.SST`, all three must be found. A `.txt` file and a `.hidden/` folder must not be.
  - `only_the_library_walks_folders`: fails if any file in `src/` other than `library.rs` calls `read_dir(` itself. `main.rs` is allowed its one call, which looks for a `.sf2` in `soundfonts/`.
- `src/theory.rs`: `corpus_every_rule_every_chord` used to read `.sty` only.
- `src/synth.rs`: the `loudness_probe::part_loudness` probe (`#[ignore]`) used to read `corpus/MOX_v2/*.sty` only.
- `src/sim.rs`: the `corpus()` helper had its own extension list, which was missing `pst`.
- `src/sff.rs`: `corpus_cntt_bass_parts_follow_slash_chords` and `sint_structures_every_corpus_style` used to read every file, junk included. `sint_structures_every_corpus_style` now fails when a style doesn't load, instead of skipping it.
- `src/golden.rs`: `find_style` now uses the shared helper. The golden list and digests are the same.

## Styles scanned (full corpus: 208 styles, of which 35 .sty, 155 .prs, 18 .sst)
| Consumer | Before | After |
|---|---|---|
| theory `corpus_every_rule_every_chord` | 35 (.sty only) | 208 (7537 channel rules) |
| synth `part_loudness` probe | 35 (MOX_v2 .sty) | 208 |
| sim `corpus()` tests | 208 (`pst` missing from its list) | 208 |
| sff Cntt / SInt tests | every file, errors skipped | 208; the SInt test now fails on a load error |
| oracle, library index, ui | 208 (already on `Library::scan`) | 208 |
| golden | fixed list of 6 | same fixed list of 6 |

## Oracle delta
None. `oracle::corpus_scores` already went through `Library::scan`, so `tests/oracle/scores.txt` already pinned all 208 styles, and it was not re-pinned.

## New failures and issues opened
None. Every test passes on the full corpus: 244 passed, 1 ignored (the loudness probe, as before). Clippy warnings went from 31 to 29.


ryangavin:
## Adversarial review: no blocking findings, approve

I checked out `origin/corpus-all-types` at 6acbfef with the full corpus symlinked in. `cargo build` works. `cargo test` gives 244 passed and 1 ignored. Clippy reports 29 warnings, and none of them are in the lines this PR changes.

### Does every consumer now see all 208 styles? Yes.
I checked this with a throwaway test, which I have since reverted:
- `corpus_styles()` returns 208 styles: 35 sty, 155 prs, 18 sst. `Library::scan(corpus/)` also returns 208.
- `corpus_every_rule_every_chord` prints `208 styles, 7537 channel rules`. Before this PR it saw 35.
- The old SInt/Cntt walk visited every file. That was 210 files, and exactly 2 of them failed to load: `corpus/.DS_Store` and `MOX_v2/README_PERF_V2.TXT`. So the old SInt test covered 208 styles too, and the new one covers the same 208. The Cntt test finds 7 Cntt files both before and after.
- Golden: none of the file names in the corpus repeat, so switching `find_style` from DFS order to sorted order can't pick a different file.

### Did any test get weaker? No.
- **SInt test (nonblocking, confirmed):** it is stricter. Before, `let Ok(..) else continue` would silently skip a real style that failed to load. Now it panics with the path. The only files the old code skipped were the two non-style files above, and the new scan excludes those by extension.
- **Cntt test:** it still calls `parse(..).unwrap()`, so it is as strict as before, and it covers the same 7 files.
- **Theory test:** it now calls `Style::load(..).unwrap()` on prs and sst files as well as sty, so it is stricter.
- **sim `corpus()`:** it covers the same set, and now includes `pst`. The corpus has no pst files, so nothing changes in practice.
- **Loudness probe:** it now reads 208 styles instead of 35 MOX_v2 styles. It is `#[ignore]`d, but it will run much longer when someone runs it by hand. Also, its output can no longer be compared with earlier runs. (nonblocking, just FYI)

### Is `Library::scan` unchanged? Yes.
The walk, the hidden-folder rule and the symlink `seen` set are unchanged. `seen` is still shared across roots. `q.parent()` is the same `d` the old code stripped. Order doesn't matter because `from_entries` sorts. The existing scan tests (hidden, fifo, multi-root, symlink loop) pass.

### Are any `.sty`-only filters left? None.
I grepped `src/` for `read_dir`, `extension(`, `"sty"` and `.sty`. The only `.sty` hits left are fixed-file fixtures (`SlowWalker.T552.sty`, `FunkyFinger.S930.STY`, and similar), and those are fine. The per-folder corpus tests in `sim.rs` (`corpus/MOX_v2/...`, `corpus/T5Style/...`) each load one named file. They are not scans.

### `only_the_library_walks_folders` (nonblocking): it works, but it is brittle
- It counts raw substrings, so a comment, doc line or string that mentions `read_dir(` fails the build. The test gets around this in its own source by splitting the needle into `["read", "_dir("].concat()`.
- It only reads top-level `src/*.rs`. A future `src/foo/mod.rs` or anything under `tests/` escapes it.
- `main.rs` is allowed exactly one call. Any other legitimate folder listing will trip it, for example a soundfont picker, an export folder or a registration bank dir. Most of those have nothing to do with styles.
- It is easy to get around: `fs::read_dir (`, `walkdir`, `glob`.

A better alternative: add `clippy.toml` with `disallowed-methods = [{ path = "std::fs::read_dir", reason = "list styles with library::style_files" }]`, and put `#[allow(clippy::disallowed_methods)]` on `walk` and on the sf2 lookup in `main.rs`. That check understands types, covers every module, and lets a new site opt in on the spot, with a reason. The catch is that clippy doesn't gate CI today, and there are 29 existing warnings, so the lint would need `#![deny(clippy::disallowed_methods)]` in `main.rs` to have teeth. If you keep the text test, a smaller fix is to skip lines starting with `//`, walk `src/` recursively, and put the allowlist in a named `const` with a comment explaining each entry.

### Verdict
**Approve.** No blocking findings, so I made no fixes. I did not merge.

PR #67, opened 2026-09-24
Seamless section changes: every section change re-sent the whole SInt ahead of the downbeat

ryangavin:
## Symptom

On `integration/m2-chord-conversion`, fills and section changes feel abrupt and off the grid, where M1 (f962dc1) felt queued and on time.

## Root cause

Since #59, `reapply_init` → `send_setup(init_resend)` re-sent the style's entire part setup at **every** section change, Main → Fill → Main included. That meant a bank select and program change for every part, the XG part parameters, the drum setup SysEx and six bend-range RPN messages per part. All of it went out on the boundary tick, in the same packet list as the new section's downbeat notes and **ahead of them**. In every case measured below the setup it re-sent was identical to what the parts already had.

The timing itself did not regress, and neither did the notes. What changed is the burst of redundant setup sent at the boundary.

## Evidence (offline sim, same script on f962dc1 and on the tip)

Script: Main A. Press Main A at beat 2.5 of bar 2 for a fill. Press Main B with Auto Fill on. Turn Auto Fill off and press Main A. Press Break. Chords are placed 13–24 ms before and after the boundaries.

**Timing is identical on both builds.** The fill was queued at the press (tick 10560) and started at tick 11520, the next beat. Main B was queued at tick 40704 and started at tick 46080, the next bar. The queued state (the flashing LED) was visible from the press until the boundary and cleared exactly there. First note-on ticks match.

**Notes are identical.** For FunkyFinger, the note stream is identical, and rendering both streams through rustysynth gives sample-identical audio. The other styles differ only through the NTT and #63 chord logic, not at the boundaries.

**Traffic per section change (after Start):**

| style | M1 msgs / bytes ahead of 1st note | tip (before fix) | program changes (all identical to current voice) | SysEx | **this PR** |
|---|---|---|---|---|---|
| FunkyFinger | 24–29 / 72–75 B | 78–83 / 244–247 B | 2 | 2 | 0–5 / 0–3 B |
| AnalogBallad | 31–36 / 72–93 B | 155–197 / 543–678 B | 9 | 18 | 7–13 / 0–21 B |
| 60s8Beat | 29–39 / 72–102 B | 146–155 / 619–649 B | 8 | 34 | 5–14 / 0–30 B |
| 30sBigBand | 26–33 / 72–75 B | 170–190 / 630–681 B | 12 | 24 | 3–12 / 0–27 B |
| SlowWalker | 27–33 / 75–87 B | 104–110 / 299–311 B | 7 | 0 | 3–9 / 3–15 B |
| Merengue | 28–31 / 72–84 B | 112–131 / 347–404 B | 7 | 5 | 4–7 / 0–12 B |
| ClubJazz1 | 24–34 / 72–96 B | 90–100 / 277–301 B | 5 | 2 | 0–10 / 0–27 B |

At 31.25 kbaud (DIN MIDI), 543–681 bytes is **174–218 ms** of wire time before the downbeat, against 23–33 ms on M1. Over USB, the bytes cost nothing, but the receiver still does the work: a Genos or XG module reloads each of 5–12 voices on a program change and rewrites the drum kit on each drum setup SysEx, at the moment the new section starts.

**What the built-in synth sees.** SysEx never reaches it: `Out::push` drops SysEx. Redundant program changes and controller values do not reset rustysynth voices or tails (`set_patch` only stores the number). So the synth's audio is unchanged by the burst. The burst lands on the `yahaha` port, which means an external instrument, a DAW, or a GM/XG plugin.

**The other suspects, measured:**
- `chase()`: a mid-bar Fill replayed every controller and bend before its entry. AnalogBallad Fill BB sent 50 bends in one instant. The final values were correct, but the burst added to the pile. It now sends each value once.
- #63, the 40 ms early window: it does not move the boundary's all-off. Held notes are cut exactly at the boundary, as on M1. The only change is that a chord landing within 40 ms before a boundary no longer re-attacks notes the boundary cuts 13–24 ms later. M1 did that: for example, 8 re-attacks on 60s8Beat at tick 15302 for notes cut at 15360. The new behaviour is cleaner.
- Latency: `yahaha bench` puts engine work per wake at p99 13 µs, max 64 µs. The engine thread is not the bottleneck. The cost is on the wire and in the receiver.

## Fix

- The engine keeps a `Mirror` of what it has sent on each channel: controllers, voice (bank MSB, LSB, program), (N)RPN data entries and bends. All engine sends that change that state go through it.
- A section change still puts back what a pattern changed in the SInt, but **sends only values that differ**:
  - A program change is sent only for a voice that differs. It is not sent at all if the new section sets its own voice by its entry point, so a part changes voice once, not twice.
  - The part's XG parameters are sent only after a program change for that part, and the drum setup SysEx only after a program change (both are reset by one).
  - The effect SysEx and resets are never sent. System On was already stripped.
  - RPN/NRPN values are sent only where they differ, and the selection is left as the SInt leaves it.
- `chase()` sends the value each controller and bend has at the entry, once. Bank selects, program changes and RPNs still go in order.
- A pattern program change for the voice a part already has is dropped.
- The boundary's bend-centre, mod-0 and sustain-0 messages go only to parts not already at rest. Stop, Start and style load still send them in full.
- Start and style load still send the full SInt.
- Mixer principle unchanged: CC7 is the fader, and a fader the player moved is never overridden.

What newer instruments re-apply: the Genos manuals don't document re-sending the setup on a section change. The Style Creator notes that each section can carry its own voices, so the only thing a section change needs is to undo a voice or controller change the previous section's pattern made. That is exactly what is now sent, and nothing else.

## Tests

- `section_changes_land_on_the_grid`: a fill starts exactly on the next beat and a Main exactly on the next bar (ns-exact). The queued state is visible at every wake from the press until the boundary and cleared after it. Main B's first note is on its downbeat.
- `main_fill_main_resends_no_setup`: a Main → Fill → Main → Main B cycle whose patterns change nothing sends no bank select, program change, (N)RPN or SysEx after Start. This fails on the tip.
- `corpus_section_changes_send_no_redundant_setup`: across the whole corpus, a section change never sends a program change for the voice a part already has, and never sends SysEx except right after a program change it had to send.
- Updated tests: `corpus_drum_setup_survives_section_change` now checks that the drum setup is sent at Start and never rewritten, and the voices stay the SInt's. `mid_bar_fill_plays_its_own_voice` now checks that the Fill's voice replaces the part's voice with one program change, not the SInt's and then the Fill's.

## Gates

- `cargo test --release` with the corpus: 306 passed, 1 ignored (pre-existing).
- Clippy: 18 warnings, identical to the base commit.
- **Golden digests: unchanged.** They hash notes, and the note output is unchanged.

## What the owner should feel

On the `yahaha` port (Genos, XG module, DAW), a fill or section change now arrives on the beat with nothing ahead of it: no voice reloads, no drum-kit rewrite, no wire-time delay. It should feel as queued and seamless as M1, with slightly less boundary traffic than M1 had.

In the built-in synth, the audio at boundaries was already sample-identical to M1 before this fix, and it is unchanged by it. If the abruptness was heard only through the built-in synth, this PR is not the cause, and I'd want to know which style, output and control were used to chase it further.


ryangavin:
## Adversarial review (posted after the merge)

**Verdict: one blocking finding. It is fixed in follow-up PR #69, which targets `integration/m2-chord-conversion`.** Everything else is nonblocking.

I reviewed head `7f58cfa`, which is identical to the squash `d335695`. Build ok. `cargo test --release` with the corpus: 306 passed, 1 ignored. Clippy: 18 warnings, unchanged from the base.

### Method
I wrote a throwaway receiver-state probe that models what an XG receiver holds on each part at every note-on on ch 9–16:
- voice (bank and program)
- all CCs except CC7
- bend
- (N)RPN data values
- XG Multi Part params, cleared on a program change
- whether the drum setup is still valid on a kit, i.e. received after the kit's last program change

I ran it over all 208 corpus styles with an 18-step script: chords every bar, a mid-bar Fill, Main A–D, Break, Auto Fill, Stop/Start, Ending A and Intro A. I ran it on the base (`b02ec44`) and on this PR, then compared the outputs.
- 205 styles give identical output.
- 3 styles are better on this PR: SmoothItOver, SmoothLeadJam and SunsetDrive. Dropping a pattern program change that restates the current voice keeps the XG part-mode and param values that the base's redundant program change wiped.
- In no case is this PR worse on the corpus.

### Findings

**1. BLOCKING: a pattern voice change that returns to the SInt voice loses the XG part params and the drum setup until Stop/Start.**
`reapply_init` sends a part's `08 pp` params only for channels in `voice_sent`, and the drum setup only if `voice_sent != 0`. Take a pattern that goes X → K inside a section, where K is the SInt voice. The receiver resets the part twice. At the next boundary, `mirror.voice == K`, so the section change sends no program change, no XG params and no drum setup. Before this PR, the next section change restored them.
- The corpus has this voice pattern: BrazilianSamba Main A/B goes 120 → 101 → 120, and 82 sections change voice mid-section. None of them happened to land on a part that has SInt XG params or a drum setup in my script, so the corpus probe didn't catch it.
- A synthetic test proves it: `section_change_restores_xg_setup_after_a_pattern_voice_round_trip`, with ch 13 going 5 → 9 → 5 and the kit going 0 → 1 → 0. It fails on `d335695`.
- Fix in #69: `pattern_pc` records the channels where a pattern or `chase()` actually sent a program change. The next section change that doesn't set its own voice there re-sends that part's XG params, plus the drum setup for a kit. The corpus probe gives the same output with the fix, and `corpus_section_changes_send_no_redundant_setup` still passes.

**2. Nonblocking: `chase()` defers CCs, which reorders mode messages.**
Non-bank, non-(N)RPN CCs are collapsed and sent in CC-number order after the program changes. Suppose a pattern sends CC121 (Reset All Controllers) before, say, CC11. The chase would send CC11 and then CC121, which wipes it. `Mirror::track` also doesn't model CC121's reset of CC1, CC11, CC64, bend and RPN selection, so the Mirror would go stale. The same applies to CC96/97 (data increment/decrement), which `reapply_init` dedupes as if they were absolute values. The corpus has **zero** CC96–99 or CC120+ in any pattern or SInt, so this is theoretical. A cheap guard would be to send CC ≥ 120 and CC96/97 in order through `emit_control`, and to reset the Mirror entries on CC121.

**3. Nonblocking: the Mirror assumes it is the only writer on ch 9–16, and that every message it sends arrives.**
Both hold today:
- Keyboard passthrough and part CC7 go only to ch 1–4 (`parts::CHANNEL`).
- Stop Accompaniment sends only notes, and Manual Bass sends only note-offs.
- Panic sends only CC123, after `stop`.
- The mixer's CC7 goes through `mirror.send`.
- Style load and Start still send the full SInt.

What's lost is the old side effect that every section change **self-healed** the receiver. If the Genos is power-cycled or reset mid-song, or someone changes a part voice on its panel, the receiver now stays wrong until Stop/Start. The same goes for a message the built-in synth ring dropped (`let _ = s.push(m)`). This seems acceptable, but it's worth one line in the docs.

**4. Nonblocking and pre-existing: a section's own tick-0 program change still resets the SInt XG params and drum setup after the boundary.** The base does the same (SInt program change, then params, then the pattern's program change). This PR doesn't make it worse.

**5. Order-dependent sequences in `chase()`: OK.** Bank select → program change, RPN/NRPN select → data entry, and RPN 0 → bend all stay in order. The deferred bends go out after the bend range, so the final output bend is scaled correctly. The bank selects are compared with the Mirror before the program change, and the voice tuple includes the bank, so a bank-only change still triggers the program change. The SInt's pending bank select (CC0 after the program change) is handled correctly.

**6. Real-time safety: OK.** The `Mirror` is boxed in `Engine::new`, off the RT thread. The per-boundary scratch state in `reapply_init` and `chase` is fixed-size stack arrays (about 2 KB in `chase`). `Msgs::get` is slice indexing. There is no allocation or lock.

**7. Changed test expectations: justified.**
- `corpus_drum_setup_survives_section_change`: AustinCityBlues' patterns never change a voice, so not re-sending anything is correct. The test still checks the drum setup block at Start and the final voices.
- `mid_bar_fill_plays_its_own_voice`: the resulting voice is identical (8/112/20). Only the throwaway SInt program change 5, which was immediately overwritten, is gone.

Neither change hides a regression.

**8. Transport paths checked:** Main → Fill → Main, Break, Intro → Main, Ending → stop, Stop/Start, style load while running, and a section repeating itself. Stop, Start and load use the full `notes_off(false)` and `send_init`. I found no OTS-recall path on the engine side that touches ch 9–16. Keyboard channels 1–4 are untouched.

**9. Nit:** `own_voice` indexes `sec.rules[e.src as usize & 15]`, while `chase` and `emit_at_index` use `sec.rules[e.src as usize]`. They're equivalent, but should be consistent.

PR #69, opened 2026-09-24
Section change restores XG part params and drum setup after a pattern voice round trip

ryangavin:
Follow-up to #67 (a blocking finding from the adversarial review, posted after the merge).

## Bug

Since #67, `reapply_init` sends a part's XG Multi Part parameters, and the drum setup SysEx, only after a program change that it sends itself. Take a pattern that changes a part's voice and then goes back to the SInt voice inside one section, for example program 120 → 101 → 120 (BrazilianSamba Main A does exactly this on one part). On an XG receiver, each of those program changes reset the part's XG parameters, and on a kit they also reset its drum setup. At the next section change the Mirror sees the SInt voice already in place, so it sends no program change, and it sends no XG parameters or drum setup either. The part stays without them until Stop/Start. Before #67, the next section change put them back.

## Fix

- `Engine::pattern_pc` records the channels where a pattern, or `chase()`, actually sent a program change. The dedupe from #67 still applies, so a restated voice doesn't count.
- At the next section change, `reapply_init` re-sends that part's XG parameters, plus the drum setup if the part is a kit. It does this even when no program change is needed.
- Parts whose new section sets its own voice (`own_voice`) are skipped, because the SInt's parameters belong to the SInt voice. Their bit is kept, so the first section change that doesn't set its own voice restores them.
- `send_init` clears the record.

## Tests

- New: `section_change_restores_xg_setup_after_a_pattern_voice_round_trip` (synthetic). Main A changes ch 13 from 5 → 9 → 5 and the drum kit from 0 → 1 → 0. The test checks that the move to Main B re-sends the XG part parameter and the drum setup after the last program change and before Main B's first note. It fails on d335695.
- The corpus is unchanged. A receiver-state probe (voice, CCs, bend, (N)RPN values, XG part params and drum-setup validity at every note-on, over all 208 corpus styles, with an 18-step script covering Fill, Break, Main A–D, Stop/Start, Ending and Intro) gives identical output before and after this fix. `corpus_section_changes_send_no_redundant_setup` still passes.

## Gates

- `cargo build --release`: ok
- `cargo test --release` with the corpus: 307 passed, 1 ignored (pre-existing)
- Clippy: 18 warnings, the same as the base


ryangavin:
Superseded by the same fix cherry-picked onto integration/m3-ui, since M2 has already merged to main.

Issue #16, opened 2026-09-23
Extract the `engine` library with a command/event API


ryangavin:
- Move every mutation behind `Cmd`, including SynthControl voice/OTS/left writes that currently bypass the engine.
- Publish state as `Snapshot` and events. The TUI becomes the first client. No real-time behaviour changes; bench latency must not regress.

---
Milestone: **M3 · Engine API + Tauri shell** · Spec: [docs/genos-features.md](../blob/main/docs/genos-features.md) · Backlog: [docs/backlog.md](../blob/main/docs/backlog.md)

Issue #17, opened 2026-09-23
Tauri app skeleton


ryangavin:
- The webview only displays state: it subscribes to snapshots (~60 Hz) and sends commands. MIDI, audio and timing stay in Rust.
- Panel: transport, sections with lamp states, chord display, tempo, parts, mixer.

---
Milestone: **M3 · Engine API + Tauri shell** · Spec: [docs/genos-features.md](../blob/main/docs/genos-features.md) · Backlog: [docs/backlog.md](../blob/main/docs/backlog.md)

ryangavin:
Owner requirement: **every control in the app has a tooltip**, so you can learn how anything works by hovering. Each tooltip covers:
- what the control does (and its Genos name, if different)
- the keyboard shortcut
- where it lives on the Launchkey, including pad page, button and fader

Tooltip text lives in one catalog, which is the single source of truth.

PR #71, opened 2026-09-24
Engine library + Session API (AppCmd/AppState) (#16)

ryangavin:
This splits the engine into a library with a `Session` API. The terminal UI becomes its first client, and the Tauri app (#17) will be the second. **The contract is [docs/app-api.md](../blob/m3/engine-api/docs/app-api.md).** Types are in `src/api.rs` and the runtime is in `src/session.rs`.

## What changed
- **Library and binary.** `src/lib.rs` exposes the engine, the runtime and the app API. The `yahaha` binary holds the terminal UI and the tools, behind the default `tui` feature. The app depends on `yahaha = { path = "../..", default-features = false }`, which pulls in neither ratatui nor crossterm. Every existing command still works.
- **One Session API.**
  - `Session::start(opts)` and `stop()` (also on drop).
  - `send(AppCmd) -> Result<(), CmdError>`.
  - `state() -> Arc<AppState>` and `version()`.
  - `subscribe()` returns a receiver of `Event::{StateChanged, LibraryChanged, Stopped}`.
  - `library()` / `library_list()`.
  - `Session` is `Send + Sync`. The synth's audio stream lives on its own thread for this.
- **`AppCmd` covers every user action:**
  - sections and transport, tempo and tap, Sync Start/Stop
  - fingering, detection area, Manual Bass, split, transpose
  - keyboard parts (on, voice, volume, octave, select), Style part mute and volume
  - fader page, pad page, master
  - OTS 1–4 and OTS Link, Stop ACMP
  - style load by id, by path, or by step
  - synth mute, output pair, panic
- **`AppState` holds everything the panel shows:**
  - style info and the voices it was written for
  - transport, with the section lamps as the pads show them
  - chord, keyboard parts, mixer with soft-takeover flags
  - pads per page, each with its meaning (`action: AppCmd`) and light
  - OTS, library status, I/O and synth, engine stats, messages
  - Everything serializes with serde: camelCase, and commands tagged `{"type": …}`.
- **Every mutation goes through `AppCmd`.**
  - The TUI no longer touches `Shared`, `Parts` or `SynthControl`. It draws `AppState` and sends commands.
  - Launchkey actions, including the Panel fader buttons, become the `AppCmd` their keyboard shortcut sends, and run on the session's control thread.
  - What stays on the MIDI thread, for real-time reasons and entirely inside the session:
    - engine buttons, straight to the engine ring as before
    - hardware faders with soft takeover
    - Pad Bank ▲/▼ and the fader-page button, so the next pad press or fader move already sees the new page
- **The Launchkey belongs to the session.** Its LEDs, OTS Link and library indexing run on the session's control thread, so the Launchkey works whichever client is running.
- **Offline session.** `Session::offline(opts)` runs with no CoreMIDI, audio or threads, on a virtual clock. It adds `advance(ns)`, `midi_in(port, bytes)`, `take_output()` and `finish_indexing()`, for tests and the app's dev mode. `yahaha state-json <style> ["C Am F"] [--library]` prints real `AppState` or library JSON to use as mock data.
- **Real-time path unchanged.**
  - The engine loop is now an `EngineLoop::step(now)` that the live thread and the offline clock share. The hot path is the same code.
  - One addition in the engine thread: a non-blocking semaphore signal to the control thread when the snapshot changes. It runs after the flush.
  - Still no locks or allocation on the engine, MIDI or audio threads.

## Behaviour
- **Screen output is identical.** `yahaha screen` before and after (single style, and folder with browser) produces the same HTML.
- **Live smoke tests.** In a pty, `yahaha play` with the synth responds to keys and quits cleanly on Esc Esc. `yahaha drive` against `play --all-inputs` runs through `Session::start` with real CoreMIDI.
- **One visible difference.** A Left fader button under Manual Bass now shows the same "Left plays the bass under Manual Bass…" message as the `8`/`l` key and the page-3 pad, because it runs the same command. The refusal itself is unchanged.

## Bench
`yahaha bench corpus/MOX_v2/TickingAway.T162.sty`, interleaved A/B, 3 rounds each. Before is `f91a3c9`, after is this branch.

| path | before p50 | after p50 | before p99 | after p99 |
|---|---|---|---|---|
| keyboard → passthrough out | 78.7–80.2 µs | 76.9–79.2 µs | 128–159 µs | 142–160 µs |
| chord → first band note (sync)* | 108–116 µs | 118–124 µs | (n=1) | (n=1) |
| chord published → engine applied | <6–8 µs | <7–8 µs | <13–28 µs | <11–16 µs |
| engine wake vs. deadline | <1 µs | <1 µs | <1 µs | <1 µs |
| engine work per wake | <2 µs | <2 µs | <8–10 µs | <8–9 µs |
| engine CoreMIDI send per wake | <1 µs | <1 µs | <8–11 µs | <7–9 µs |

\* The bench's sync-start stage only gets one sample on both builds. Its re-arm does nothing while the style is running, which predates this PR. Everything else is within run-to-run noise.

## Gates
- `cargo build --release` passes, and so does `--no-default-features`.
- `cargo test --release`: 314 + 6 pass, with the corpus present. This includes #70's new test after the rebase.
- New tests:
  - AppState, AppCmd, Event and LibraryList JSON round trips, plus the JSON shape
  - offline-session dispatch: sync start, sections and fills, position, tempo, chord settings, split and transpose (reaching the played notes), keyboard parts, mixer and soft takeover, pad pages and pad actions, OTS and OTS Link, library loading and errors, Launchkey pads and fader buttons, versions and events
- Clippy: no new warnings. 11 remain, down from 18. All are on pre-existing lines.

Closes #16


ryangavin:
## Adversarial review: Session API (#16)

Reviewed at `7115f25`. I fixed the blocking finding and three small ones in **`8c58824`**, which is signed and pushed to this branch. Gates on `8c58824`:
- `cargo build`, with and without `--no-default-features`: pass.
- `cargo test --release` with the corpus: 318 + 6 pass (4 new tests).
- Clippy: 11 warnings, all on pre-existing lines. None are new.

### Behaviour parity: OK
- I diffed the TUI keymap against `integration/m3-ui`. Every key maps to the same `Action`, and `From<Action> for AppCmd` maps each one to the handler with the same effect as the old UI loop. That includes transpose stepping, the Left refusal message, and style stepping that skips error rows.
- Tab, Shift+Tab, `a`, `k` and `\` map to the same effects.
- **New test** `launchkey_hardware_matches_its_commands`: 282 pairs, run from three starting states:
  1. stopped
  2. playing, with Upper, Manual Bass and OTS Link on
  3. playing Main C after an OTS recall
- Each pair sends the hardware message to one offline session and the matching `AppCmd` to its twin. It covers every pad on all three pages, Play/Stop/Scene/Function/Track, Pad Bank ▲/▼ with and without Shift, and all 9 fader buttons (Panel/Style × Shift). Resulting `AppState` and synth output are identical in every case.

### Findings

**1. Blocking (fixed): live `send` did not publish.**
- The doc says "`Ok` means the control side has applied the command", and says only engine commands show up "a few milliseconds later". In a live session, `send(SetSplit{70})` followed by `state()` returned split 54. The throwaway test against real CoreMIDI proved it.
- A Tauri `invoke('send')` then `invoke('state')` would flash the old value.
- Fix: `send` publishes under the control lock it already holds, then signals the control thread.

**2. Nonblocking (fixed): `LibraryList.revision` could be ahead of its entries.**
- While indexing, `library_list()` took its revision from the state (the current `lib_rev`), but its entries from the library published up to 250 ms earlier.
- Measured on the full corpus: revision 2 carried 208 pending entries while the state said 185. A client that dedups by revision keeps the stale list.
- Offline, the 250 ms gate runs on the virtual clock, so the list never republishes until indexing finishes.
- Fix: the library and its revision are published together. `state.library` (revision, count, position, pending) describes that same published list. Explicit changes (a path added, a style that fails to load) publish at once.

**3. Nonblocking (fixed): `stop()` race.**
- It took `live` before swapping the stop flag. A second, concurrent `stop()` could win the swap while the first held the threads. The first then returned without joining, turning off the LEDs, or disposing the client.
- Fix: swap first.

**4. Nonblocking (fixed): `CyclePadPage{delta: 127}` overflowed.**
- It panicked in debug builds (`i8` add in `Page::cycle`) and wrapped to the wrong page in release. Any client can send this.

**5. Nonblocking: `masterWaiting` is stale after `SetMasterVolume`.**
- The master takeover lives on the MIDI thread and only re-judges on the next fader CC. After a software set, `masterWaiting` stays false until the fader moves.
- Part and Style volumes do update at once. The master would need its hardware position in an atomic, as `Parts::fader_hw` has.

**6. Nonblocking: `stop()` leaves keyboard-part notes and pedals.**
- `engine.stop` only clears channels 9–16. Held Right/Left notes and sustain on channels 1–4 get no note-off or CC 123/64 on the port before the client is disposed.
- This predates the PR, but the session now owns shutdown.

**7. Nonblocking: device lifecycle.**
- There is no CoreMIDI notify handling. `io.inputs` and `pads.connected` are fixed at start. A Launchkey that is unplugged and replugged is not put back into DAW mode.
- `Session::start` twice in one process creates two `yahaha` virtual sources, and both drive the same Launchkey. There is no guard.
- An error after `Client::new` returns without `dispose`, so the virtual source leaks.
- The input handler, and with it `Shared` and its mach semaphores, is leaked per session, so a restarting app leaks each time.
- `Wakeup` has no `Drop`.

**8. Nonblocking: state tearing is benign.**
- `build_state` reads atomics that the MIDI thread writes (page, fader page, volumes) without a lock. A snapshot can mix values from either side of a concurrent change. The next pass, at most 10 ms later, corrects it.
- `version()` can briefly be ahead of `state().version`.
- Commands from the UI and hardware actions are serialised by the control lock.
- Page changes use a CAS (`step_page`). `toggle_fader_page` is load-then-store, so a UI toggle and a hardware toggle at the same instant can cancel. That predates the PR.

**9. Nonblocking: bench coverage.**
- `yahaha bench` calls `live::run_engine` directly. It measures the `EngineLoop` refactor, but not the control thread, the Launchkey path through it, or snapshot publishing.
- The numbers support "hot path unchanged". They don't cover the Session.

### Real-time safety: OK
- The MIDI thread gained only ring pushes, atomic swaps and `semaphore_signal`.
- The engine thread gained one `semaphore_signal`, after the flush and only when the snapshot changed. `Snapshot` is `Copy`.
- `AppState` is built only on the control thread, or on the caller's thread for `send` and offline use. It is never built on an RT thread. Measured live: about 2–3 `StateChanged` events per second while idle and while playing.

### AppState gaps the Tauri UI will hit
1. **No Shift state.** A mirror can't show the Shift layer: Shift+▲ = Left on/off, Shift+▼ = OTS Link, Shift+fader button = select part.
2. **Non-pad controls aren't described.** Pad Bank ▲/▼, Track ◀/▶, Play, Stop, Scene, Function and the 9 fader buttons have no label, `action` or LED state like `Pad` has. The app has to hard-code `cc_control` and `fader_button_msgs`. Suggest `state.buttons: Vec<Button>`, including Shift-layer actions.
3. **Track neighbours.** There is no prev/next style id or name. The UI has to reimplement `Library::step`, which skips error rows, over `library()`.
4. **No beat phase.** `session.beats()` is Rust-only. A webview flashing pads from its own clock will drift from the hardware. Suggest adding the beat clock to the state, for example `{beats, atNs}`, or a `beats` command.
5. **No physical fader positions.** Only the `waiting` flags are exposed. With no hardware value, the UI can't show which way to move the fader to pick up.
6. **`masterWaiting` is stale** (finding 5), and `pads.connected` / `io.inputs` are static (finding 7).
7. **Palette LEDs.** With `--palette-leds`, the hardware shows `pad_leds` colours and firmware flashing, not `looks`, so the mirror differs slightly.

### Serde: OK
- `AppCmd` and `Event` use `type`-tagged camelCase variants with camelCase fields. `CmdError` is `{"kind":"busy"}` or `{"kind":"failed","message":…}`. `Page`, `FaderPage`, `Fingering`, `Level` and `Anim` are camelCase.
- All of this matches the JSON in the doc, and the round-trip tests cover it.

**Verdict:** approve with `8c58824`. None of the remaining findings block. Items 1–5 of the AppState gaps are worth filing against #17 before the app mirrors the Launchkey.

PR #72, opened 2026-09-24
Tauri app foundation: Launchkey mirror, tooltips everywhere, design system (#17)

ryangavin:
This is the foundation for the desktop app (#17). It covers the Tauri shell on the real engine (#71), the tooltip system, the design system, and the main screen, which is a hardware mirror of the Launchkey 49/61 MK4. Keyboard parts + OTS, Mixer, Settings and the style browser are slots, each in its own folder, so they can be built in parallel. `app/CONTRIBUTING.md` explains how to build each slot.

## Run it
```bash
cd app && npm install
npm run dev                     # browser, mock session: http://localhost:5173
cargo tauri dev                 # desktop app on the real engine (repo corpus/ + soundfonts/, or YAHAHA_STYLES / YAHAHA_SF2)
YAHAHA_MOCK=1 cargo tauri dev   # desktop app on the mock (no MIDI or styles needed)
npm run verify                  # svelte-check (fails on warnings) + eslint + vitest
```

## Stack
The frontend is Svelte 5 + TypeScript + Vite, inside Tauri 2.
- **Updates:** Svelte compiles to direct DOM updates. At up to 60 state updates a second, only the values that change get touched.
- **Size:** no framework runtime, and CSS is scoped per component.
- **Offline:** the fonts are bundled through @fontsource, so the app never reaches for a CDN.
- **Look:** everything is drawn in CSS, with no image assets.

## Main screen: the Launchkey mirror (`app/src/panels/launchkey/`)
The layout follows the MK4.
- **Status display:** sits where the screen is. It shows the style, tempo, bar/beat, the section playing and the one queued, a big chord, fingering, chord detection, split and transpose.
- **Pads:** the 2×8 pads are lit from the engine's `Pad` looks on a beat clock (dim, bright, flashing, pulsing, dark). They show the current page's labels, and clicking one sends `pad.action`.
- **Buttons:** Pad Bank ▲/▼ sit left of the pads, lit in the page's colour. Tempo +/− (Scene Launch/Function) sit to the right, then Stop and Play. Track ◀/▶ are captioned with the neighbouring style names.
- **Shift:** the on-screen Shift latches it, and holding Shift on the computer keyboard also shows it. The Shift layer turns Pad Bank into Left on/off and OTS Link, and the Panel page's fader buttons into Edit.
- **Faders:** 8 faders + master, with a slot, cap, tick scale and readout. The labels and levels follow the active fader page. While a level waits for its Launchkey fader, it shows the soft-takeover mark and the hardware fader's position. The fader buttons mute or unmute the part, and the master button switches Panel/Style.
- **Nothing hard-coded:** no button function lives in the UI. Pads come from `state.pads`. Every other control comes from a provisional `SurfaceState`, which holds the label, action, Shift label/action and LED of every control, plus fader positions, Track names and the beat clock. `surfaceOf(state)` derives it until the API follow-up adds `state.surface`, and the mock already sends it.
- **Scaling:** the surface scales with the window. Below 820 px wide, the fader bank moves under the pads.

## Tooltips
- **Catalog:** `app/src/help/tooltips.ts` is the single source of truth. Each entry has a title, 1–3 plain sentences, the Genos name, the keys in README notation, and the Launchkey location. `app/docs/controls.md` is generated from it.
- **Mirror tooltips:** every mirror element's tooltip is its function on the current page and Shift layer.
- **Behaviour:** a tooltip shows after 280 ms of hover, or at once when you move between controls or tab onto one. It never takes pointer events. `?` turns on help mode, which pins the full entry in a bar at the bottom.
- **Tests:**
  - Every interactive element, in 13 app states, has a catalog key.
  - Every entry is complete.
  - Catalog keys match README's terminal keys, both ways.
  - Catalog keys match the app's key bindings, both ways.
  - Each pad's entry names that pad's location.
  - `controls.md` is up to date.

## State layer and shell
- **TypeScript types:** `app/src/lib/api/types.ts` mirrors `docs/app-api.md`.
- **Shape check:** `engine-shape.json` records the key paths of the real engine's JSON, taken from `yahaha state-json` with no style data. A TypeScript test and a Rust test check the mocks against it.
- **Tauri:** `app/src-tauri` depends on `yahaha` (`default-features = false`) and serves `send`/`state`/`library` plus `yahaha` events. The frontend coalesces `stateChanged` to one fetch per frame.
- **Mock fallback:** the shell falls back to a Rust mock built on the engine's own types if the engine can't start, or with `YAHAHA_MOCK=1`.

## Design
The look is a restrained skeuomorphic instrument, built from:
- anodised/brushed metal (`.mat-chassis`)
- recessed wells
- backlit raised buttons
- display glass with glowing text
- rubber pads whose LED glow bleeds onto the panel
- engraved labels

Tokens and material recipes live in `app/src/app.css`, and CONTRIBUTING.md documents them. It has dark and light themes. LEDs and fader caps animate with opacity and transform only.

## Verification
- `npm run verify` is clean: 147 tests pass.
- In `app/src-tauri`, `cargo test` and `cargo clippy` are clean.
- `cargo tauri build --debug` builds `yahaha.app`. The earlier build rendered live over IPC; this build launched and ran on the mock.
- The repo's `cargo build --release` and `cargo test --release` pass (318 tests). Nothing outside `app/` changed.
- I checked the browser UI at 1920, 1440, 1280, 1024 and 375 wide. At 375 there's no horizontal overflow. Hover and keyboard tooltips, the shortcuts, pad clicks, Shift, fader pages and pickup all work.

## Not done yet / worth checking
- I built the mirror's layout from the MK4 programmer's guide diagrams and the CC map. They don't include the faders, so the fader bank's position (right of the transport) should be checked against the real 49/61.
- The encoders, Settings, Arp/Scale and Rec/Loop controls are left out because yahaha doesn't use them.
- `SurfaceState` field names are my proposal. When the API follow-up lands, rename them in `types.ts`/`surface.ts` if they differ.

## Screenshots (mock mode)
Laptop, 1440 × 900:
laptop

Large, 1920 × 1080:
large

Tooltip on a pad:
tooltip

Shift layer:
shift

Help mode:
help

Light theme:
light

Laptop, 1280 × 800:

Narrow, 1024:


ryangavin:
## Adversarial review: PR #72 (at 51ac70d, fixes pushed in 4d4f6dc)

**Gates at 4d4f6dc:** `npm run verify` passes (svelte-check 0/0, eslint, 148 vitest tests). `cargo test` in `app/src-tauri` passes (8, two of them new). `cargo build --release` and `cargo test --release` at the repo root pass (318 + 6). Clippy on the Tauri crate is clean. The commit is signed.

### Blocking (all four fixed in 4d4f6dc)

1. **The Launchkey stayed in DAW mode after the app quit.** Nothing in `app/src-tauri` ever called `Session::stop()`. `Drop` doesn't cover it:
   - Tauri ends the process with `exit()`, so managed state is never dropped.
   - The `session-events` thread holds its own clone of the `Arc<Backend>`, so the session can't be dropped either.

   As a result, the pads, the fader-button lights and DAW mode all stayed on, and the audio and MIDI clients were never closed.

   **Fix:** `Builder::build()` + `app.run(|_, RunEvent::Exit| shutdown(..))`, with `shutdown` stopping a live session. `stop()` sends the lights-off messages and `EXIT_DAW`, then emits `Stopped`, which also ends the forwarder thread. A new test, `shutdown_stops_a_live_session`, drives an offline `Session` through `shutdown` while a second `Arc` clone is held, and checks that `Stopped` arrives and that a second `shutdown` is harmless.

2. **The layout didn't match the hardware: the faders were on the wrong side.** On the MK4 49/61 the nine faders and their buttons sit **left** of the screen, between the octave buttons and Shift/Track. The transport block is at the far right, right of Scene/Function (User Guide p. 9, items 3, 4, 11 and 12). The mirror had the fader bank at the far right, after the transport.

   **Fix:**
   - The fader bank is now the first grid column, and first in the DOM, so tab order also runs left to right.
   - The transport stays right of Scene/Function.
   - Added a DOM-order test.
   - Regenerated the screenshots.

   The narrow (<820px container) stacked layout is unchanged.

3. **The fallback to the mock passed for a working rig.** If `Session::start` failed, or no corpus was found (for example, a bundled app on another machine), the shell quietly ran `MockSession::new()`. That mock reports:
   - `pads.connected: true`, so the mirror shows a green "Launchkey connected".
   - A synth on "MacBook Pro Speakers" and Launchkey inputs.
   - `io.offline: false`.

   `app.kind` is `'tauri'`, so no badge appeared either. The user saw a band playing on screen with no sound and a Launchkey that did nothing, and no hint why.

   **Fix:** `MockSession::fallback(reason)` sets `io.offline = true` (the header then shows "offline session"), no Launchkey, no inputs and no synth, and puts the reason in the status line. `YAHAHA_MOCK=1` still gives the full demo rig. New test: `the_fallback_mock_says_it_is_not_the_engine`.

4. **The Break tooltip was wrong.** It said "a one-bar break at the next bar". The engine queues Break with `queue_fill`, so it starts at the **next beat** and plays to the end of the bar (`src/engine.rs`, "Fills and breaks start at the next beat…"). I reworded it and regenerated `docs/controls.md`.

### Nonblocking

5. **The UI works out button behaviour itself.** `lib/surface.ts` `deriveSurface` re-implements the `src/launchkey.rs` rules in TS: what every button does on each layer, and every button light. It is documented as a stopgap until the engine sends `state.surface`, and the seam is clean because the UI only reads `surfaceOf()`. It has already drifted:
   - The Play button lights green while running. The engine never drives the Play LED.
   - Track ◀/▶ are lit by `neighbours()`, which skips unreadable styles. The hardware lights them when `lib.len() > 1`.

   Until the engine sends the surface, add a parity test that runs `state-json` through `deriveSurface` against the Rust rules, or ship `surface` in the #71 follow-up and delete `deriveSurface`.

   Pads are fine: they come from the engine's `looks()` in RGB mode, the default, so screen and hardware agree on every page by construction. I also checked the browser mock's `mock-pads.ts` against `looks()` for all three pages, and it matches. Separately, and not in this PR: the engine's palette-mode `pad_leds` (`--palette-leds`) differs from `looks()`. For example, Sync Start, Auto Fill and Sync Stop are OFF when off rather than dim, and Start/Stop is dim red when stopped rather than bright. In that mode the screen and the pads disagree.

6. **Tooltip coverage is real.** I deleted the `transport.sync_stop` entry: 27 tests failed (coverage in every state, plus the controls.md check). I removed `use:tip` from the chord display: 13 tests failed. The one gap is a clickable element with no role or tabindex, which the checker can't see. Svelte's a11y lint (`--fail-on-warnings`) catches most of those.

   I spot-checked 18 entries against `src/launchkey.rs`, `parts.rs`, `engine.rs` and README, covering pad positions on all three pages, the Scene/Function tempo buttons, Shift + Pad Bank, the part channels (1/3/4/2), `PICKUP_RANGE` = 2, the transpose-lamp conditions, OTS dark when absent, and Manual Bass dark in Lower. All were correct except Break (#4).

7. **Performance.** Measured on the mock at 1440×900.
   - **Browser pane (GPU, dpr 2):** 301 frames in 5 s, p50 16.7 ms, p99 17.7 ms, max 17.7 ms, no long tasks or long animation frames. Mutations: 13/s playing, 61/s with a pulsing Sync Start.
   - **CDP trace, headless software raster, demo playing:** 60 fps, Paint 3.5/s, Layout 1.8/s. No layout thrash.
   - **Pulsing lamp:** only **38 fps**, with RasterTask 75/s. The `.bleed` glow is a blurred `box-shadow` whose opacity changes every frame, so it was re-rastered each frame. I put `.bleed` on its own layer (`will-change: opacity`): 60 fps, raster 3.2 ms/s instead of 9.5. That change is included.

   There is no CSS `filter: blur` or `backdrop-filter` anywhere.

8. **IPC permissions were `core:default`** (window, app, path, menu, tray, image, resources…). The webview only needs `listen`/`unlisten`, and the app commands need no permission. I reduced it to `core:event:default`, which is included in 4d4f6dc. I couldn't launch the GUI here, so please confirm once with `cargo tauri dev` that state events still arrive.

9. **The CSP exists and is tight:** `default-src 'self'`, `connect-src ipc: http://ipc.localhost`. In production builds `style-src 'unsafe-inline'` may be droppable, because Svelte extracts CSS and `style:` directives use the CSSOM.

10. **Event coalescing works:** at most one `state` fetch per frame. Two nits:
    - The rust side forwards every engine event, up to about 100/s. That's fine.
    - `schedule()` clears `pending` before the `invoke` resolves, and accepts any `st.version !== last.version`. If two fetches ever resolved out of order, an older state would overwrite a newer one. Sync commands run in order on the main thread today, but `st.version > last.version` would make that safe regardless.

11. **Shutdown gaps outside `RunEvent::Exit`:**
    - Ctrl+C or SIGTERM on `cargo tauri dev` still skips it, so the Launchkey stays in DAW mode. A signal handler would cover that.
    - An engine-side edge case: in `Session::start`, if a thread fails to spawn after `ENTER_DAW` has been sent, the device is left in DAW mode.

12. **Assets:**
    - I found no Yamaha or Novation logos, images or fonts. The mirror draws its own controls, and "Launchkey" appears only as plain text naming the device.
    - Barlow and Barlow Condensed come from `@fontsource` under **OFL-1.1**, which is fine to bundle.
    - `src-tauri/icons/*` are still the stock Tauri logo, a placeholder to replace before any release.

13. **Layout and space.**
    - **1024 wide:** `--u` hits its 10.5px floor, so labels truncate ("Temp…", "Encor…") and the pads are small. About 55% of the height is empty.
    - **1440 and 1920:** the mirror fills the width but only about 40% of the height. The owner's screenshot shows the empty space above and below.

    Suggestions:
    - Size `--u` from `min(width / 96, height / 45)`, using container query units on a container with a defined height, so the mirror grows on tall windows.
    - Above the mirror, add a lead-sheet or "what's next" strip: chord history, the next section queued, and later lyrics or a chord chart.
    - Below it, draw a keyboard strip mirroring the Launchkey keys. It would show held notes, the split point, and the chord-detection area (Lower/Upper). That is the most useful missing piece of the hardware mirror, and it would make Split −/+ and Upper/Lower visible at a glance.

**Verdict:** approve once 4d4f6dc is green. The blocking items (#1–#4) are fixed in that commit. Items 5–13 can follow up.

ryangavin:
The engine now sends `state.surface` in #77. Contract: docs/app-api.md on `m3/launchkey-mirror`, sections `surface` and `surface.clock`. It follows your provisional `SurfaceState`: the same control ids, 9 faders, `shiftLabel`/`shiftAction` equal to `label`/`action` where Shift changes nothing, and `set` commands with `volume: 0`.

Where it differs:
- **`trackPrev` / `trackNext`** are `{ id, name, path }` objects, not strings. Use `.name` for the label.
- **Controls have two extra fields:** `cc`, and `colour` (the palette index, or null).
- **Play, Stop, Scene and Function** report `level: 'off'` and `colour: null`. yahaha doesn't drive those LEDs, so it has nothing to report about them. `deriveSurface` shows Play as green while running, but the hardware doesn't.
- **Labels are engine-side, in hardware style:** `PAGE ▼`, `RIGHT 1`, `EDIT R1`, `PAD`, `MASTER`, `PANEL`. A Style fader button's muted state is in `level` (dim), not in the label.
- **An unused fader still reports `position`.** Its `label` is `''`, and `value` and `set` are null.
- **`clock` adds fields:** `running` and `beatsPerBar`, plus two anchors, `sectionAnchorMs`/`sectionAnchorBeats` and `ledAnchorMs`/`ledAnchorBeats`.
  - `atMs` is the session's monotonic clock in ms, not `performance.now()`. Extrapolate with `t = atMs + (performance.now() − receivedAt)`.
  - The `led` beats drive pad flash and pulse, in sync with the hardware.
  - Have the Tauri `state` command return `session.state_now()`, so `atMs` is fresh at fetch time.
- **Also new:** `pads.paletteLeds` and `Pad.palette` (palette-LED mode), and a `fader` field on each keyboard part and Style part.
- **Fixtures:** `docs/fixtures/state.json` and `library.json`, from `yahaha state-json`, can serve as mock data.

Issue #33, opened 2026-09-23
Arpeggio engine (our own patterns)


ryangavin:
_Details in the spec._

---
Milestone: **M6 · Parts, voices & live play** · Spec: [docs/genos-features.md](../blob/main/docs/genos-features.md) · Backlog: [docs/backlog.md](../blob/main/docs/backlog.md)

ryangavin:
Wiring is in PR #100. Its Decisions section records the choices the manuals leave open. For the arpeggio: Assign and Volume are shared with Harmony, it runs on its own clock while the style is stopped (the engine clock running on at tempo), timing is in 1/8-tick sub-ticks, set_ppq is called on a style change, and all_off is called on switch-off, type change, PANIC and shutdown. Velocity mode and Keep Key On are labelled as yahaha extensions. The Live Control percentages and Hold on a pedal are not wired yet.

ryangavin:
Done in #100, merged to main via #113.

PR #75, opened 2026-09-24
Arpeggio engine core with original patterns (#33, part 1)

ryangavin:
Part 1 of #33: the arpeggio engine core. It is a pure, tempo-synced, deterministic module in `src/arp`, and it is **not wired in yet**. There are no changes to live.rs, engine.rs, session.rs, api.rs or ui.rs. The only shared-file edit is `pub mod arp;` in src/lib.rs.

## API
```rust
let mut arp = Arp::new(ppq, library::find("Climb 16").unwrap().clone());
arp.set_settings(settings, tick);   // also arp.set_hold(on, tick)
arp.note_on(note, vel, tick);       // vel 0 = off
arp.note_off(note, tick);
arp.set_sustain(down, tick);
arp.set_pattern(pattern, tick);     // cuts sounding notes and restarts at step 1 (on the grid if quantized)
arp.process(from..to, &mut sink);   // impl ArpSink { note_on(tick, note, vel); note_off(tick, note) }
arp.stop(tick); arp.all_off(tick, &mut sink);
```
- `process` does not allocate: all state lives in fixed arrays. `tests/arp_no_alloc.rs` checks this with a counting global allocator in its own test binary.
- Note-offs always balance. The engine tracks each pitch it starts, releases a retriggered pitch before restarting it, and cuts sounding notes on pattern change, stop and all-off.
- Step times are computed from an anchor rather than accumulated, so they do not drift, and the output does not depend on how the tick ranges are chunked.

## Pattern model
A loop of steps. Each step has a `sel` (`Rest`, `Walk`, `Idx(i)`, `Top(i)`, `All`, `Strum{up, spread}`), an octave, a velocity and a gate %. Each pattern also has a step length, swing, walk motion (up/down/up-down/down-up/seeded random/as-played), octave span, sort order and seed. It is serde-serialisable, so TOML patterns can be added later.

## Starter library: 23 original patterns
None comes from Yamaha preset data.
- Up & Down: Climb 16, Fall 16, Peak 8, Valley Triplet, Sky Ladder 32
- Random: Dice 16, Scatter Octaves
- As Played: Echo Order 8, Shuffle Order 16
- Chord stab: Four Stabs, Offbeat Pump, Syncopated Hits, Gated Pad 16
- Broken chord: Alberti 16, Waltz Broken, Rolling Eights, Thumb Pick
- Guitar: Strum Quarters, Campfire Strum, Muted Sixteens
- Sequence: Octave Pulse, Root Fifth Seq, Pluck Line

## Settings
- Quantize (Off/8th/16th; starts on the nearest grid line)
- Hold (latch; a new chord after a full release replaces the old one)
- Velocity (Original/Thru/Fixed)
- ArpVel, ArpGateT and Unit Multiply (percent)
- Keep Key On (the clock runs through releases, so the phrase keeps its phase)
- Sustain pedal holds arp

Our definitions where the manuals are silent are in docs/arpeggio.md.

## Tests
41 unit tests plus 1 integration test:
- playback for each category
- tick accuracy at 1920 PPQ and with fractional Unit Multiply
- chunk invariance
- latch, pedal and Keep Key On
- swing and quantize
- velocity modes
- pattern-change cuts and legato retriggers
- a 40-seed fuzz of random operations that checks note-offs balance
- every pattern × 1–6 held notes balanced

`cargo test`: all pass. `cargo clippy --all-targets`: no new warnings.


ryangavin:
## Adversarial review: PR #75, arp core

I reviewed 69be55c. Build, all tests and clippy passed, and the PR adds no new clippy warnings. I proved each finding below with throwaway tests. **I found three blocking issues and fixed them in 78ed568, which is pushed to `m6/arp-core`.** The regression tests are now in `src/arp/tests.rs`.

### Blocking (fixed in 78ed568)

**B1. A clock that goes backwards leaves notes stuck and the arp silent.**
The engine resets its tick to 0 on every style start (`engine.rs` `start()`: `anchor_tick = 0.0`).
Test: I started Four Stabs at tick 100000 and then processed `0..5000` with the keys still held. The result was **0 events, and 3 notes still sounding**. Their offs, and the next step, were due at about tick 100192 and 100480, so the arp stays silent and the notes stay stuck for about 100k ticks.
Fix: `process()` remembers where the last range ended. If a range starts before that point, it cuts every sounding note at `range.start`, drops the queued note-ons, and restarts a running pattern from step 1 there, or on the next grid line when Quantize is on.

**B2. A gap in the clock sends a burst of every missed step, with unbounded CPU cost on the RT thread.**
Test: I ran Sky Ladder 32 with a triad, then jumped from tick 100 to `96000..96010`. That emitted **3197 events, all stamped 96000**, and the loop cost grows linearly with the gap.
The same thing happens on a smaller scale with a late quantized start. Sky Ladder 32 with Eighth quantize can snap back several 32nd-note steps, and all of them fire together.
Fix: `skip_late_steps` jumps straight to the latest step due at `range.start` and plays only that one. The existing behaviour of "one late step plays now" (`quantize_starts_on_the_nearest_grid_line`) is unchanged. Empty and reversed ranges now do nothing. Before the fix, `500..100` emitted a note-on at 500, which is outside the range.

**B3. An unvalidated pattern panics or hangs.**
`Pattern` has public fields and is serde, and `set_pattern` does not call `validate()`.
- With `steps: []`, `play_step` panics with a modulo by zero (`k % steps`).
- With `step_len: 0`, `step_num()` is 0, so every step falls on the same tick and `process()` never returns. My test hit a 3-second timeout.

Fix: an empty pattern plays nothing, and a zero step length counts as 1 tick at 480 PPQ. `tests/arp_no_alloc.rs` now also covers a backwards jump, a gap and a `set_pattern`.

### Nonblocking

**N1. u64 edges.** `tick + gate` overflows at ticks near `u64::MAX`, which panics in debug builds. It cannot happen in practice: at 1920 PPQ and 300 BPM that tick is billions of years away. The `NONE = u64::MAX` sentinel would also collide there.

**N2. Overlapping notes on the same key.** The sequence `on C, on C, off C` removes C, because there is no per-key count. The output stays balanced, and this matches normal MIDI practice. live.rs should not send duplicate note-ons. A `note_off` with no matching note-on is ignored, which is correct.

**N3. Timing.**
- Triplets are exact at 96, 480 and 1920 PPQ.
- Unit Multiply at 300% and at 133% does not drift.
- Output is identical with 1-, 13- and 4000-tick chunks while I changed unit multiply, gate, pattern, hold, release and stop in the middle of the run.
- Swing 67 on eighths at 1920 PPQ puts the offbeat at 1286, against 1280 for a true triplet (0.5%). The docs say "about", but a finer swing unit or a triplet flag would make it exact.

**N4. The allocation test is real.** `GlobalAlloc`'s default `realloc` and `alloc_zeroed` go through the counted `alloc`, so a growing Vec would be caught. The arp has no heap containers, so nothing can slip through today. Coverage was narrow before this change: there was no `set_pattern`, `stop`, Velocity mode or Quantize. I added a jump, a gap and a pattern swap. Other threads in the test harness could cause a false positive, but not a false negative.

**N5. Plausibility against the Genos manual.**
- **Quantize (RM p.41):** "synchronized with Song or Style playback… imperfections are corrected". This matches the Eighth and Sixteenth grid. The manual does not say whether to use the nearest or the next grid line, so the choice is fine.
- **Hold (RM p.41, p.141 footswitch):** playback continues after release and stops when Hold is turned off. This matches. Notes that are sounding at that point finish their gate.
- **ArpVel, ArpGateT and ArpUnitM (OM p.57, Live Control):** each is a percentage of the type's default. This matches.
- **Velocity (Original, Thru, Fixed) and Keep Key On** do not appear anywhere in the Genos manuals. They are Motif/MONTAGE concepts. They are fine as extensions, but docs/arpeggio.md should label them as yahaha extensions, not Genos behaviour. ArpVel at 0% clamps to velocity 1 rather than silence, which is reasonable.

**N6. Readiness for wiring. These need to be handled in part 2.**
- **Fixed PPQ.** PPQ is fixed at `new()`, but styles have different PPQs. Rebuilding the `Arp` on a style change throws away the off tracking and leaves notes stuck. Either add `set_ppq`, or call `all_off` before rebuilding.
- **Tick domain.** The engine tick is an `f64` from `tick_at(now)`, so wiring must floor it to a u64 in the same way on both the input and output sides, and convert output ticks back to time with `ns_at`.
- **Arp without a running style.** The Genos arp runs with the style stopped, and the engine's tick only means something while the style runs. The arp therefore needs a free-running tempo clock while the style is stopped. The new clock-reset handling covers the handoff to tick 0 when the style starts.
- **MIDI input timestamps.** If the engine has already processed past a key's tick, its first step goes out at `range.start`, one period late. That is acceptable.
- **Stop.** `stop()` only sends the offs on the next `process`. If the engine thread stops calling `process`, for example when the button goes off or on shutdown, use `all_off`.
- **ACMP.** Only right-hand keys should reach the arp in split mode. With Hold on, left-hand chord changes must not clear the latch. The Arp API cannot tell the difference, so the routing has to live in live.rs. Also decide whether the arp follows the recognized chord when the ACMP split is off.
- **Pattern changes.** `set_pattern` uses the next grid line (`div_ceil`), while a first key uses the nearest one. This is documented and reasonable.

**N7. Content.** There is no Yamaha pattern data. All 23 patterns are short hand-written step lists. None of the names matches the Keyboard Harmony/Arpeggio type list in the Data List. The only text hits are the voice `CampfireGuitar` and "Speaker", neither of which is an arpeggio type.

### Verdict
After 78ed568 I **approve**. I have not merged.

Issue #32, opened 2026-09-23
Keyboard Harmony (our own implementation of the documented types)


ryangavin:
- Includes chord source rules for each ACMP/LEFT combination (§6).

---
Milestone: **M6 · Parts, voices & live play** · Spec: [docs/genos-features.md](../blob/main/docs/genos-features.md) · Backlog: [docs/backlog.md](../blob/main/docs/backlog.md)

ryangavin:
Wiring is in PR #100. Its Decisions section records the choices the manuals leave open. For Harmony: the chord source (the Style chord for every ACMP/LEFT combination, because there is no ACMP switch), which key counts as the melody, no re-voicing on a chord change, how switching behaves, and the Echo-category struck note going out on the engine thread.

ryangavin:
Done in #100, merged to main via #113.

PR #76, opened 2026-09-24
Keyboard Harmony core (#32, part 1)

ryangavin:
Part 1 of #32. This adds the pure Keyboard Harmony module `yahaha::harmony`. It is **not wired** into live/engine/session/api/ui; a later ticket connects it. The only shared-file edit is one `pub mod harmony;` line in `src/lib.rs`.

## What's in it
- **`voice(ty, melody, chord)`**: our own voicing for every Data List Harmony type:
  - Standard Duet 1/2, Standard Trio, Full Chord, Rock Duet
  - Country Duet 1/2, Country Trio
  - Block, 4-Way Close 1–4, 4-Way Open 1–3
  - 1+5, Octave, Strum

  Voicings are deterministic, return notes highest first, and stay on the side the type specifies. They never put a semitone against the melody. No Yamaha data is used.
- **Chord source (§6):** `chord_zone(acmp, left, style_split, left_split)` and `harmony_chord(...)` are pure functions covering all four ACMP/LEFT combinations.
- **Detail settings:** `harmonize()` and `route()` implement:
  - Volume
  - Assign (Auto / Multi / Right 1–3). A Mono part counts as off in the Harmony category.
  - Chord Note Only
  - Minimum Velocity
- **`HarmonyTracker`:** note-off returns the notes each melody key added. **`MultiAssign`:** assigns keys to parts in press order.
- **`EchoGen`:** the timing generator for Echo, Tremolo and Trill. Its API is `note_on`, `note_off`, `set_tempo`, `next_due()` and `next_events(now, &mut [EchoEvent]) -> usize`, with times in engine nanoseconds.
- Real-time safety: the per-note path uses only fixed-size arrays, never allocates, and never panics.
- `docs/harmony.md` lists every type's rule and marks each guess.

## Tests (29)
- A table-driven test for every type across C, Cm, C7, Cdim and Caug, with the melody both on and off the chord. The same table is also checked transposed to all 12 roots.
- Invariants over every root × all 38 chord types × all 128 melody keys: determinism, side, no semitone clash, chord-tone-only types, and Cancel.
- Hostile input produces no panic.
- Chord source table, Chord Note Only, Minimum Velocity, Volume, Assign routing, tracker and Multi Assign.
- Echo timing: tremolo period and gate, echo decay and fade-out, trill alternation and pair changes, minimum-velocity bypass, tempo change, buffer limits and time ordering, all-off, and voice overflow.

`cargo build --release` and `cargo test --release` pass (347 lib tests). There are no new clippy warnings.

## Needs a hear-test on a real Genos
The manuals don't give voicings, so these are guesses; `docs/harmony.md` has the checklist:
- Duet 2, Rock Duet and the Country types
- 1+5 (above?) and Octave (below?)
- What the 4-Way Close/Open variant numbers mean
- Block and Close 4 melody doubling, and the extra root in Full Chord
- Strum direction and timing
- Echo decay and whether it stops at release
- Trill start order and what happens to older held keys
- Whether Assign Auto puts the effect on one part or on all parts that are on
- Speed values, the Volume scale and gate lengths


ryangavin:
## Adversarial review: PR #76 (Keyboard Harmony core), reviewed at 6d299d5

I proved every finding below with a throwaway test (13 cases: scans over every root × all 38 chord types × all 128 melody keys, randomised press/release/tempo/settings/all_off runs over Echo, Tremolo and Trill, and a counting-allocator run of the per-note path).

### Blocking (fixed in 5735b51, see the next comment)

1. **BLOCKING: the PR's own test suite fails in a debug build.** `harmony::tests::no_panic_on_hostile_input` panics with an add overflow at `src/harmony.rs:590`: `(chord.root + f) % 12` in Rock Duet overflows when root is 255. The PR body says "`cargo test --release` passes", but release builds wrap on overflow, so the test only passes there. Plain `cargo test` fails: 346 passed, 1 failed.
2. **BLOCKING (stuck note): Trill's note-off can carry the wrong `effect` flag.** `stop_trill` always sends `effect: true`. The trill's first note is the struck note, which is sent with `effect: false`. Suppose you release one of the pair, press a third key, or hit all_off while that first note is sounding. The note-on then goes to the melody parts and the note-off goes to the effect parts, which is the routing the docs tell the engine to use with Assign = Multi. That leaves a stuck note. A randomised run found about 40 keys with an unbalanced on/off per (key, effect) route in one seed. The existing test `all_off_silences_everything` asserted this buggy flag.
3. **BLOCKING (stuck notes): EchoGen's pending queue silently drops note-offs.** `emit` discards events once 64 are queued. Test: 40 plain (below Minimum Velocity) presses and releases between two polls leave **16 keys stuck**. Nothing in the API says the engine must drain after every call, and one MIDI burst on the input thread is enough to trigger it.
4. **BLOCKING (wrong notes): 4-Way Close 3 adds a natural 9th on 7(b9) and 7(#9).** `four_part(.., ninth)` puts in the 9th unconditionally. With C7(b9) and E on top it gives `[62, 61, 58]`, a D and a Db together. On C7(#9) it gives D against D#. Neither is a chord tone or a valid scale tone for those chords.
5. **BLOCKING (voicing convention): the drop voicings create minor 9ths between voices.** 4-Way Open 2 (drop-3) and Open 3 (drop-2&4) open a close semitone into a b9 on every maj7-family chord (M7, M7(#11), M7(9), mM7, M7aug, M7b5, mM7b5), on m(9)/m7(9)/mM7(9), on 7(13) and on 7(#9). Example: Cmaj7 with E on top gives E C G B(47), and B47 sits a b9 under C60. Standard jazz practice avoids a b9 between voices on anything except a dominant b9 chord. The PR's "no semitone clash" invariant only checks each voice against the melody.

### Nonblocking

6. **Panic on absurd tempo.** `period_ns` saturates to u64::MAX for tiny bpm (say 1e-300), and then `period * 3 / 4` overflows in debug. Fixed along the way (the division now comes first).
7. **`next_events` does not return events in time order**, although its doc says it does. Queued immediate events come out before scheduled ones that were due earlier, whenever the engine polls late. Each event has the right `at`, so the harm is limited, but either fix the doc or merge by time.
8. **Trill at Volume 0 still sounds its repeats at velocity 1** (`.max(1)`), while docs/harmony.md says "At 0 the effect is silent". Echo and Tremolo stop instead. Also, lowering Volume to 0 mid-note makes Echo/Tremolo send a pulse with `vel = 0` while it marks the voice as sounding. The engine then gets a double note-off. That's harmless, but untidy.
9. **HarmonyTracker is not a sounding-note refcount.** Two held melody keys can add the same harmony key, and a harmony note can equal another held key. Examples: hold G, which adds E below, then play E; or re-voice after a chord change. The first note-off then cuts a note that should keep sounding, or two note-ons stack on one key. The wiring ticket needs per-(part, key) reference counting.
10. **Wiring readiness.** EchoGen is pull-based and counts in engine nanoseconds. The PR #75 `Arp` is push-based (`ArpSink`) and counts in style ticks at `ppq`. Both can be driven from `live.rs` (`rt::now_ns`, and the engine has `ns_per_tick`/`tick_at`). They are mutually exclusive on the Genos: one [HARMONY/ARPEGGIO] function and one type selector (OM p.56-57: "select a category other than Harmony or Echo" for Arpeggio). Nothing in either API expresses that. The wiring ticket should add one `KbdHarmonyArp { Off, Harmony(HarmonySettings), Arp(arp::Settings, Pattern) }` slot, with one switch path that calls `EchoGen::all_off`, the tracker's `release_all` and the arp's stop. EchoGen repeats also run free from the key press, while the manual says "in time with the currently set tempo". Consider an option to snap to the style grid like Arp Quantize.
11. **No allocation on the per-note path.** Confirmed with a counting global allocator over 1000 harmonize/tracker/MultiAssign/EchoGen presses, polls and releases: 0 allocations. The side and range invariants also hold for all 23 types × 12 roots × 38 chord types × 128 keys. No note goes outside 0-127, and nothing crosses the melody against its documented direction.

### Manual consistency: no contradictions found

I checked the 11 guesses against OM p.56-57, RM p.46-47 and the Data List harmony list:
- **Type list and order:** matches the Data List.
- **Assign Auto:** a single, prioritised part is consistent with RM: "parts are prioritized in the following order: Right 1, Right 2, Right 3".
- **Volume as a velocity scale:** backed up by the RM note that Volume has no effect on voices with Touch Sensitivity Depth 0.
- **Inclusive Minimum Velocity:** matches RM's "the lowest velocity value at which the ... note will sound".
- **Trill:** "the last two notes" matches.
- **Multi Assign:** the manual has no detail settings for it, and neither does the PR.
- **Mono parts:** counting a Mono part as off in the Harmony category matches.
- **Chord source table:** matches all three OM diagrams.

The manuals say nothing about the voicings, Speed values, Echo decay, Strum or the ACMP-off/LEFT-off case, so those remain hear-test items as listed.

### After the fixes
Build is clean and all tests pass: 353 lib tests, including 6 new regression tests. No new clippy warnings in harmony.

ryangavin:
## Fixes pushed: 5735b51 (on top of 6d299d5)

This commit fixes blocking items 1–5 and nonblocking item 6 from the review above.

- **Rock Duet:** the root is now taken mod 12 before the fifth is added, so a debug build no longer panics.
- **`stop_trill`:** the note-off's `effect` flag is now `t.count > 1`, the same as its note-on. I updated `all_off_silences_everything`, which asserted the old, wrong flag.
- **`EchoGen::emit`:** a note-off whose note-on is still queued now cancels that note-on. The queue is therefore bounded by the voices (at most 17 ons plus 17 offs, under the 64 slots), and a burst no longer loses note-offs.
- **Gates:** computed as `period / den * num`, so they can't overflow.
- **`four_part`:** with Close 3 on a chord that has a b9 or #9, the chord's own altered 9th stays as the rootless tone (3‑5‑b7‑b9 or 3‑5‑b7‑#9), and no natural 9th is added.
- **`drop_voices`:** a drop that would make a minor 9th against another voice (melody included) is skipped, so that voice stays in close position. Every existing table row is unchanged, including C, Cm, C7, Cdim and Caug on all the Open types.
- **docs/harmony.md:** updated for the Close 3 and Open rules and for the EchoGen queue contract.
- **New regression tests:** `rock_duet_takes_any_root_byte`, `close3_keeps_an_altered_ninth`, `drop_voicings_never_make_a_minor_ninth` (all roots × types × melodies), `a_burst_of_plain_notes_loses_no_note_off`, `trill_release_before_the_second_pulse_balances_per_route` and `absurd_tempo_does_not_overflow`.

Gates: `cargo build` passes. `cargo test` passes in a debug build: 353 lib tests, 6 others, 0 failures. There are no new clippy warnings.

Nonblocking items 7–10 are still open. They are either for the wiring ticket or a doc fix.

Issue #74, opened 2026-09-24
Session: follow MIDI devices coming and going (Launchkey replug back to DAW mode)


ryangavin:
A `Session` connects its MIDI sources once, in `Session::start`. After that:
- `AppState.pads.connected` and `io.inputs` never change.
- A keyboard plugged in later is not heard.
- A Launchkey that is unplugged and plugged back in stays in its standalone mode. Its DAW port is not reconnected and it doesn't get `ENTER_DAW`, so the pads, buttons and LEDs stop working until the session restarts.

The fix needs CoreMIDI change notifications (`kMIDIMsgSetupChanged` / `kMIDIMsgObjectAdded` / `kMIDIMsgObjectRemoved`) or a device poll. Either way:
- CoreMIDI delivers notifications, and updates other processes' endpoints, through the CFRunLoop of the thread that created the client (see the comment in `bench::drive`). Today that is the caller of `Session::start` (the TUI's main thread, or a Tauri command thread), and nobody runs a run loop there. So the `Client` has to be created on a session-owned thread that runs a CFRunLoop, and connections re-made from there.
- On a change:
  1. Diff `midi::sources()`.
  2. Connect new keyboard sources with the same rules as `Session::start` (`--input`, `--all-inputs`, Launchkey keys).
  3. Connect a new Launchkey DAW source as `TAG_PADS`.
  4. Re-open the LED output port to its DAW destination, send `ENTER_DAW`, and clear the `Leds` caches so every LED is re-sent.
  5. Update `inputs` and `pads_connected`, and republish the state.
- CoreMIDI disconnects sources that vanish by itself. Their held notes need an all-notes-off on the keyboard parts, or they hang.

This needs testing with a real Launchkey replug, and with a virtual source from another process as in `yahaha drive`. It was left out of the AppState mirror PR for that reason.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

ryangavin:
Addressed in #101.

Finding: CoreMIDI binds its notifications, and its updates of other processes' endpoints, to the run loop of the thread that *first* uses CoreMIDI in the process. So a client created on its own thread isn't enough. Decision: `midi::init` starts one process-wide run-loop thread that makes the first CoreMIDI call, and every `midi` function calls it first. Sessions follow `midi::setup_generation`, with a 2 s poll as a fallback, and count only online endpoints.

On a Launchkey replug: the DAW port is reconnected, ENTER_DAW is sent and every LED is re-sent.

[...]

PR #77, opened 2026-09-24
AppState: full Launchkey mirror (buttons, Shift, Track neighbours, beat clock)

ryangavin:
This fills the AppState gaps for the Tauri app's 1:1 Launchkey mirror. There is a new `state.surface`, shaped after the provisional `SurfaceState` in #72 (`app/src/lib/api/types.ts`, `app/src/lib/surface.ts`), so `surfaceOf()` can return `state.surface` directly. The contract is in [docs/app-api.md](../blob/m3/launchkey-mirror/docs/app-api.md), under `surface`, `surface.clock` and `pads`.

## What's new
1. **Shift.** `surface.shift` is true while the Launchkey Shift button is held. It is mirrored from the MIDI thread, and every change republishes the state.
2. **Every non-pad control** is listed in `surface.controls`: Pad Bank ▲/▼, Track ◀/▶, Play, Stop, Scene/Function, the 8 fader buttons and the master button. Each one has:
   - `label` and `action` (an `AppCmd`)
   - `shiftLabel` and `shiftAction`
   - `rgb`, `level` and `anim`, plus the palette `colour` actually sent

   The data comes from the same code paths as the hardware:
   - `cc_control` for button meanings.
   - The `nav_button_msgs` / `fader_button_msgs` output for LEDs.

   A test checks this for every control, on every pad page and both fader pages, with and without Shift: pressing it on the hardware and sending its described `action`/`shiftAction` must end in the same state and output.
3. **Track neighbours.** `surface.trackPrev` and `surface.trackNext` are `{ id, name, path }` from `Library::step`, which skips files that fail to load.
4. **Beat clock.** `surface.clock` has these fields:
   - `atMs`, `bar`, `beat`, `phase`, `tempo`, `running`, `beatsPerBar`
   - `sectionAnchorMs` / `sectionAnchorBeats`: the engine snapshot now carries the position anchor, which only changes on a tempo, section or loop change.
   - `ledAnchorMs` / `ledAnchorBeats`: the pad flash clock, re-anchored only on tempo changes.

   The extrapolation formula is documented. Use `Session::state_now()` for the Tauri `state` command: it re-reads the clock at fetch time. Time passing alone does not bump `version`.
5. **Physical fader positions.** `surface.faders[i].position` is where each fader is, including the master. There is also a `fader` field on each keyboard part and Style part.
6. **`masterWaiting` fix.** `setMasterVolume` now sets `masterWaiting` from where the master fader physically is. Before, it stayed at its old value until the fader moved.
7. **Device hotplug** is not in this PR. I filed #74: it needs the CoreMIDI client on a thread that runs a CFRunLoop, and testing with a real replug. The limitation is documented.
8. **Palette-LED mode.** `pads.paletteLeds` is set, and each pad carries `palette` (mode, colour or colours, looks). The doc states the limitation: the Launchkey times palette flash and pulse itself, so they aren't in step with the tempo.

The unabridged fixtures from `yahaha state-json` are in `docs/fixtures/` (`state.json`, `library.json`). The example in the doc was regenerated.

## Differences from the app's provisional shape (#72)
- `trackPrev` / `trackNext` are `{ id, name, path }`, not strings. Use `.name` for the label; `id` and `path` are there for tooltips and for loading.
- Controls have two extra fields: `cc`, and `colour` (the palette index, or null).
  - Play, Stop, Scene and Function report `off` with `colour: null`, because yahaha doesn't drive those LEDs. #72 derived Play as green while running, but the hardware doesn't show that.
- Labels come from the engine in hardware style: `PAGE ▼`, `RIGHT 1`, `EDIT R1`, `PAD`, `MASTER`, `PANEL`. A Style fader button is labelled with the part name, and its state shows in `level` (dim when muted) rather than as a `… off` label.
- An unused fader still reports its physical `position`. Its `label` is `""`, and `value` and `set` are null.
- `clock` adds `running`, `beatsPerBar` and the two anchors. Times are ms (f64) on the session's monotonic clock. That is not `performance.now()`: extrapolate from `atMs` plus the local time since the state arrived.
- `shiftLabel` / `shiftAction` equal `label` / `action` where Shift changes nothing, as in #72. On Pad Bank ▲ on page 1, `label` is `""` but `shiftLabel` is `LEFT`.

## Gates
- `cargo build --release` passes, and so does `--no-default-features`.
- `cargo test --release`: 325 + 6 pass, with the corpus present. The new tests cover:
  - controls matching the hardware, on every page and layer
  - button descriptions and LEDs
  - Shift
  - Track neighbours, including skipping a bad file
  - clock extrapolation checked against the engine, including after a tempo change
  - fader positions
  - the master takeover fix
  - palette mode
  - the palette table
- Clippy: no new warnings. The same 11 remain, all on pre-existing lines.
- `yahaha screen` output is identical, and the TUI still quits cleanly with or without the synth.
- Bench, TickingAway, interleaved against `a5909c2`, 3 rounds each. Everything is within noise:

  | path | before p50 | after p50 |
  |---|---|---|
  | passthrough | 68.4–81.9 µs | 71.1–83.0 µs |
  | chord → engine | <6–7 µs | <7 µs |
  | wake vs deadline | <1 µs | <1 µs |
  | engine work per wake | <2 µs | <2 µs |

  The only engine-thread change is two extra fields in the snapshot: an anchor time and position.


ryangavin:
## Adversarial review of #77

Reviewed at 6039f64. I pushed fixes; the head is now **3cc6539**.
- The Rust gates pass at the new head: `cargo build --release` with and without `--no-default-features`, and `cargo test --release` (418 + 6 + 1, with the corpus). Clippy shows the same 11 pre-existing warnings.
- The app gates pass at the new head too: `app/src-tauri` `cargo test`/`clippy`, `svelte-check`, `vitest` (179) and `eslint`.

**Verdict: approve, with my commits.** Two blocking issues are fixed. The nonblocking issues left open are listed below.

### Blocking (fixed)

**B1. Merging #77 broke the app build.** #77 was cut from a5909c2, before #72.
- Against `integration/m3-ui`, `app/src-tauri` failed with 5 × E0063. `mock.rs` builds `KeyboardPart`, `StylePart`, `PadsState`, `AppState` and `Pad` literals without the new `fader`, `paletteLeds`, `surface` and `palette` fields.
- In live mode, `surfaceOf()` now returns `state.surface`. `trackPrev`/`trackNext` became `{ id, name, path }`, so the Track captions would have rendered `[object Object]`.
- Fixed in a7eeae8, after merging `integration/m3-ui` into the branch (c24db7d, 3cc6539):
  - The mock fields are filled in.
  - The Tauri `state` command now uses `state_now()`, as the doc says.
  - For the mock, the `state` command drops `surface`, so the UI keeps deriving it there.
  - `SurfaceState.trackPrev`/`trackNext` are now `Neighbour` objects in `types.ts`, and `deriveSurface` builds them.
  - The captions use `.name`.

**B2. A style change while playing, between styles with different ticks per quarter, played the band at the wrong speed, and `surface.clock` disagreed with the engine.**
- This bug predates the PR. `Engine::load` only re-timed `ns_per_tick` when stopped.
- The corpus mixes 480, 960 and 1920 ticks per quarter. For example, SlowWalker is 960 and TickingAway is 480.
- A throwaway test measured it at 75 BPM: 1 s after the switch, the engine had moved 2.5 beats and the clock formula 1.25.
- Fixed in 57b7772: a running load re-anchors at the same tempo with the new ticks per quarter. `style_change_keeps_tempo_across_resolutions` covers it.

### Nonblocking

**N1 (fixed, 57b7772). The hardware Track LEDs could stay dark while the mirror showed them lit.**
- `Leds` cached the nav LEDs by page only.
- A one-style session that gains a second style through `loadStylePath` never re-sent Track ◀/▶.
- The cache key is now `(page, styles)`. Test: `track_leds_follow_the_library`.

**N2 (fixed in part). The twin test can fail on actions, but not on LEDs or labels.**
- Mutating the Panel Shift action (`SelectPart` → `TogglePart`) fails both new tests. Good.
- These two mutations survived the whole suite:
  - dropping the Manual Bass mask from the `style_on` in `surface()`
  - changing Scene's label to `TEMPO -`
- The reason: the offline session has no `Leds`, so "same output" means synth output only. LED fidelity depended on a copy of the `style_on` logic.
- Now:
  - `launchkey::style_lit` is shared by `Leds::update` and `surface()`.
  - `launchkey_button_descriptions` checks the Bass button dims under Manual Bass, and the Scene/Function labels.
- Labels are otherwise unverified. There is no hardware truth for them.

**N3 (open). Track neighbours can be stale by up to 250 ms while the library indexes.**
- The neighbours come from `published`, which is throttled. Track ▶ steps through `lib`.
- A throwaway test marks the next file bad while other entries are pending. `trackNext` still named the bad file (id 3), and Track ▶ loaded id 2.
- This only happens while indexing. Using `lib` would republish the state on every 10 ms pump during indexing, so I left it. It could be documented.
- At the ends of the list the neighbours wrap, which matches the hardware and #72's `neighbours()`.
- With one style, both are null.
- If every other file is bad, both are null, but the Track controls still carry an action and a lit LED. That matches the hardware, which does nothing when pressed.

**N4 (open, doc). Two gaps in `docs/app-api.md`:**
- The clock formula doesn't clamp `pos` at 0. `ClockState::at` does, and `sectionAnchorBeats` can be negative: the fixture has −16.
- It could also say that a running style change restarts the section position at bar 1, mid-bar.

**N5. The beat clock is otherwise exact.**
- A throwaway test sampled `state_now()` every 37 ms and compared `clock.position(t)` with the engine's `tick_at`. It stayed within 1 ms on every sample, with bar/beat equal to `transport`, across:
  - start
  - TempoUp mid-bar (the anchors carry on correctly)
  - 4 × tap tempo
  - Main B with its fill
  - Break
  - Stop, then restart
  - 40 s of Main A loops
  - a style change between styles with the same ticks per quarter
- `led_beats(t)` matched `Session::beats()` within 1e-6, including after tempo changes from the Launchkey Scene button.

**N6. Real-time safety is OK.**
- MIDI thread: two atomics, plus a `ctl_signal` when Shift or the master fader changes.
- Engine thread: two more fields in the `Copy` snapshot.
- `surface()` runs in `build_state` on the control thread, at most every 10 ms. It makes about 20 small allocations, including `button_colours`' `Vec`.
- `state_now()` deep-clones the `AppState` on every Tauri fetch. That is fine at the fetch rate.
- No locks or panics were added on the real-time threads.

**N7. API naming is consistent.** Serde camelCase matches the doc and `docs/fixtures/*.json`.

### Frontend checklist: from `deriveSurface` to `state.surface`
Only the Track caption type change (B1) was done here.
1. **`types.ts`**
   - `SurfaceControl`: add `cc: number` and `colour: number | null`.
   - `SurfaceState.clock`: use the full `ClockState` shape: `running`, `beatsPerBar`, `sectionAnchorMs`/`sectionAnchorBeats`, `ledAnchorMs`/`ledAnchorBeats`.
   - Make `AppState.surface` required once both mocks send it.
   - Add `fader: number | null` to `KeyboardPart` and `StylePart`, and `paletteLeds` to `PadsState`.
   - Add `Pad.palette: PaletteLed | null`, and the `PaletteLed` type.
2. **`store.svelte.ts` `BeatClock`**
   - Use the doc formula: record `receivedMs` on each state, then derive `pos` from the section anchor.
   - Lamps: run the lamps' `beats` from the **LED anchor**. The hardware pads flash on the free-running LED clock, not on the section position.
   - Remove the 0.999 cap.
   - Stop mixing `transport.bar`/`beat` (as of build time) with `clock.phase` (as of fetch time).
3. **`surface.ts`**
   - Change `surfaceOf` to `s.surface`.
   - If `deriveSurface` stays for the mocks, make it match the engine:
     - labels: `PAGE ▲/▼`, `◀ STYLE`/`STYLE ▶`, `PLAY`, `STOP`, `TEMPO +/-`, `RIGHT 1`…, `EDIT R1`…, `PANEL`/`STYLE`
     - Style fader buttons: upper-case names with no ` off` suffix; state goes in `level`
     - Pad Bank ▲ `label: ''` on page 1
     - Play/Stop/Scene/Function reported `off`
     - include `cc` and `colour`
4. **Pads:** when `pads.paletteLeds` is set, render `pad.palette`.
5. **Launchkey mirror:**
   - Show the Shift layer from `surface.shift` or `ui.shiftLatched`.
   - Show fader position markers from `faders[i].position`.
   - The Parts and Mixer drawers can use `fader`.
6. **Mocks:** have `mock.ts` and `app/src-tauri/src/mock.rs` produce a full `surface`, then remove the `surface` stripping from the `state` command in `lib.rs`.
7. **Shape test:** re-record `app/src/lib/api/engine-shape.json` from `yahaha state-json`, so `engine-shape.test.ts` covers the `surface`, `fader` and `palette` paths.

Issue #37, opened 2026-09-23
Multi Pads: parse .pad, playback, Repeat, Chord Match, Synchro Start


ryangavin:
_Details in the spec._

---
Milestone: **M7 · Registration, Multi Pads, Playlist** · Spec: [docs/genos-features.md](../blob/main/docs/genos-features.md) · Backlog: [docs/backlog.md](../blob/main/docs/backlog.md)

ryangavin:
Done in #95, merged to main via #113.

PR #78, opened 2026-09-24
Multi Pads core: .pad parser and player (#37, part 1)

ryangavin:
Part 1 of #37. This adds pure Multi Pad modules. Nothing is wired into engine, session, api or ui yet; a later ticket does that.

## Format research
- There are **no .pad files** in the corpus or anywhere under the user's home directory, including ~/Library. The manuals say the format is undocumented (genos-features §E, §G.6).
- The parser is therefore **provisional**. It uses the style-file analogy (an SMF followed by Yamaha chunks) and what the manual does say: each pad is recorded on one MIDI channel, and each pad has a name, Repeat and Chord Match (RM p.64–65).
  - **Type 0**: one pad per channel. When a CASM chunk is present, its Ctab/Ctb2 records set the pad order, the names and the chord rules. Without CASM, pads are assigned lowest channel first.
  - **Type 1**: one pad per track. The track name is the pad name.
  - **Chord Match** is inferred from the CASM rule. It is off only when every zone is Bypass/Root Fixed. With no CASM rule it is unknown (`None`).
  - **Repeat**: we don't know where it is stored, so it is always `None`. Both flags have panel overrides on the player.
  - Unknown chunks are kept raw. `yahaha pad <file>` hex-dumps them, so the first real bank will show what the parser got wrong.
- All tests use synthetic SMF/CASM data built in code. No Yamaha data is committed.

## Player (`multipad::player`)
- `MultiPadPlayer::new(&bank, host_ppq)` does all the allocation. `trigger(pad, start_tick)`, `stop(pad, sink)`, `stop_all`, `stop_repeating` (for Synchro Stop), and `process(tick_range, Option<Chord>, sink)`. `process` does not allocate or panic, and runs events in time order across all pads.
- Repeat loops on the pass length, rounded up to whole beats. A late `process` call skips passes that have already finished. Pressing a playing pad restarts it.
- Chord Match uses `theory::transpose_group` with the pad's CASM rule. When the pad has no rule it uses `default_rule`: CM7 source, Melody, Root Trans. With no chord, or with Chord Cancel, the pad plays as written. A held note keeps its pitch through a chord change.
- Synchro Start is modelled as data: `arm` (toggle), `sync_fires(trigger, acmp)`, `fire_sync(start)`, and the `PadState` lamps.
- `start_tick(now, Clock)` answers "start at the next measure while the style is playing".
- Pads output on MIDI ch 5–8 by default (`set_out_channel` changes this).

## Shared-file edits
- `src/lib.rs`: `pub mod multipad;`
- `src/main.rs`: `pad` subcommand.
- `src/sff.rs`: `parse_casm` is now `pub(crate)`. This is a visibility change only.

## Gates
- `cargo test`: 340 + 6 tests pass, 22 of them new.
- No new clippy warnings.


ryangavin:
## Adversarial review of #78

I reviewed de9feb1. My fixes are pushed as a2f43e7 (plus the merge commit d9944a8, which brings in integration/m3-ui). **Verdict: approve once a2f43e7 is in.** I have not merged anything.

### 1. Format research: BLOCKING (fixed in a2f43e7)
The .pad format does have a public description: Jørgen Sørensen's [Multi Pad Format](http://www.jososoft.dk/yamaha/articles/multipad.htm) article on The Unofficial YAMAHA Keyboard Resource Site. It covers the Tyros layout and notes that other models differ slightly:
- The file is a **type 1 SMF with 5 tracks**. Tyros uses 1920 ppq; older PSRs such as the PSR-740 use 96.
- **Track 0** holds 10 text events (meta 01H), all at tick 0:
  - `CMxxxx` is **Chord Match** for pads 1–4.
  - `RPxxxx` is **Repeat** for pads 1–4. **This is where Repeat is stored.**
  - `N<y><name>…` holds the pad names (52 bytes, blank-padded).
  - `I<y>…` holds the image references.
- **Tracks 1–4** are the pads, one measure each. Pad n plays on channel n.
- The article never mentions CASM, so the format is **not** CASM-based like styles. Chord Match is a single on/off flag per pad.

The PR's version fell short in three ways:
- It never set Repeat.
- On a real bank it guessed Chord Match (or left it unknown) and named every pad "Pad n". I confirmed this with a Tyros-layout synthetic bank.
- In type 1 it skipped a track with no channel events, so **an empty pad shifted every later pad down a slot**.

What a2f43e7 changes:
- Parses CM/RP/N/I. These flags take priority over the CASM inference, which is kept as a fallback, along with type 0.
- Track n is now pad n; empty tracks no longer shift the others.
- Adds `Pad::image` and `PadBank::texts`, and `yahaha pad` prints the texts.
- The module docs now cite the source and say plainly what is still unverified. Genos files may differ from Tyros.
- Adds 4 tests. Text events that aren't UTF-8 or are truncated can't panic, because parsing works on chars rather than byte offsets.

Other sources I checked: psrtutorial.com's multipad pages ("a multipad is a four track midi file"). Bedesem/Wierzba's *Style Files* doc has nothing on .pad. The PSR Tutorial forum and Tyrosmania are behind bot protection and I didn't get past it.

### 2. Player safety
- **BLOCKING (fixed):** changing the channel with `set_out_channel` while a pad is sounding left **stuck notes**, because note-offs went to the *new* channel. I reproduced this with a random-op test. Each sounding note now remembers the channel it started on. Regression test: `rerouting_a_sounding_pad_ends_its_notes_on_the_old_channel`.
- A throwaway fuzz test ran 200k mutated Tyros-layout banks and 200k random buffers through `parse`, then `MultiPadPlayer::new` with a random ppq, then `process` over a random range of up to 1M ticks with all pads repeating. No panics.
- A throwaway random-op test ran 2000 rounds × 200 ops of trigger/stop/stop_repeating/set_repeat/set_chord_match/set_out_channel/process, with random chords including out-of-range types, then `stop_all`. Every on/off was balanced per (channel, key) after the fix.
- `process()` made **0 allocations** under a counting global allocator. It needs `--test-threads=1`, because the counter is global.
- Chord Match reuses the transposer correctly: `transpose_group` with the pad's rule, `dest_ch` forced to the output channel, and out-of-range chord types or Cancel play the notes as written. Nonblocking: if a pad is routed to channel 10 (index 9), the transposer treats it as drums. That is probably correct, but worth knowing.
- Nonblocking: if two pads are routed to the same channel and play the same key, one pad's off ends the other's note. The default channels 5–8 can't collide.

### 3. Engine fit: nonblocking
- Channels 5–8 match the Genos Song convention (RM p.76, genos-features §1026), with keyboard parts on 1–4 and Style on 9–16. Real files record pads on channels 1–4, so the rechannelling in `process` is **essential**. Without it, pads would play on the Right1/Left/Right2/Right3 parts.
- Harmony (#76) plays on the Right parts and doesn't conflict. For plugin hosting (#79), the synth or router needs parts on channels 5–8 with their own PartGain. That should go on the wiring ticket.
- Tick domain: the engine runs in *style* ppq (`style.ppq`, f64 ticks), and that changes when the style changes. `MultiPadPlayer` bakes `host_ppq` in when it is built. A pad playing across a style change to a different ppq would run at the wrong speed. Recommendation: have the engine drive pads from a fixed-ppq clock derived from ns and bpm (e.g. 1920, which is Tyros's native pad resolution), or add a rescale method.
- Nonblocking: the PR conflicted with integration/m3-ui in the `main.rs` usage line (the plugin-test arm from #79). d9944a8 resolves it and keeps both arms, and the build is clean with and without `--features plugins`.

### Gates on a2f43e7
- `cargo test`: 345 + 6 pass.
- clippy: no new warnings. The remaining warnings are pre-existing, in bench/engine/live/sff/theory/ui/main.
- `cargo build --features plugins`: passes.

### Getting real test files
Export a bank from the instrument's Multi Pad Creator (Save to USB). Or download free banks: psrtutorial.com's Multi Pad library, and Yamaha's own Genos/Tyros preset pad packs on the psrtutorial "Yamaha Keyboards Preset Multi Pad Files" page. Then run `yahaha pad <file>` and compare its texts/chunks output with the article's layout.

ryangavin:
I merged integration/m3-ui again (b076664) to pick up #75, #76, #80 and #81. The only conflict was the `main.rs` subcommand arms and usage line; I kept `pad`, `ireal` and `plugin-test`. On b076664, `cargo test` passes 436 + 6 + 1 and clippy shows no new warnings. I approve with head b076664.

Issue #73, opened 2026-09-24
iReal Pro core: parse irealb:// playlists, expand the form, map chords


ryangavin:
A pure library module with no engine or UI wiring. It should:
- decode irealb:// and irealbook:// links and HTML playlists: title, composer, style, key, tempo, time signature, and the unscrambled chart
- tokenise the chart: bars, sections, time signatures, repeats, endings, D.C./D.S./Coda/Fine, N.C., repeat-bar shorthands, alternate chords and slash chords
- expand the form into a linear list of bars, each with its chords placed on beats
- map iReal chord qualities to yahaha chord types, in a documented table, falling back to the nearest type where there's no exact match

Tests use synthetic charts only. Don't commit real song data.

PR #80, opened 2026-09-24
iReal Pro core: playlist parser, form expansion, chord mapping (#73)

ryangavin:
Closes #73.

This adds `src/ireal/`, a pure library module. It has no engine, session, API or UI wiring. The only edits to shared files are one `pub mod ireal;` line in `src/lib.rs` and a `yahaha ireal` debug arm in `src/main.rs`, both claimed on the board.

## API

- `parse(text) -> Result<Vec<Playlist>>` reads a bare link, or an HTML or text file that contains links.
  - It also exposes `parse_url` and `percent_decode`.
  - `Playlist { name, songs }`.
  - `Song { title, composer, style, key, transpose, groove, tempo, repeats, chart }`.
  - `Song::chart()`, `Song::bars(choruses)` and `Song::key_root()`.
- `unscramble` / `scramble` handle the `1r34LbKcu7` chart. It is scrambled in 50-character blocks and uses the `Kcl`/`LZ`/`XyQ` abbreviations. The source is cited in `scramble.rs`, and no code was copied.
- `tokenize(chart) -> Vec<Token>` covers every token type. It never fails.
- `parse_chart(chart) -> Chart { bars: Vec<ChartBar> }` gives the written bars.
  - Each bar has its marks (repeats, endings, segno/coda/fermata, Fine, D.C./D.S. jump, `3x` count, `x`/`r`).
  - Chords are placed on beats as `BeatChord { beat, chord: theory::Chord, alt }`. N.C. is `CANCEL`.
- `expand(&chart, choruses) -> Vec<Bar>` plays the form out.
  - Each `Bar` has `{ time, section, section_start, chords, source, chorus }` and implements `Display`.
  - Output is capped at `MAX_BARS` and every jump spends from a step budget.
- `QUALITIES`, `map_quality`, `to_chord` and `fallback_type` map iReal qualities to yahaha chord types. The table is in `docs/ireal.md`.
- `yahaha ireal <file-or-link> [--choruses N]` prints the expanded bars.

## Rules worth reviewing (all in `docs/ireal.md`)

- **Beats:** each chord, N.C., slash and space is one cell.
  - In a 4-cell bar with at most 4 beats, the chord in cell `i` goes on beat `i`. Otherwise it goes on `floor(i*B/n)`.
  - Collisions move the chord to the next beat.
- **Endings:** on pass `p` of `t`, the passes before the last play ending `min(p, last-1)`. The last pass plays the last ending, and an `N1` with no `N2` is skipped on it.
- **D.C./D.S.:** the jump is taken once per chorus. After it, repeats play once and take the last ending.
- **Choruses:** the coda, Fine and `U` end the song, so earlier choruses restart the form at those points. Without a D.C. or D.S., the "to coda" sign is taken on the last chorus.
- **irealb fields:** songs are found by their music field, not by splitting on `===`. A song with an empty composer contains `===` inside it.

## Chord mapping decisions

- **Dominants with several alterations** keep the first altered tension: `7alt -> 7#9`, `7#9#5 -> 7#9`, `7b9#11 -> 7b9`. **Altered 13ths** keep the 13.
- **Sus dominants** (`9sus`, `13sus`, `7b9sus`, `7b13sus`, `11`) all map to `7sus4`.
- **Re-rooted:** `-#5` and `-b6` are re-rooted to exact chords rather than approximated: `C-#5 = Ab/C`, `C-b6 = Abmaj7/C`.
- **Other approximations:** `^13 -> maj9`, `-69 -> m6`, `h9 -> m7b5` and `2 -> sus2`.

## Tests

`src/ireal/tests.rs` has 20 tests, all on synthetic charts. No real songs or playlists are committed. They cover:

- the scramble round trip, including block boundaries;
- every token type;
- cell layout;
- repeats, endings and N3;
- D.C./D.S. al Coda, al Fine and al 2nd ending;
- `x`/`r`, choruses and `U`;
- every quality in the mapping table;
- 3000 rounds of fuzzing with no panics and bounded output.

`cargo build --release` and `cargo test --release` pass. There are no new clippy warnings; the two in `main.rs` are older.


ryangavin:
## Adversarial review of #80

I checked out `cb711ae`, then built and ran the tests: 338 passed. I tested each area below with throwaway tests, which I did not commit. There was **one blocking item, and I fixed it in `3a8afa6`**. Everything else is nonblocking.

### Blocking (fixed in 3a8afa6)

**Chord mapping: `-#5` and `-b6` re-rooted the chord.** `C-#5` came out as `Ab/C` and `C-b6` as `Abmaj7/C`. The pitch sets do match, but that isn't a strong enough reason:
- the player sees an Ab chord where the chart says C minor;
- the style follows an Ab root, which also affects the chord display and anything keyed off the root;
- in a line cliché (`C- C-#5 C-6 C-7`) the root would jump to Ab and back.

**Fix:** both now map to `m`, marked Approx. The written root is kept, and so is any written bass (`A-#5/G` gives `Am/G`).
- I removed the `Quality::shift` field and the shift logic in `to_chord`, since nothing uses them now.
- I updated the docs table and the rule text.
- I updated the tests and added a cliché test.
- Gates: build passes, all tests pass, and there are no clippy warnings in `src/ireal` or `ireal_cmd`. The main.rs warnings at lines 22 and 278 were already there.

### Nonblocking

1. **Unscramble fidelity: OK.** It matches the algorithm described for pianosnake/ireal-reader and ireal-musicxml:
   - 50-character blocks, scrambled while more than 51 characters remain;
   - positions 0–4 swap with 49–45, and 10–23 with 39–26;
   - then `Kcl` becomes `| x`, `LZ` becomes ` |` and `XyQ` becomes three spaces.
2. **Playlist fields: OK for the cases tested.** Multiple songs, an empty composer, and `Title===Style` inside a song all parse correctly.
   - One gap: an older irealb song with no groove, tempo or repeats after its music field, followed by `===Name`, loses the playlist name. For example, `...=<music>===Named` gives `name=None`, and `repeats` falls back to 3 after failing to parse "Named". This is rare.
3. **Bare link with "ireal" in the text:** `irealbook://My ireal tune=...` goes down the HTML path and is cut at the first space, so the title becomes "My". The `!t[1..].contains("ireal")` check could look for a second `irealb://` / `irealbook://` instead.
4. **N2 with no N1:** `{C |N2D }E Z` plays `C E`. It skips both the ending and the repeat. The chart is malformed, but reading it as a plain repeat would be kinder.
5. **`x` in a second ending** repeats the *written* bar before it, which is the last bar of N1, not the bar played before it. I think this is right for iReal's written layout, but it's worth one line in the docs.
6. **Comment heuristics:**
   - Any comment that contains `D.S.` or `D.C.` becomes a jump, even `<| G7 x D.S. |>`. A `|` or chord text inside `<...>` is correctly kept inside the comment.
   - `<C7 3 x>` reads as a repeat count of 3.
   - Both are acceptable heuristics.
7. **Mapping judgement calls. All defensible, none blocking:**
   - `7alt → 7#9`: fine.
   - The sus family and `11 → 7sus4`: fine.
   - `-11 → m11`: fine.
   - `^13 → maj9`: fine.
   - `7susadd3 → 7sus4` drops the 3rd the name asks for. `7` might be the closer match.
   - `2 → sus2`: many lead sheets read C2 as add2, with the 3rd. I'm not certain how iReal voices it.
8. **Form expansion matches how I understand iReal plays:**

   | Chart | Result |
   |---|---|
   | N1/N2/N3 | `C D E F C D G A C D B B7` |
   | `3x` with N1/N2 | N1, N1, N2 |
   | D.S. al Fine | `C D E F E F` (stops after Fine) |
   | D.C. al Fine, 2 choruses | the full form each time |
   | D.S. al Coda, 2 choruses | the coda only on the last chorus |
   | no jump, `Q` inside a repeat | coda taken on the last pass of the last chorus |
   | D.S. al 2nd End. inside a repeat | takes N2 |
   | D.C. al Coda | repeats play once after the jump |
   | `x`/`r` across a section boundary | correct |
   | `W/E`, `p`, `(alt)`, `s`/`l` | correct |
   | T34/T54/T68/T12/T22, mid-chart time changes | correct beats |

   A lone `Q` with no coda target is ignored, which is fine.
9. **Robustness: OK.**
   - Token-level fuzz: 30k random charts built from navigation-heavy tokens, each expanded at 1, 3 and 1000 choruses. Every result was at most `MAX_BARS`, and every expansion took under 500 ms.
   - `"{C7 |D7 <99x>}"×2000` at 1000 choruses: capped at 10,000 bars in 2.5 ms.
   - 80k characters of unterminated `(`, `<`, `T` and `*`: 0.25 ms.
   - Malformed and unicode `irealb://` input: no panics.
10. **Content: clean.** Every title, composer and chart in the tests and docs is synthetic ("Test One", "Doe John", ...). There are no real songs.

**Verdict:** approve once `3a8afa6` is in. Head is now `3a8afa6`.

Issue #30, opened 2026-09-23
Part solo and Style Track Mute; tempo range 5–500; metronome


ryangavin:
_Details in the spec._

---
Milestone: **M5 · Sections & transport parity** · Spec: [docs/genos-features.md](../blob/main/docs/genos-features.md) · Backlog: [docs/backlog.md](../blob/main/docs/backlog.md)

ryangavin:
Implemented in #96. Decisions: two independent solo groups (Style tab, Panel tab), one part each; solo overrides the part's Off switch and leaves the switches alone (no gain change: notes stop, CC7 untouched). Style Track Mute: the eight steps split the 0-127 knob evenly and set the parts' on/off switches (RM p.148 orders A/B). Tempo 5-500 (+ setTempo). Metronome: the style's time signature and quarter-note beats, bell on by default, free-running while stopped; the click is a synthesized voice of the built-in synth, never on the MIDI port, with its own volume. Details: docs/solo-metronome.md.

ryangavin:
Done in #96, merged to main via #113.

PR #81, opened 2026-09-24
Mixer drawer: Panel and Style tabs like the Genos Mixer

ryangavin:
Builds the Mixer drawer (`app/src/panels/mixer/`), laid out like the Genos Mixer's Panel and Style tabs.

## What it shows
- **Tabs = the Launchkey fader page.** Clicking Panel/Style (or ←/→ on the tabs) sends `setFaderPage`, so the hardware follows. The Launchkey's page button (or F9) switches the tab too. The page isn't copied locally. A lamp shows the master button's LED colour.
- **Panel tab:** Right 1, Right 2, Right 3 and Left on faders 1–4, in Launchkey order. Faders 5–8 are drawn as unused, like on the hardware.
- **Style tab:** Rhythm 1, Rhythm 2, Bass, Chord 1, Chord 2, Pad, Phrase 1 and Phrase 2 (channels 9–16).
- **Each strip has:**
  - the MIDI out channel number, for mapping in Ableton, with the output port shown in the header;
  - a big fader showing the real CC 7 value (0–127);
  - the ↕ soft-takeover mark, plus a dashed ghost cap where the physical fader sits (from `state.surface.faders[i].position`);
  - an On/Off button lit from the fader button's own LED;
  - Solo, **shown disabled** because the engine has no solo yet (#30);
  - the voice. For style parts it shows what the synth plays (≈ name) and the Yamaha bank/program or GM voice the style was written for. The Bass part shows a "Manual Bass" badge when the engine mutes it, and Left shows "Plays bass".
- **Master fader:** "100 = unity · soft clip above −1 dBFS", which matches `synth.rs` `CLIP_KNEE`.
- **The mixer principle:** an info line and the tooltips explain it. A fader IS the channel's CC 7, with no hidden gain, and loading a style sets the Style faders to the style's own levels.
- **No level meters.** The engine exposes none, so none are faked.

Every control has a catalog tooltip. New keys: `mixer.solo`, `mixer.channel`, `mixer.voice` and `mixer.info`. The `mixer.panel.*`, `mixer.style.volume` and `mixer.master` bodies were rewritten to state the principle. `docs/controls.md` is regenerated. A coverage STATE was added for the Style tab.

Outside the panel folder: the Mixer section of `tooltips.ts`, one `coverage.test.ts` STATES line, and one `screenshots.sh` line. All were claimed on the board. The drawer is widened to 48rem with a scoped `:global([data-overlay='mixer'])` rule, and the shared Overlay is unchanged.

## Engine API NEEDs (posted on the board)
- `state.surface.faders[i].position` from the real engine. Only the mock sends it today, so on the real engine no ghost cap shows. Everything else degrades cleanly.
- Per-part solo (#30): a command plus a flag on StylePart/KeyboardPart.
- Per-channel output level meters (1–4, 9–16, master after the soft clipper).

## Screenshots
Style tab, 1440:
style
Panel tab, 1440:
panel
1024 / 1920:

Light theme:
light style
light panel
Tooltips:
solo
info

## Verify
- `npm run verify` passes: svelte-check shows 0 errors and 0 warnings, eslint is clean, and all 162 vitest tests pass, including the new `panels/mixer/Mixer.test.ts` and tooltip coverage.
- Checked in the browser on the mock at 1024, 1440 and 1920, in dark and light themes.


ryangavin:
## Adversarial review: Mixer drawer (head 1bd1543, then fixes in 40834c2)

What I ran: `npm run verify` (0 errors, 0 warnings, 162/162 tests) and `cargo build --release`, both passing before and after my fix. I checked the drawer on the mock at 1024×900, 1440×900 and 1920×1080, in dark and light themes. I also removed `mixer.solo` from the catalog and confirmed that `coverage.test.ts` fails (2 tests), then restored it.

### Blocking (fixed in 40834c2 and pushed)
1. **The drawer scrolls into empty space at every width.** Each `Fader` has a cap carrier: a full-height box that moves by `transform`. At low values that box hangs up to a whole track below the fader. The four unused Panel faders (value 0) and any low style fader do this. As a result the drawer's `.body` had a scrollbar and scrolled about 240px (1024×900) to about 420px (1920×1080) into blank space. The PR screenshots don't show it because `screenshots.sh` passes `--hide-scrollbars`. Fix: `.strips :global(.track) { overflow: clip }` in `Mixer.svelte`. The cap and the ghost cap always sit inside the track, so nothing visible is clipped. After the fix, scrollHeight equals clientHeight (771/771 at 1024×900). The scrollbar is gone, and "Brass Section" now fits without an ellipsis at 1024.
2. **Keyboard focus on the tabs was invisible.** `.tab.pressed { outline: 1px }` out-specifies the global `:focus-visible` ring. The tabs use a roving tabindex, so only the selected tab can take focus, and a focused tab looked the same as an unfocused one. Fix: a `.tab:focus-visible` rule with a 2px accent outline and a 2px offset.

### Mixer-specific checks
- **Tab ↔ hardware fader page sync: OK in both directions.** The page is `$derived(app.state.mixer.faderPage)`, with no local copy. A click or ←/→ sends `setFaderPage`, and only when the page differs, which matches `session.rs`. In the browser, a click on Style switched the mirror to "Faders · Style: the band", and F9 (`toggleFaderPage`) switched the tab back to Panel. The `masterButton` lamp follows too. The tests cover both directions.
- **Mixer principle text: accurate.** I checked it against the engine. `Engine::set_volume` sends the part's CC 7 unchanged. `set_mixer` and `restore_untouched_levels` set the faders from the SInt CC 7 on a style load, and `pattern_volume` moves only faders the player hasn't moved. `synth::apply` passes CC 7, CC 11 and velocity straight through. `master_gain` is linear with `MASTER_UNITY = 100`. `CLIP_KNEE = 0.8912509` is −1 dBFS, and the clipper is transparent below it. The master applies only to the synth, which matches the `mixer.master` text. `voice.ts` parses `api.rs::voice_label` and computes nothing itself.
- **48rem at 1024: OK.** The drawer is 768px wide (`min(48rem, 100vw − 32px)`), with no horizontal overflow on the page, the drawer or `.strips`. It covers the right three-quarters of the Launchkey mirror at 1024, which is expected for a drawer.

### Nonblocking
- `mixer.info`: "no hidden gain anywhere, so the MIDI output and the synth hear the same level" is true per channel. The master fader is the exception, because it scales the synth but not the MIDI out. Consider "…same channel levels". `mixer.master` already states the exception.
- The root cause of blocker 1 is in the shared `lib/ui/Fader.svelte`. `.track { overflow: clip }` there would protect every future scrolling container. It needs a board CLAIM, so it's worth a follow-up.
- Voice names that don't fit end in an ellipsis. The tooltip is the generic `mixer.voice`, so there's no way to read a truncated name. A `title`, or the full name in `aria-label`, would help at 1024.
- On/Off is an `HwButton` whose `aria-label` is "X on"/"X off". An `aria-pressed` toggle would read better, but that belongs to the shared component.
- The info line, the "Launchkey faders" lamp and the MIDI-out label carry tooltips but can't take focus, so they're mouse-only. That matches other panels.
- Solo is correctly `aria-disabled` and is not faked. There are no fake meters. The NEEDs for fader positions, solo and meters are on the board.
- **Performance:** there are no filters or blur in the panel. The caps and ghost caps move by transform with `will-change`, and the only static glow is a `box-shadow` on the lamps.
- **Conflicts:** `git merge-tree` against #82 (m3/ui-parts) is clean. `controls.md` may need regenerating after whichever of the two merges second.

**Verdict:** approve with the two fixes above.

PR #82, opened 2026-09-24
Keyboard parts + OTS drawer

ryangavin:
This replaces the Parts drawer slot (`app/src/panels/parts/`) with the Keyboard parts and One Touch Settings drawer.

## What it shows

- **Your hands.** A keyboard map shows who plays on each side of the split:
  - the Left voice;
  - Left playing the style's Bass under Manual Bass;
  - chords only, when Left is off in Lower;
  - the Right parts.

  The map also marks where the chords are read. Split −/+ and Manual Bass sit under the map. Manual Bass says "needs Upper" when it's on but not active.
- **Right 1–3 and Left as channel strips.** Each strip has:
  - **Edit**: the part that Voice −/+ changes, lit and outlined;
  - **on/off**, lit from `sounding`;
  - **a voice screen** with a transparent native picker over it, grouped by GM family. The picker blurs after you choose, so the performance keys keep working;
  - **octave −/+**;
  - **the volume fader** (CC 7), which is the shared `Fader` with the soft-takeover ↕ and a ghost cap at the hardware fader's position.
- **Layering.** A bar over Right 1–3 joins the Right parts that sound together and reads "Layer: R1 + R2", "Right 1 alone", and so on.
- **Manual Bass.** Left's screen shows the Bass voice with "own: Strings" under it, and its switch reads "Bass". In As written for, the style's Bass row says it's "your left hand".
- **OTS 1–4.** Each card shows what the OTS sets (voice, on/off, volume, octave for each part). The last recalled card is lit and tagged. Cards the style doesn't have say "Not in this style".
  - A plain one-line explainer sits above the cards: the setups are for *your hands*, and the band doesn't change.
  - OTS Link has a timing readout: **Real Time**, meaning it recalls as soon as you press a Main. That matches the engine, which recalls when `transport.main` changes.
  - With OTS Link on, each card is tagged with its Main, and the card for the Main that's playing is outlined.
- **Recalling an OTS** flashes the strips once (opacity only). The parts and faders update, and every fader the recall moved shows ↕ in the drawer and on the mirror. A line under the strips names which Launchkey faders are waiting.
- **As written for.** Channels 9–16 with each style part's written-for voice label, so you know what to load in Ableton.
- **Mirror link.** Hovering or focusing a strip highlights its fader and fader button on the Launchkey mirror's Panel page. If the faders are on the Style page, it highlights the fader page button (under master) instead.

## Tooltips

Every control has a catalog entry. The coverage test has two new states: parts drawer with Upper + Manual Bass + OTS Link, and parts drawer with the Style fader page.

- **Rewritten:** `ots.1`–`4` and `ots.link`, in plain words. OTS is "a sound setup for your own hands that the style's author picked…; the band doesn't change".
- **New:** `part.voice`, `part.layer`, `part.left_zone`, `part.written_for` and `ots.link_timing`.
- **Reused:** `part.*.on/select`, `part.octave_*`, `mixer.panel.*`, `mixer.pickup`, `split.*` and `detection.manual_bass`.
- `docs/controls.md` is regenerated.

## Outside the panel folder (claimed on the board)

- `lib/mirror.svelte.ts` (new): the hover link.
- `lib/api/voices.ts` (new): the voice list. It reads a proposed `state.voices` and falls back to the engine's GM names (the same list as `gm_name()` in `src/api.rs`).
- `panels/launchkey/FaderBank.svelte`: one `class:linked` on the strip, plus its CSS.
- `help/tooltips.ts`, `help/coverage.test.ts` and `docs/controls.md`, as listed above.

I didn't update the "SLOT" line for the parts drawer in CONTRIBUTING.md, to avoid adjacent-line conflicts with the other panel PRs.

## NEEDs from the engine API

- **`state.voices`** (or `LibraryList.voices`): `{ program, name }[]`, the list `setPartVoice` picks from. Today the UI falls back to GM.
- **OTS Link timing** (Genos Real Time / Next Bar), if we want a Next Bar option. The engine only does Real Time today, so the drawer shows that as a read-only value.
- **`surface.faders[i].position`** from the real engine, for the ghost cap (the same NEED as the Mixer's).

## Findings

The mock differs from the engine in two places:

- It recalls OTS under Link when the new Main *starts* at the bar. The engine recalls on the press.
- Turning OTS Link on doesn't recall right away in the mock. The engine does (`!last_link`).

## Verify

- `npm run verify` passes: svelte-check, eslint, and 163 vitest tests including tooltip coverage and 8 new drawer tests.
- Checked on the mock at 1024, 1440 and 1920, in dark and light.

## Screenshots

**OTS 4 just recalled:** faders waiting (↕ and ghost caps), with the Left strip hovered, so fader 4 is lit on the mirror.
recall

**Hovering Right 2** lights fader 2 on the mirror.
drawer

**The whole drawer: Upper + Manual Bass, OTS Link on.**
manual bass

**Light theme.**
light

**1024 wide.**

**1920 wide.**


ryangavin:
## Adversarial review: Keyboard parts + OTS drawer

I reviewed b21f6d1 and pushed fixes for the two blocking findings in **afe3351**.

**Gates on afe3351:**
- `npm run verify` passes: svelte-check, eslint, and 165 vitest tests.
- Coverage test: removing `ots.link_timing` from the catalog fails all 3 parts-drawer STATES, as it should.

**Browser (mock):** I checked 1024, 1440 and 1920 in dark and light. There is no truncation or overflow in the drawer, and no layout shift on recall. With the drawer open and an OTS recall running I measured 60 fps with 0 long tasks, and no blur or backdrop-filter anywhere in the drawer. The recall flash animates opacity only.

### Blocking (fixed in afe3351)

1. **The voice picker stole the performance keys, and type-ahead turned them into voice changes.** `pick()` blurs only on `change`. So after a mouse click that dismisses the list, or that picks the voice already selected, the `<select>` keeps focus. `shortcuts.ts` skips SELECT targets, so every shortcut went dead. Worse, the select's native type-ahead acted on the keys:
   - In the browser, with Right 1's picker focused, `7` didn't toggle Right 3.
   - `s` changed Right 1 to Steel Gtr and sent `setPartVoice`.
   - `b` (Style part 5) would pick Bagpipe, `t` (Tap Tempo) Taiko, and so on.

   **Fix:** the picker now keeps only the list keys: ↑/↓, Home/End, Tab, modifiers, and anything with Ctrl, Alt or Cmd. Any other key blurs it, and the event is re-dispatched to the window shortcut handler. I checked this in the browser and added a regression test, which fails without the fix.
2. **The Manual Bass toggle was lit in Lower.** It was driven by `chord.manualBass`, the stored setting, which defaults to on. So in Lower it showed lit with "set · needs Upper", while:
   - the engine ignores `toggleManualBass`/`setManualBass` in Lower (`session.rs` ~430), so clicking did nothing;
   - the Launchkey MAN BASS pad is dark there (`launchkey.rs` 549, test at 742);
   - its own tooltip says "Dark in Lower".

   **Fix:** it now lights from `manualBassActive`, and the note reads "Upper only" in Lower, "off" in Upper with the setting off, and "Left plays the style's Bass" when it's in effect. I added a test.

### Nonblocking

3. **OTS explanation: clear.** Both the drawer explainer and the rewritten `ots.*` tooltips answer the owner's question in plain words: it's a sound setup for your hands, it sets voice, on/off, volume and octave for R1–3 and Left, and the band doesn't change.
   - This matches `parts.rs apply_ots`, which sets program, on, volume and octave.
   - It also matches genos-features.md "OTS contents", where tempo and split aren't in OTS.
   - OTS Link "Real Time · on press" is correct: `transport.main` changes on the press (app-api.md), and `session.rs` recalls on that change and when Link is turned on.
   - One suggestion: the explainer could add one clause tying OTS 1–4 to Main A–D even when Link is off, since that's the mental model that makes "four" make sense. Optional.
4. **Layering and Manual Bass vs the engine.**
   - `leftZone`/`chordsWhere` replicate `live.rs sounds()`/`chord_only` exactly. Left sounds means Left, or the Bass under MB. Lower outside Full Keyboard means chords only. Otherwise the Right parts play the whole keyboard (OM p.48).
   - The strip lamps use `sounding`, and the Left strip shows "Bass" / "own: …" from `playsBass`. The style Bass row uses `mutedByManualBass`. All of that is state-driven and correct.
   - Two nits:
     - (a) This is still a TS copy of engine routing. Ideally `state.surface` or `chord` would expose "who plays below the split", so it can't drift. Worth a NEED line.
     - (b) The `part.layer` tooltip and the "Right hand · G2 and up" label say "above the split". But with Left off, outside Lower + non-Full-Keyboard, the Right layer plays the whole keyboard. The map gets this right ("Right parts" in the left zone), but the label and tooltip don't.
   - Genos Right 3 split (Split Point Right 3) isn't modelled. That's fine, since the engine doesn't have it either.
5. **Pickup line.** "Launchkey faders 1–4 are in step…" shows an empty square where the hidden ↕ mark sits (see 1440 dark). It reads like a stray checkbox, so hide the mark when nothing's waiting.
6. **Double flash bump.** A click on an OTS card bumps `recalled` locally, and the `ots.applied` effect bumps it again, so the `{#key}` remounts twice. It's harmless, since only one animation is visible, but the local bump is only needed when re-recalling the same OTS.
7. **Escape on a focused picker** now reaches the global handler, which closes the drawer. That's the same as Escape anywhere else in the drawer, so I think it's fine. Mentioning it in case the owner wants Escape to only leave the picker.

### Conflicts with Mixer PR #81

`git merge-tree` of `m3/ui-mixer` and this branch merges cleanly. I also ran a trial merge of both with `npm run verify`: 177 tests pass, including the `docs/controls.md` sync test.
- #81 doesn't touch `FaderBank.svelte` or `lib/mirror.svelte.ts`.
- In `tooltips.ts`, the two PRs edit disjoint sections. #82 reuses `mixer.panel.*`/`mixer.pickup`, whose bodies #81 rewrites. The new wording ("CC 7 itself, no hidden gain"; Left "is the bass's level too under Manual Bass") still reads correctly on the Parts strips.
- Whichever lands second should still rerun `npm run docs:controls` after rebasing.
- The two drawers overlap in content: #81's Style tab and this drawer's "As written for" both list the written-for voices. `mixer/voice.ts` and `lib/api/voices.ts` don't duplicate code. The branch is behind `integration/m3-ui`, but merges into it cleanly.

**Verdict:** approve with afe3351 (blocking items fixed). Not merging.

PR #83, opened 2026-09-24
M3 UI: Settings drawer (Genos-grouped: chord, split, transpose, style, audio, MIDI, library)

ryangavin:
Builds the Settings drawer (`app/src/panels/settings/`). Its pages follow the Genos menus. Every control has a catalog tooltip, and every change applies at once, with no Save button.

| Page | Genos | Controls | Engine |
|---|---|---|---|
| Chord | Menu › Split & Fingering | The 7 fingering types. Each has its own tooltip and a one-line summary, and Single Finger's four shapes are drawn as a chart. Also Lower/Upper and Manual Bass, which is shown as "Upper only" in Lower. | `setFingering`, `setUpper`, `setManualBass` |
| Split | Split Point (Style + Left) | A mini keyboard from C0 to C6. You drag the amber marker or click a key, and it also works with ←/→, PgUp/PgDn and Home/End. It shows the note name in Yamaha numbering (C3 = middle C) and has −/+ buttons. | `setSplit`, `moveSplit` |
| Transpose | Menu › Transpose | Keyboard and Master steppers, each showing a signed readout, plus Reset. The page also shows "you play X; the style follows Y". | `stepTranspose`, `resetTranspose` |
| Style | Style Setting | Sync Start, Sync Stop (disabled when `syncStopAvailable` is false), Auto Fill and Stop ACMP. Section change timing, OTS Link timing and the Synchro Stop window are shown disabled and marked **Coming soon · M5**. | transport toggles |
| Audio | (yahaha) | Built-in synth on/off, the output pair (one button per stereo pair from `io.synth.channels`, so a multi-output interface like the Model 16 gets a grid), SoundFont and master volume. | `setSynthMuted`, `setAudioOutput`, `setMasterVolume`; SoundFont is **mocked** |
| MIDI | Menu › MIDI | Inputs set to "All, merged" or "Selected", with a switch for each source. Also the `yahaha` output port, the Launchkey connection and Palette LEDs. | input selection and palette LEDs are **mocked** |
| Library | (yahaha) | Style folders, counts (styles, categories, files that won't load, files indexing) and Rescan. | the counts are real; folders and rescan are **mocked** |

## Mocked settings (NEEDs on the board)
The engine doesn't have these settings yet. `app/src/lib/api/settings.svelte.ts` mocks them in the proposed shape. `view()` already prefers the real `AppState` fields once they exist. When the drawer runs on the real engine, mocked controls carry a "Not in the engine yet" badge.

- `{"type":"setSoundFont","file":"FluidR3_GM.sf2"}`, with `io.soundFonts: string[]` and `io.soundFontFile`
- `{"type":"setMidiInputs","all":false,"names":[…]}`, with `io.sources: {name, listening}[]` and `io.allInputs`
- `{"type":"setPaletteLeds","on":true}`, with `pads.paletteLeds`
- `{"type":"rescanLibrary"}`, with `library.roots: string[]` and `library.scanning`

## Shared files touched (claimed on the board)
- `help/tooltips.ts`: I only appended to the Settings and audio section. The new keys are `settings.tab.*`, `settings.split_strip`, `settings.section_timing`, `settings.ots_link_timing`, `settings.synchro_stop_window`, `settings.style_folders`, `settings.rescan`, `audio.synth_on`, `midi.merge_all`, `midi.output` and `midi.palette_leds`. `settings.open` has a new body. `docs/controls.md` was regenerated.
- `scripts/screenshots.sh`: I added settings shots. `?tab=<page>` opens a page; it is read in `panels/settings/nav.svelte.ts`, so `urlparams.ts` is unchanged.
- New file: `lib/api/settings.svelte.ts`.

All pages stay mounted and the inactive ones are `hidden`. That makes switching pages instant, and the existing "settings open" coverage state sees every control, so I didn't add any STATES. The split marker and its shading move with `transform` only.

## Checks
- `npm run verify` passes: svelte-check reports 0 warnings, eslint is clean, and all 180 tests pass. That includes the new `Settings.test.ts`, whose 15 tests cover the tabs, fingering, detection and Manual Bass, split keys, split drag, transpose, the style toggles and placeholders, audio, the SoundFont adapter, MIDI and the library, plus the strip geometry.
- I checked the drawer on the mock at 1024, 1440 and 1920 wide, in the dark and light themes.

## Screenshots
Chord (dark, 1440): chord
Split: split
Transpose: transpose
Style: style
Audio: audio
MIDI: midi
Library: library
Light theme: light light split
1024 / 1920:  
Tooltip: tooltip

## Notes
- #82 (Parts drawer) adds its own `ots.link_timing`, shown read-only. My `settings.ots_link_timing` is the M5 placeholder in Style Setting. When M5 lands, the two can merge into one key.
- The mock's synth has 2 channels, so the output-pair row shows a single "1/2" in the mock. On a TASCAM Model 16 it lists every pair.


ryangavin:
## Adversarial review: Settings drawer (#83)

I checked out the PR head (d913076), ran `npm run verify` (0 warnings, 180/180 tests passing) and `cargo build --release` (passes), and looked at every page on the mock at 1024, 1440 and 1920, in dark and light.

**Result:** 1 blocking finding, which I fixed and pushed. Head is now `ff9c43f`: the merge `b0991bd`, then the fix `ff9c43f`.

### Merge with integration/m3-ui (Mixer #81)
I merged `origin/integration/m3-ui` into the branch (no rebase). Git reported no textual conflicts in `tooltips.ts`, `controls.md` or `screenshots.sh`: the Mixer and Settings sections don't overlap. After the merge I regenerated `docs/controls.md` and it was byte-identical. `npm run verify` passes 194/194 and `cargo build --release` passes.

### Blocking (fixed in ff9c43f)
1. **On the real engine, mocked controls pretended to work and showed made-up data.** The "Not in the engine yet" badge was there, but under Tauri:
   - Clicking a SoundFont, an input switch, All/Selected, Palette LEDs or Rescan changed the adapter's local state. The UI then showed the change (SoundFont "switched", a source "off", a 1.2 s "Scanning…" pulse), but nothing reached the engine.
   - The lists used the mock's sample data: fake SoundFont files (`FluidR3_GM.sf2`, `MuseScore_General.sf2`), fake MIDI sources (`TASCAM Model 16`, `IAC Driver Bus 1`) and fake style folders (`~/Music/Styles/Genos`…).
   - One `mocked` flag, keyed only on `io.soundFonts`, drove all four badges. If the engine shipped SoundFonts first, the other three would lose their badge while still mocked.

   **Fix:** `settings.view(state, real)` now keeps a separate `mocked` flag for each field. On the real engine a missing field shows only what the engine actually reports:
   - SoundFont: only the file the synth plays.
   - Inputs: only `io.inputs`, all shown as listening. All/Selected shows neither option selected, because the engine doesn't say which.
   - Palette LEDs: "Set at launch".
   - Style folders: "The engine doesn't report its style folders yet".

   These controls are now inert, with each badge on its own field and a note naming the launch flag to use for now (`--sf2`, `--all-inputs` / `--input`, `--palette-leds`). The mock session behaves exactly as before. A new test covers real-engine mode: 4 badges, no sample data, and clicks change nothing.

### Nonblocking (fixed in the same commit, since they were small)
2. The Style page M5 placeholders now use the Genos option names from `docs/genos-features.md` §7:
   - Section Change Timing was "Next bar / Next beat"; it is now **Immediate / Next Bar**.
   - OTS Link Timing was "Real Time / Next Bar"; it is now **Immediate / At Main Section Change**.
3. `settings.synchro_stop_window` was reworded. It said the chord "stops counting as a release"; it now says it the way the RM does: hold longer than the window and Sync Stop cancels itself, so the style keeps playing.
4. `settings.section_timing` said "a change always waits for the bar line". It now says what the engine does: a new Main waits for the next bar, and with Auto Fill the fill starts at the next beat (`queue_at_bar` / `queue_fill`). `settings.ots_link_timing` now says it is always Immediate for now, which matches the engine recalling on the Main press.
5. The Launchkey fields of the tab tooltips were incomplete:
   - The Chord tab now also names P2 bottom pad 1 (Manual Bass).
   - The Style tab named only P1. It now also names P2 bottom pad 2 (Stop ACMP), matching `src/launchkey.rs`.

### Nonblocking (not changed)
6. On the split strip, the amber wash is cut with a straight line at the split key's right edge. With a black-key split, such as the default F#2, a sliver of G2 (a right-hand key) gets shaded under F#. It's only cosmetic, and the marker, flag and readout are all correct.
7. At 1440 wide, the drawer covers the right half of the Launchkey mirror, including the chord readout. All the drawers do this, so it isn't specific to this PR, but the layout owner may want to push the mirror aside.

### Settings-specific checks: all correct
- **Fingering explanations** match `src/fingering.rs` and `docs/genos-features.md`: Fingered and Multi put the root in the bass, On Bass uses the lowest note, AI infers chords from fewer than 3 keys, and Full Keyboard reads the whole keyboard.
- **Single Finger chart** matches `fingering::single`, which reads the highest key as the root: C = C; Cm = Bb + C (black key left); C7 = B + C (white key left); Cm7 = Bb + B + C. I checked the drawn keys pixel by pixel. The layout is A Bb B | C C# D Eb E and correct.
- **Yamaha numbering:** `noteName` equals `api.rs note_name` (`n/12 − 2`, same NOTE_NAMES). C3 = 60, the default is F#2 = 54 (`session.rs`), the range is 24–96, matching `SetSplit`'s clamp, and keys at and below the split are Left, matching `live.rs`'s `k <= split`. The marker renders between C2 and C3 at the right place.
- **Transpose:** the readouts show `transposeKeyboard` / `transposeMaster` directly. The buttons send `stepTranspose` ±1 and `resetTranspose` (the engine clamps to ±12). "You play X; the style follows Y" uses `fingered` / `name`, which `api.rs` documents as before and after Keyboard transpose.
- **Sync Stop** is gated on `syncStopAvailable`, and the note ("Full Keyboard types in Lower") matches `live.rs`. The Stop ACMP note matches `sound_stop_acmp`.
- **Tooltip coverage:** I deleted `settings.split_strip` and `coverage.test.ts` failed ("data-tip "settings.split_strip" is not in the catalog"), then restored it.
- **Performance:** a synthetic 120-frame split drag averaged 16.5 ms per frame with no long tasks. The marker and wash move by `transform` only, and there's no blur anywhere in `panels/settings`.
- **Accessibility:** the tabs use roving tabindex with ←/→/Home/End. The strip is a `role="slider"` with valuetext set to the note name, and it responds to arrows, PgUp/PgDn and Home/End. Fingering and choices are radiogroups, and the global `:focus-visible` ring applies.

**Approve:** yes, with ff9c43f. Not merged.

Issue #21, opened 2026-09-23
Audition on hover


ryangavin:
- Preview a style with a short default progression without disturbing the current performance, or as a clean handoff when it's stopped.

---
Milestone: **M4 · Style browser** · Spec: [docs/genos-features.md](../blob/main/docs/genos-features.md) · Backlog: [docs/backlog.md](../blob/main/docs/backlog.md)

PR #84, opened 2026-09-24
Style browser: categories, filter, favourites, recents, preview (#20, #21)

ryangavin:
Closes #20. Builds the UI for #21 on a mocked engine command. The engine work it needs is listed under NEEDs below.

## What it does
The style browser modal lives in `app/src/panels/browser/`.

- **Categories:** All, Favourites, Recent, then every library folder as a tree with style counts. A folder shows its own styles and its subfolders' styles. The category a style belongs to is its folder, the same rule the TUI and `src/library.rs` use.
- **Filter-as-you-type:** matches the style name, file name or folder, case-insensitive, like `Entry::matches`.
- **Columns:** tempo, time signature, section lamps, and an SFF version badge. The lamps are Intro I–III · Main A–D · Break · Ending I–III, in the section pads' own colours from `transport.lamps`. Unreadable styles show as red error rows with the reason, and they never load.
- **Keyboard:** ↑/↓, PgUp/PgDn and Home/End move. Enter loads, and the band keeps playing, as in the TUI. Shift+Enter previews or queues. Ctrl/⌘+D stars. Esc closes.
- **Track sync:** the list keeps the library's order, which is the order < Track / Track > step through. The ‹ and › marks show the Track neighbours, and the highlight follows any Track press, including the hardware's. The footer has its own ◀ Track / Track ▶ buttons, captioned with the neighbour names.
- **Favourites, Recent, Preview on select, last category:** stored client-side in `prefs.svelte.ts`, behind a small `PrefsStorage` interface. Only localStorage is implemented. The project has no Tauri store or fs plugin, and adding one would mean Rust and capability changes outside this panel. In Tauri the webview's localStorage is kept in the app data dir. Recent follows every style change, from the browser or from Track.
- **Performance:** rows are virtualised with a fixed 36 px height, so only about 30 row elements ever exist. `?styles=60000` gives the mock a deterministic synthetic library of 60,043 styles (`lib/api/mock-library.ts`, with error rows and files that have no name marker). Measured in Chrome at 60k: a filter keystroke re-renders within one frame (12–19 ms including the rAF wait), and scrolling holds a median frame of 16.7 ms with p95 of 18.3 ms.
- Every control has a catalog tooltip. New entries: `browser.all/favourites/recents/row_error/favourite/preview/preview_stop/queue/auto_preview`. The filter/folder/row/close bodies were rewritten, and `docs/controls.md` was regenerated. The filter field keeps focus all the time, so its tooltip shows on hover only; on every focus it would cover the list.

## #21 Preview UX
- **Band stopped:** each row has a ▶ button, and Shift+Enter works too. It plays a few bars of that style over a short default progression (C Am F G7, one bar each) without loading it. The footer shows *Previewing X · bar 2/4 · Am* with a ■ Stop button. **Preview on select** is an opt-in switch that previews the highlighted or hovered row after a 600 ms pause.
- **Band playing:** previews are off so nothing disturbs the performance, and the mock refuses `auditionStyle` while running. Each row instead offers **Next bar**, as does Shift+Enter: it queues the style to load on the next bar line. The queued row flashes "Queued" on the beat clock, and the browser stays open so you see it arrive.
- A preview ends when you load a style, start the band, press ■, or close the browser.
- Preview and Next bar only appear when the state includes `state.preview`. The real engine doesn't send it yet, so the Tauri app is unaffected until the engine supports it.

## Shared files touched (claimed on the board)
- `lib/api/types.ts`: an append-only provisional block with `PreviewCmd`, `PreviewState`, `AppState.preview?` and `LibraryEntry.format?`.
- `lib/api/mock.ts`: implements the preview commands, adds `MockOptions.styles` (the synthetic library, sorted folder then name as the engine sorts), and gives entries a `format`.
- `lib/api/session.ts` and `lib/urlparams.ts`: the `?styles=N` switch, documented in CONTRIBUTING.
- `help/actions.ts`: three `tipFor` cases.
- `help/tooltips.ts`: the Style browser section only.
- `help/coverage.test.ts`: three browser states (stopped, previewing, queued).
- `scripts/screenshots.sh`: browser shots appended.

## NEEDs (engine API; posted on the board)
1. `{"type":"auditionStyle","id":N}`
   - Only while the band is stopped. It plays style N's Main A at the style's own tempo and voices over the default progression, 4 bars, then stops by itself.
   - It leaves the loaded style, OTS, parts, mixer and transport untouched.
   - It is refused while running.
   - It ends early on `stopAudition`, `loadStyle`/`stepStyle`, `startStop` or a Sync Start chord.
2. `{"type":"stopAudition"}` ends the preview at once, with all notes off on channels 9–16.
3. `{"type":"queueStyle","id":N}`
   - Running: it loads the style on the next bar line. A later queue replaces an earlier one, and if the band stops first the style loads then.
   - Stopped: it behaves like `loadStyle`.
4. State: `AppState.preview = { audition: { id, bar, bars, chord } | null, queued: number | null }`.
5. `LibraryEntry.format: "SFF1" | "SFF2" | null` for the SFF badge. Without it the badge shows "—".

**Finding:** `Engine::load` switches styles immediately while the band runs. It seeks the new style's Main to 0 at the current tick, mid-bar, even though its comment says "continue from the next bar". `queueStyle` would give performers a clean change on the bar line.

## Verify
- `npm run verify`: typecheck, lint and 203 tests all pass. New tests: `panels/browser/model.test.ts` (filter rules, folder tree, lamps, cursor, Track order, 60k timing) and `panels/browser/Browser.test.ts` (load, filter, folders, error rows, favourites, preview, Preview on select, queue at the bar line, band start, Track sync, Recent, 60k virtualisation, prefs persistence).
- Root `cargo build --release` passes.
- Checked on the mock at 1024, 1440 and 1920, dark and light.

## Screenshots
| | |
|---|---|
| Stopped, 1440  | Previewing, with favourites  |
| Playing, 1440  | Queued for the next bar  |
| Light  | 1024  |
| 1920  | 60k styles, filtered  |

The preview, queued and 60k-filtered shots need interaction, so `npm run screenshots` can't take them. They were captured over CDP.


ryangavin:
## Adversarial review: Style browser (#84)

Reviewed `1b656e0`. I pushed two commits on top of it, so the head is now **`619c73a`**.

**Gates on the new head:** `npm run verify` passes (0 errors or warnings, 253 tests). `cargo build --release` passes. `docs/controls.md` regenerates with no changes.

### Shared files and conflicts
- **Blocking, fixed in `2c77652`.** Merging `origin/integration/m3-ui`, which now contains #81, #82, #77, #78 and #83, conflicted in `app/scripts/screenshots.sh` only. That conflict came from #83. I kept both the browser and the settings shot lines, and made a merge commit instead of rebasing.
- `types.ts`, `mock.ts`, `session.ts`, `actions.ts`, `tooltips.ts`, `coverage.test.ts`, `controls.md` and `CONTRIBUTING.md` all auto-merged cleanly against #81 and #82. `coverage.test.ts` passes with all three panels' STATES.

### Findings
1. **Blocking, fixed in `619c73a`: the browser re-implemented Track stepping in TypeScript.** It worked out ‹/› and the footer captions with `neighbours(app.library, position)`. #77 now has the engine send `state.surface.trackPrev/trackNext`, which is `Session::neighbour` → `Library::step`. The browser now reads them through `surfaceOf(app.state, app.library)`, like `Launchkey.svelte`, and falls back to deriving them only when no surface is sent. I added a test showing that the engine's `trackNext` wins.
2. **Track order is OK.** The list keeps `LibraryList.entries` in the engine's order and never re-sorts it. The mock's 60k library sorts by `(folder_lc, name_lc, path)`, which matches `Library::sort` in `src/library.rs`. The folder *tree* in the sidebar sorts segment by segment, so `Pop/Sub` comes before `Pop & Rock`, while the engine's string order puts `Pop & Rock` first. That only affects the order of the sidebar, never the row order. Nonblocking.
3. **60k styles is OK.** I measured in headless Chrome at 1440×900 with `?styles=60000`:
   - A filter keystroke takes 9–30 ms including one rAF. The worst cases are the broad one-letter queries `p` and `po`, which match 32k and 15k styles. No long tasks were observed.
   - Scrolling 240 frames at 180 px per frame: median frame 16.7 ms, p95 17.5 ms, max 17.7 ms. The viewport was never uncovered, and there are always 23–32 row nodes.
   - Edges: the filter going to 0 results shows the empty message with no rows. At 1 result, scrollTop is 0 and that row is highlighted. Clearing re-centres on the cursor. Narrowing the filter while deep in the list resets to the top. Resizing from 900 to 500 to 1200 re-measures via ResizeObserver, with no gap at the bottom. Jumping to the end shows the last row.
   - Nonblocking: `p` at about 30 ms costs a dropped frame on a very broad query. If it matters, debounce only when `rows.length` is above about 20k, or reuse the previous result when the query extends the last one.
4. **Keyboard is OK.** I checked each key over CDP with real key events. Enter loads and closes, and the band keeps playing. With the band stopped, Shift+Enter auditions, and pressing it again stops the audition. With the band playing, Shift+Enter queues the style, which loads on the next bar. Ctrl+D stars. Esc closes. The filter keeps focus after clicks on the sidebar and on stars. After the browser closes, whether by Enter, Esc or the scrim, focus returns to the body and Space/Enter reach the performance shortcuts again. Space toggles start/stop and Enter reopens the browser.
5. **Mocked features are OK: fully hidden on the real engine.** With `state.preview` absent, `rowAction` is null, so the ▶ and Next bar buttons are not rendered. The footer hint and the Preview on select switch are not rendered either, and Shift+Enter is a no-op. Neither `src-tauri` nor `src/api.rs` sends `preview`, and a test covers this case. The NEEDs are on the board.
   - Nonblocking: the `browser.filter` tooltip still says "Shift+Enter previews, or queues…", even on an engine without preview. Consider "(when the engine supports it)".
6. **Persistence is OK.** I seeded localStorage with these stale values and reloaded:
   - favourites: a deleted path plus the non-strings `42` and `null`
   - recents: a renamed path
   - `autoPreview: "yes"`
   - category: a deleted folder

   The result: Favourites 1 and Recent 1 count only paths that still exist. The deleted folder falls back to All. `autoPreview` goes to false. Nothing throws. Nonblocking:
   - (a) A renamed or moved file loses its star, because prefs are keyed by path. That is acceptable and documented.
   - (b) Stale paths are never pruned from storage. Recents are capped at 50, but favourites are not. Consider pruning against the library once it is indexed.
7. **Visual is OK.** I checked 1024, 1440 and 1920 in dark and light. There is no horizontal overflow, no truncated names at these widths, and no blur filters anywhere in the overlay. The footer hint hides under 1100 px on purpose. Nonblocking: Next bar only shows on the hovered or highlighted row, so it can't be discovered while playing without hovering. The footer hint covers this at 1440 and above, but the hint is hidden at 1024.
8. **Tooltips are OK.** Every browser control has a catalog entry. Deleting `browser.queue` makes `help/coverage.test.ts` fail (2 tests), and the typecheck fails too. The texts match `library.rs`: folder then name, substring match on name, stem or folder, and error rows skipped by Track.
9. **Accessibility is OK.** The filter is a combobox with `aria-activedescendant` pointing at `role=option` rows. The rows, stars and preview buttons have `tabindex=-1`, which is fine because the filter drives the list. Buttons have labels.

**Verdict:** approve after my two commits. Both blocking items are fixed, and nothing else is blocking.

ryangavin:
Correction to the gates line above: on the pushed head `619c73a`, `npm run verify` passes with **254** tests (253 plus the new surface-neighbours test).

PR #85, opened 2026-09-24
Main screen: stage scaling, keyboard strip, lead-sheet band, app icon, state.surface switch-over

ryangavin:
Main-screen shell work from the #72 review, plus the frontend switch-over to the engine's `state.surface` (#77 checklist).

## What changed

**Stage scaling** (`App.svelte`)
- The lead-sheet band, the Launchkey mirror and the keyboard strip now form one "stage". It fills the window by both width and height, with no page scroll down to the 900×600 minimum window.
- The stage is a size container. It uses `--u = min(100cqw / w, 100cqh / h)` and sizes everything in `em`, with no `transform` scaling, so it stays crisp at every size. The mirror keeps its proportions, and the band and the strip take whatever height is left, up to a limit.
- When the stage is taller than 1.45:1 (a 1024-wide window, for example), the mirror uses its stacked layout, with the faders under the pads, and grows larger.
- At 1024 nothing is cut off. Panel print scales with `--u`, the Scene/Function and transport columns are wider, and the Track style names wrap instead of truncating.

**Keyboard strip** (`panels/keystrip/`)
- Held keys are coloured by the part that sounds them (Right 1–3, Left, or grey when a key only feeds chord detection). Layered parts show as bands.
- It shows the split point, which you can drag or move with the arrow keys (it sends `setSplit` / `moveSplit`), and the Left split when it differs.
- It shades the chord-detection area (Lower, Upper or Full Keyboard) and marks the recognised chord's tones, with the bass ringed.
- It has 49 or 61 keys to match the connected Launchkey. You can switch it to 49, 61 or 88; the choice is remembered, and picking it again goes back to matching the Launchkey.
- Keys are positioned by percentage, so a key press changes one class and one custom property. It holds 60 fps (frame mean 16.67 ms, p95 17.5 ms).

**Lead-sheet band** (`panels/leadsheet/`)
- It shows the section playing (or what the band will start with), one cell per bar with a beat slash per beat, a progress bar across the section (driven by `transform`), and the queued section.
- Its middle lane (`data-slot="chart"`) is the slot the M8 iReal chord chart will fill. The contract is in CONTRIBUTING.md under "The lead-sheet band".

**App icon**: original artwork in `src-tauri/icons/icon.svg` (a graphite panel, pads and keys, no third-party marks). The icon set was generated with `cargo tauri icon`.

**Hardening**: the store applies a state only when its `version` is higher than the last one applied, and it resets when a new session attaches. `TauriSession` only emits newer snapshots.

**`state.surface` switch-over** (#77 frontend checklist)
- `types.ts` adds `SurfaceControl.cc` and `colour`, the full `ClockState`, `Pad.palette` with `PaletteLed`, `pads.paletteLeds`, and `fader` on keyboard and style parts. `AppState.surface` is now required.
- `BeatClock` uses the documented anchor formula. `clock.beats` is the free-running LED clock the lamps run on, and `clock.pos` is the section position. It no longer mixes bar/beat with phase and no longer caps at 0.999.
- `surfaceOf` returns `s.surface`, and `deriveSurface` is gone.
- The browser mock builds the surface the way `Session::surface` does (`lib/api/mock-surface.ts`), with the engine's labels and colours and Play/Stop/Scene/Function reported off.
- The Rust mock sends a full surface with section and LED anchors and simulated fader positions. `lib.rs` no longer strips the surface and reads it now (`state_now`).
- Palette-mode pads render `pad.palette`. The transport captions come from the surface labels.
- `engine-shape.json` was re-recorded from `docs/fixtures`, adding 43 surface, fader and palette paths.
- Two small edits in other drawers' files, both noted on the board:
  - Settings: Palette LEDs stays a mocked switch, because the engine has no `setPaletteLeds` yet, but it now shows the value the engine reports.
  - The Mixer test fixture gains `fader`.

## NEEDs (engine API, on the board)
- `AppState.keyboard`: `{ held: [{ note, zone, parts }], leftSplit, chordTones, chordBass }`. The keyboard strip shows held keys and chord tones from it. Only the mock sends it today; on the real engine the strip shows just the splits and the detection area. I'll align to m3/engine-needs' `keys` once its shape lands.
- `transport.sectionBars` (the length of the section playing). Without it the band shows one cell per bar.

## Verify
- `npm run verify` passes: 17 files, 286 tests, including tooltip coverage and new tests for the strip, the band, the store version guard, the surface and the clock.
- In `app/src-tauri`, `cargo test` (11 tests) and `cargo clippy -D warnings` pass. Root `cargo build --release` passes.
- Checked in the browser at 1024, 1440 and 1920 in both themes, and at 900×600.

## Screenshots
1440×900

1920×1080

1024×900 (stacked)

Light: 1440 · 1024
light 1440
light 1024

900×600 (minimum window)
min

Keyboard strip tooltip · help mode
keystrip tip
help

Icon source: `app/src-tauri/icons/icon.svg`


ryangavin:
## Adversarial review (coordinator relay)
The reviewer's fixes were pushed as 92e9b69.

**Blocking, all fixed:**
- On the real engine, the Shift layer compared actions by object identity, so it highlighted every button.
- The keyboard strip claimed there were no held keys or chord tones when the engine simply doesn't report them yet.
- The lead band stayed stuck on bar 1 when `sectionBars` was missing.

**Verified:**
- Mirror vs the `docs/fixtures/state.json` fixture: every label, LED and action matches.
- BeatClock formula.
- Scaling at 900×600 up to 1920×1200 with no scroll or clipping, 60 fps.
- Split drag commands.
- Version guard.
- The Settings and Mixer edits are justified.

**Nonblocking:**
- `detectionArea()` duplicates an engine rule, so it needs a `keyboard.detection` field.
- The page-tab colours are hardcoded.
- `clock.pos` isn't clamped (its only consumer clamps it).
- The unused fader buttons have no aria-label.

Issue #28, opened 2026-09-23
OTS Link timing (at section change vs immediate); OTS turns ACMP and Sync Start on


ryangavin:
_Details in the spec._

---
Milestone: **M5 · Sections & transport parity** · Spec: [docs/genos-features.md](../blob/main/docs/genos-features.md) · Backlog: [docs/backlog.md](../blob/main/docs/backlog.md)

ryangavin:
Implemented in #92 (full notes in docs/fills-and-rules.md).

- Decision: with "At Main Section Change", the OTS is recalled when the pressed Main starts playing, which is after its fill when there is one. While stopped, both timings recall at the press, because there is no section change to wait for.
- Decision: every OTS recall arms Sync Start while the band is stopped: the OTS buttons, OTS Link, and a style change with Link on. This follows OM p.47 and the DL OTS column (ACMP on, Sync Start on). While the band plays nothing changes, because Sync Start pressed while playing would stop the band. yahaha has no ACMP-off state.

ryangavin:
Done in #92, merged to main via #113.

PR #86, opened 2026-09-24
Engine: voices, runtime SoundFont/MIDI/LED settings, rescan, style preview & queue, next-bar style change, meters, held keys

ryangavin:
The engine side of the UI drawers' NEEDs on the coordination board, plus the coordinator's additions (keyboard strip, `sectionBars`, clock clamp). Docs: `docs/app-api.md` (new "Settings" commands, `keyboard`, `preview`, Meters, the style-change rule, the clock clamp); fixtures and `app/src/lib/api/engine-shape.json` re-recorded with `yahaha state-json`.

## Commands (AppCmd)

| JSON | Does |
|---|---|
| `{"type":"queueStyle","id":N}` | Playing: load at the next bar line. Stopped: same as `loadStyle`. |
| `{"type":"auditionStyle","id":N}` | Stopped only: the style's Main A over C Am F G7, 4 bars, then stops by itself. Refused (`failed`) while the band plays. |
| `{"type":"stopAudition"}` | Ends the preview. |
| `{"type":"rescanLibrary"}` | Walks `library.roots` again on a background thread. Ids stay stable. |
| `{"type":"setSoundFont","file":"X.sf2"}` | Loads another `.sf2` from the synth's folder off the audio thread and swaps it in between two buffers. |
| `{"type":"setMidiInputs","all":bool,"names":[...]}` | Runtime `--all-inputs` / `--input`. Sources connect and disconnect live. |
| `{"type":"setPaletteLeds","on":bool}` | Runtime `--palette-leds`. |

**Behaviour change (the bug from the board):** `loadStyle`, `loadStylePath` and `stepStyle` while the band plays now switch at the next bar line, like a Genos. The same section carries on (the same Main, or the nearest one the new style has), at the same bar position and the same tempo. A section queued for that bar line (an Ending, a Main) starts in the new style. A later change replaces the waiting one. Stopping first loads it then. `preview.queued` names the style while it waits. The old code switched mid-bar and restarted at bar 1.

## State (AppState and friends)

- `preview: { audition: {id, bar, bars, chord} | null, queued: id | null }`
- `keyboard: { held: [{note, zone: "left"|"right", parts: number[]}], leftSplit, chordTones, chordBass, detection: [lo, hi] }`: the shape KeyStrip already used. `detection` replaces the TS copy of the rule in `keystrip/keyboard.ts`.
- `transport.sectionBars: number | null`
- `io.sources: [{name, listening, pads}]`, `io.allInputs`, `io.soundFonts`, `io.soundFontFile`, `io.soundFontLoading`
- `library.roots`, `library.scanning`
- `pads.paletteLeds` (already there) now also follows `setPaletteLeds`
- `LibraryEntry.format: "SFF1" | "SFF2" | null`, from the header scan
- `LibraryList.voices: [{program, bankMsb, bankLsb, name}]`: 128 GM voices on bank 0
- `Session::meters()` / Tauri `meters` command → `Meters { atMs, channels: [{channel, peak}], master: [l, r], clips }`
- `ClockState::position` clamps at 0, and so does the app's `clock.pos` (review item N4 of #77)

## Deviations from the proposed shapes

- **Voices** are on `LibraryList.voices`, the side channel the NEED offered, not on `AppState`. The list is 128 entries that never change, and `AppState` is fetched up to 60×/s. `voiceList()` now takes the library.
- **`io.sources[].pads`** is new. `name` is the raw CoreMIDI name that `setMidiInputs` matches, so the DAW port no longer carries a ` (pads)` suffix there. `io.inputs` keeps the suffix.
- **Meters** are a side channel (`meters()`), not part of `AppState`. They change with every audio buffer, and republishing the state for them would flood every client. Peaks are "since the last call", for a single reader. The UI handles decay and peak hold.
- **Per-part meters: each part now has its own synthesizer.** rustysynth doesn't expose per-channel output, so the built-in synth is now a *rack* of 12 `Synthesizer`s (ch 1–4, 9–16) sharing one SoundFont. The mix is the same, because voices, reverb and chorus are linear and every instance has the same settings. The meters are real post-FX part levels. The cost is 12 reverb+chorus instances instead of 2. The other option was vendoring rustysynth with a patch, and I preferred not to do that.
- **A dropped MIDI source** gets All Notes Off (CC123) on the keyboard parts' channels immediately, from the engine thread. The input thread can't be woken from outside CoreMIDI, so it releases that source's keys (Keys, route, chord and Sync Stop) with the next packet from any source. CC123 also stops notes that another source is holding. That is rare, and it's documented.

## Real-time notes

- **Audition** is a second `Engine`, built on the control side, handed to the engine thread through a ring, played there, and returned through a ring to be freed. When it ends, the band's setup is re-sent (`Engine::resync`). It ends early on `stopAudition`, a style change, START/STOP, Arm, Panic, or a Sync Start chord (ended *before* the band starts, so they never overlap).
- **Pending styles** wait inside the engine (`change_style`). Replaced and old styles go back through the existing old-style ring.
- **`tests/engine_no_alloc.rs`** (counting global allocator) checks that a full preview and a next-bar style change allocate and free nothing on the engine thread.
- **SoundFont swap:** load and rack build happen on a `yahaha-sf2` thread. The rack goes to the audio thread through a ring. A `Shadow` of each channel's bank, program, CCs and bend is replayed into the new rack. The old rack fades out over one buffer and returns through a ring to be freed on the control thread.
- **Held keys, per-source slots and meters** use atomics only. Keyboard sources now get their own input-port tags (running status per source, too).

## Tests

- **Sim (engine level):** waits for the bar line and keeps the bar position; Main B carries over; a later change replaces the waiting one; stopping first loads it; a queued Ending starts in the new style.
- **Session (offline):** a 4-bar preview leaves style, parts, OTS and mixer alone and re-sends the band's levels; preview refused while playing and ended by a Sync Start chord, `stopAudition` or START/STOP; `queueStyle` stopped and playing; `stepStyle` from the waiting style; `format` and `voices`; a rescan adds and drops files with stable ids; the keyboard strip state (parts, zones, chord tones, detection in Lower, Upper and Full Keyboard) and `sectionBars`; palette LEDs at runtime; the MIDI input choice; the SoundFont switch (off-thread load into the swap ring, bad names refused); JSON shapes of the new commands and fields.
- **Live:** a dropped source releases only the keys it alone held (chord released, Sync Stop fires); running status is per source.
- **Synth:** each part is metered on its own channel and follows CC7; a new rack takes over voices and levels by replay.
- The existing style-change session test is rewritten for next-bar semantics.

## Gates

- `cargo build --release` and `cargo build --release --no-default-features` pass.
- `cargo test --release` passes with the corpus: 464 + 6 + 1 + 1.
- No new clippy warnings. The 10 remaining are pre-existing.
- `cargo test` passes in `app/src-tauri`: 11.
- `npm run verify` passes: 290 tests.

## App side

- **types.ts** has the new commands and fields; the provisional markers are dropped for `preview`, `keyboard` and `format`.
- **Both mocks** (`mock.ts`, `src-tauri/src/mock.rs`) implement the new commands and send the new fields.
- **`settings.svelte.ts`** reads the engine's fields and sends its commands. The per-setting "mocked" flags are now true only for an engine older than these fields; the drawer badges those settings and makes them inert. The adapter's own sample data is gone, and `Settings.test.ts` checks the real paths.
- **`keystrip/keyboard.ts` `detectionArea()`** now clips `keyboard.detection` to the strip.
- **`Session.meters()`** is added (Tauri and mock).

Out of scope, left alone: Solo and OTS Link timing (M5, #30, #28).


ryangavin:
## Adversarial review of #86 (head f23d1fe)

**Verdict: changes requested.** There is one blocking finding, the 12-synth rack. I have a fix ready, but it is **not pushed** because commit signing failed on my side (1Password: "failed to fill whole buffer"). The fix is staged on the local branch `review/pr86-rack-fix` in the reviewer worktree. There is also a patch at `scratchpad/pr86-review-fix.patch`, with the commit message at `scratchpad/pr86-fix-commit-msg.txt`. To land it, someone who can sign needs to commit it and push it to `m3/engine-needs`.

Gates on f23d1fe all pass:
- `cargo build --release` passes, with and without `--no-default-features`.
- `cargo test --release` passes: 464 + 6 + 1 + 1.
- `app/src-tauri` `cargo test` passes: 11.
- `npm run verify` passes: 290.

### 1. BLOCKING: the 12-instance rack more than doubles the audio callback, and its voice ceiling reaches the buffer period

This is an offline render benchmark. It uses the densest corpus style (`T5Style/LatinDisco2.T158.prs`, 892 note-ons per 4 bars), 16 bars of C Am F G7, plus right-hand chords: 4-note chords every beat on Right 1/2/3 and 2 notes on Left. The SoundFont is Arachno (the app's default, `first_sf2`). The settings are 48 kHz, 64 frames, and the timing covers MIDI apply, render and mix for each 64-frame block. There are 28,800 blocks per run, and each figure below is from two runs.

| design | mean | p50 | p99 | p99.9 |
|---|---|---|---|---|
| before (integration/m3-ui: band + player synth) | 99–100 µs | 95 | 181–188 | 262–285 |
| **#86 rack (12 × rustysynth)** | **227–231 µs** | 222–225 | **373–403** | 491–763 |
| same, with sustain pedal and 8th-note chords on Right 1–3 | 311 µs | 307 | 499–505 | 625 |
| synthetic full load (11 parts × 128 pad voices, pedal down) | old 76 µs / **rack 552 µs** | | old 144 / **rack 836** | old 220 / **rack 1002** (max 1283) |

Where the extra cost comes from:
- About 60 µs is the 10 extra reverb and chorus sets, roughly 5 µs per instance per block. I measured this by comparing the rack with the effects on and off.
- Most of the rest is **polyphony**. Each part now has 128 voices of its own, 1,536 in total. Before, the band shared 128 and the player shared 128. In a dense style the old band pool was already stealing voices. The rack renders all of those voices, so the typical case rises 2.3× and the worst case has no practical bound. A full load reaches about 1 ms of the 1.33 ms buffer.
- The rack with 256 voices per instance renders bit-identical (diff 6.5e-7) to the old design with 256 voices. So the rack is linear, as claimed. The difference comes only from voice stealing.

**"The mix is the same" does not hold** wherever the old pools stole voices:
- On LatinDisco, the old and rack mix differ by 1.8e-2 RMS against a 0.41 signal, which is −27 dB.
- With the pedal, the rack mix is 2.5× louder (RMS 0.97 vs 0.39), because the old 128-voice player pool stole those voices.

Static per-part polyphony caps don't fix this. I tried caps summing to 496. That bounded the CPU at about 140–170 µs mean, but pooled parts then stole more than before (5.7e-2 deviation vs 1.8e-2). Parts that need many voices at once, like a pad on Right 2 or Chord 1/2, lose the most.

A shared send bus isn't possible either: rustysynth's `Reverb` and `Chorus` are private, and the sends are per voice (CC91 + the instrument generator).

**Fix (on `review/pr86-rack-fix`):**
- It vendors rustysynth 1.3.6 (MIT) as `vendor/rustysynth` (`rustysynth = { path = "vendor/rustysynth" }`) with one patch, `Synthesizer::channel_peaks()` / `reset_channel_peaks()`. `render_block` also accumulates each voice into a per-channel row. The main mix path is untouched, so the audio stays **bit-identical to upstream**.
- `Rack` goes back to the two pooled synthesizers (band and player, 128 voices each) and reads their channel peaks. The rest of the API stays the same: `Rack::load`, the swap rings, `Shadow`, `take_meters` and `RACK_CHANNELS`.
- Results on the same benchmark:
  - Plain: mean 103 µs, p99 192–195 µs.
  - Pedal: mean 106 µs, p99 195–197 µs.
  - Full load: mean 78 µs.
  - The mix diff against the old design is **0.0**.
- The meters now show each part's dry level (after CC7, CC11, pan and master). They no longer include reverb and chorus, because those are shared. `docs/app-api.md` is updated to say so.
- Cost of the meters: about 4 µs mean, measured as the patched old path (103 µs) against the upstream old path (99 µs).

### 2. Nonblocking: a SoundFont swap loses each channel's bend range and tuning (RPN 0/1/2)

`Shadow::replay` sends controllers in number order, so CC6/38 go out before CC100/101. A fresh rustysynth channel has `rpn = -1`, so that data entry is dropped. Then CC101/100 = 127/127, the null RPN, is selected. After a swap, every part bends ±2 semitones instead of the style's range, and any tuning is lost.

This is proven by the new test `rack_tests::a_new_rack_keeps_the_bend_range_and_tuning`. It fails on the PR's replay (max sample diff 0.21) and passes with the fix, which replays each RPN's data under its own select and then restores the last select. The fix is on the branch.

### 3. Nonblocking, but it can leave notes ringing for good: a dropped source with the sustain pedal down

`Cmd::KeysOff` sends only CC123. rustysynth keeps released voices while the hold pedal is down (`voice.rs`: `ReleaseRequested && !hold_pedal`). The input thread sent the dropped source's CC64 to all four keyboard-part channels, and its pedal-up will never come. The result is that the notes keep ringing, and so does everything played after them on the other keyboard, until Panic. (Panic has the same gap on channels 1–4. That is pre-existing.)

The fix sends CC64 0, CC1 0, centred bend and channel pressure 0, then CC123. The test is `source_tests::keys_off_lifts_the_dropped_sources_pedal`.

On the documented trade-off, that dropping a source also cuts notes another source is holding on channels 1–4: **acceptable** in my view. It only happens when a keyboard is unplugged or deselected while someone holds keys on a second keyboard.

Two other things I'm flagging without a fix:
- The input thread releases the dropped source's keys only on the *next packet from any source*. If the only keyboard is unplugged, the key strip and the chord recognizer keep showing those keys held until something else sends MIDI.
- `connect_inputs` connected new sources before it queued the dropped slot's releases. A new source could take the freed slot, and its first keys could then be released by the old slot's release. The fix queues the releases first. It's a tiny race window.

### 4. Next-bar style change: OK

I ran a throwaway sweep of 17 style pairs × 4 scenarios, including 3/4 `AcousticBallad` ↔ six 4/4 corpus styles:
- The swap lands exactly on the bar line, and no new-style note comes before it.
- No note (ch 9–16) from before the swap is still held after it, and no bend is left off centre.
- The tempo is kept, and the anchor lands on a whole bar of the new style.
- The bar advances by a new-style bar after that.
- A **fill in flight** (Auto Fill, Main B pressed mid-bar) ends at the bar line and the new style comes in on **Main B, bar 0**.
- A queued **Ending** starts in the new style at its bar 0.
- **Sync Stop** (chord released while a style waits) stops the band and loads the waiting style, with nothing pending.

### 5. Audition: OK

`tests/engine_no_alloc.rs` is real. I checked it with two mutations and both fail as they should:
- Dropping the `Audition` on the engine thread in `end_audition`: "frees on the engine thread".
- A `Vec::with_capacity` in `swap_style`: "allocations on the engine thread".

Refusal is checked twice, on the control side (`snap.running`) and again on the engine thread. Every exit path (natural end, `stopAudition`, a style change, START/STOP, Arm, Panic, a Sync Start chord, `EngineLoop::stop`) goes through `end_audition` or returns the `Audition` through the ring. After a preview, `resync` re-sends the band's whole setup with a fresh `Mirror`.

Nonblocking notes:
- A preview's SysEx/NRPN that the band's init doesn't reset stays on an external XG module.
- A mixer move or `TogglePart` on the stopped band during a preview sends CC7 on the preview's channels.

### 6. Runtime SoundFont swap: OK

- A missing, empty, truncated (3 MB of a 155 MB .sf2), 16-byte or random-bytes file all give a clean `Err` on the `yahaha-sf2` thread. None panics, and the audio thread never blocks.
- The swap buffer renders both racks once and replays the shadow. With the fix, that costs about 2 × 100 µs in the dense case.

Notes that are sounding when the swap happens fade out over one buffer and are not retriggered, so held notes stop. That's acceptable, but it should be documented.

### 7. App side: OK

- `settings.svelte.ts` sends the real `setSoundFont`, `setMidiInputs`, `setPaletteLeds` and `rescanLibrary` commands. It gates them per setting only for older engines.
- The browser's Preview / Next bar controls appear on the real engine because `AppState.preview` is always serialized.
- `KeyStrip` reads `state.keyboard`, including `detection`.

### 8. Real-time safety

Apart from item 1, it's clean:
- The rings, retired-style slots, atomics and the shadow are all fixed-size.
- The patched `render_block` uses preallocated rows.
- The old-rack ring (4) is drained on every pump.

---
All the fixes above are on `review/pr86-rack-fix`, unpushed because signing failed.

